PR #4, opened 2017-11-20
refactor readme and add contributing

garfield009:
refactor readme.md and add contributing and update arc


garfield009:
agree and no problem

PR #8, opened 2017-11-22
df-daemon: change the import path

zhouhaibing089:
basically, as a go project, it should be discoverable by go tools,
thus I makde a change to update its import path to be:

  github.com/alibaba/Dragonfly/src/daemon/src/df-daemon

with this change, you can run:

  go get github.com/alibaba/Dragonfly/src/daemon/src/df-daemon

and get a compiled binary in your $GOPATH/bin.

### DIFF
--- a/src/daemon/src/df-daemon/initializer/initializer.go
+++ b/src/daemon/src/df-daemon/initializer/initializer.go
@@ -14,23 +14,25 @@
 package util
 
 import (
-	"os"
-	"path/filepath"
-	log "github.com/Sirupsen/logrus"
-	"strings"
-	"os/user"
 	"flag"
-	"time"
+	"fmt"
 	"io"
 	"io/ioutil"
-	"regexp"
+	"os"
 	"os/exec"
-	"fmt"
+	"os/user"
+	"path/filepath"
 	"reflect"
-	"df-daemon/constant"
-	. "df-daemon/muxconf"
-	"df-daemon/util"
-	. "df-daemon/global"
+	"regexp"
+	"strings"
+	"time"
+
+	log "github.com/Sirupsen/logrus"
+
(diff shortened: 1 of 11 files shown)


zhouhaibing089:
cc @allencloud PTAL.

allencloud:
Yes, we need to use gofmt, golint or some other tools to ensure the code quality. 
While I am not the one who is responsible for dragonfly. @zhouhaibing089 
But you can still find me in dragonfly's community.

zhouhaibing089:
@allencloud Thanks, yeah, there are many things to do in order to make it great(though it looks great already according to the numbers).

zhouhaibing089:
@garfield009 could you please take a look?

src/daemon/src/df-daemon/initializer/initializer.go
```diff
@@ -69,7 +71,7 @@ func cleanLocalRepo() {
 						log.Warnf("walk file:%s error:%v", path, err)
 					} else {
 						if info.Mode().IsRegular() {
-							if time.Now().Unix() - reflect.ValueOf(info.Sys()).Elem().FieldByName("Atim").Field(0).Int() >= 3600 {
+							if time.Now().Unix()-reflect.ValueOf(info.Sys()).Elem().FieldByName("Atim").Field(0).Int() >= 3600 {
```
allencloud: I think this block of code are not so good. I recommend that we should use more gotools to guarantee the code format, lint, vet and some more other things.

zhouhaibing089: There are more bits needs to be revisited, which I would like to do several PRs as a follow up, and make those changes gradually.

The tooling part which I believe we can track it as an issue.

garfield009: ok,you can do it,I will review your pull request as soon as possible.

PR #12, opened 2017-12-01
df-daemon: rewrite the cleanup code in initializer

zhouhaibing089:
the previous code on checking last access time is tricky and not
easy to read, instead we should make it more clearly that it is a
check on Stat_t.

Also, the df-daemon should deny non localhost access explicitly,
if this is reverse proxy for itself, then it is definitely a dead
loop.


src/daemon/src/df-daemon/initializer/initializer.go
```diff
@@ -62,51 +65,60 @@ func cleanLocalRepo() {
...
 	for {
-		select {
-		case <-time.After(time.Minute * 2):
-			func() {
-				log.Info("scan repo and clean expired files")
-				filepath.Walk(G_CommandLine.DFRepo, func(path string, info os.FileInfo, err error) error {
-					if err != nil {
-						log.Warnf("walk file:%s error:%v", path, err)
-					} else {
-						if info.Mode().IsRegular() {
```
zhouhaibing089: cc @allencloud I remember you have concerns on this piece of code, so here is the modification. :)

zhouhaibing089:
a little other comments:

1. the `select` is not necessary(nothing to select actually, only one channel available).
2. considering that a timer is expensive, I believe a time.Sleep is already enough. though a better approach is using Ticker.

chenchaobing:
@garfield009

chenchaobing:
@garfield009

src/daemon/src/df-daemon/initializer/initializer.go
```diff
@@ -62,51 +65,60 @@ func cleanLocalRepo() {
...
+				return nil
+			}
+			// get the last access time
+			statT, ok := info.Sys().(*syscall.Stat_t)
+			if !ok {
+				log.Warnf("ingore %s: failed to get last access time", path)
+				return nil
+			}
+			// if the last access time is 1 hour ago
+			if time.Now().Unix()-statT.Atimespec.Sec >= 3600 {
```
zhouhaibing089: This is not platform agnostic, `Atimespec` is for darwin_amd64`.

zhouhaibing089: Fixed.

lowzj:
Thanks for your excellent work!!

PR #13, opened 2017-12-01
move vendor to project root directory

zhouhaibing089:
As with a migration for other components to be written as go, we
should have a centralized place for vendoring. considering a well
defined project structure, root directory is good choice.


sandrom:
while I agree that the vendor shouldnt be where it is right now, the current strucuture isnt very much go anyway, as in go the source doesnt start in src/ directories. 

have a look at https://github.com/golang-standards/project-layout and https://peter.bourgon.org/go-best-practices-2016/ for some initial ideas about this. similar structures have been adopted in many open source projects already and they have proved very useful for closed-source-work i have seen too.

zhouhaibing089:
@sandrom I do not disagree with you. I guess the maintainer might feel more comfortable with incremental changes, for example, it is trivial change to move `vendor` to project root.

And then I would suggest creating `cmd` for binaries, and so on.

Sadly, this project does not have much activities right now.

sandrom:
@zhouhaibing089 it was really meant as explaining why your change is one of the reasonable ones but also where to possibly go from there. lets hope that someone of the maintainers will pick it up ;)

PR #59, opened 2018-04-19
chore: fix broken link & reduce build outputs

wrfly:
- the jdk8's download link is broken
- discard wget's annoying ans useless outputs
- reduce mvn installation's outputs

BTW, is it necessary to minimize the number of layers?


lowzj:
Thanks for your contributing. Welcome to submit another PR to reduce the number of layers!

PR #62, opened 2018-04-19
chore: shrink image layers & reduce image size

wrfly:
```
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
dragonfly           supernode-new       3ac6aed55569        37 seconds ago      613MB
dragonfly           supernode           63afe8270c3f        32 minutes ago      1.12GB
```
I just tested it by comparing the logs, and not sure it can work properly. So a verification of availability is quite necessary.

BTW, I'm really good at writing dockerfiles ;P


lowzj:
Thanks for your excellent work very much! We will verify the new images ASAP.

A little suggestion, it would be better if your commit message following the [Commit Rules](https://github.com/alibaba/Dragonfly/blob/master/CONTRIBUTING.md#commit-rules).

wrfly:
Oh sorry, I didn't notice those commit rules. I'll update it later.

PR #63, opened 2018-04-20
bugfix: broken link

wrfly:
Use github raw link for downloading file, otherwise it's an HTML page.

BTW, why do you use a tarball for code management? Maybe you can put the source code in this git repo, too?


wrfly:
https://github.com/alibaba/Dragonfly/issues/52

PR #106, opened 2018-07-11
refactor: use return fast to reduce ident to be more readable

allencloud:

refactor: use return fast to reduce ident to be more readable

With the original way, the code was so hard to read. This pull request tries to change this situation.

### DIFF
--- a/dfdaemon/handler/transport.go
+++ b/dfdaemon/handler/transport.go
@@ -55,16 +55,18 @@ var dfRoundTripper = &DFRoundTripper{
 var compiler = regexp.MustCompile("^.+/blobs/sha256.*$")
 
 func needUseGetter(req *http.Request, location string) bool {
-	var useGetter bool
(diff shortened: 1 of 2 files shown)


dfdaemon/handler/transport.go
```diff
@@ -56,14 +56,17 @@ var compiler = regexp.MustCompile("^.+/blobs/sha256.*$")
...
 	var useGetter bool
-	if req.Method == http.MethodGet {
-		if compiler.MatchString(req.URL.Path) {
-			return true
-		}
-		if location != "" {
-			return global.MatchDfPattern(location)
-		}
+	if req.Method != http.MethodGet {
+		return useGetter
```
zhouhaibing089: perhaps `return false` which makes it more clear.

allencloud: make sense

lowzj:
Thanks!

PR #113, opened 2018-07-16
 options: add options to replace global commandline

zhouhaibing089:
It is generally a bad idea to use global variables, it makes it is
harder to write unit tests, and harder to track where they are used.
This is an effort to make the first move, and construct an options
so it can be pass down to the underlying system.

This PR is based on https://github.com/alibaba/Dragonfly/pull/112


lowzj:
Could you please add some unit-tests for these changed codes?

zhouhaibing089:
Not sure which function I should add a unit test, could you clarify? I basically just did some code moving around..

zhouhaibing089:
@lowzj: I added a unit test for `AddFlags`.

And I did not do the same for `initParam` because of:

1. It is not testable right now.
2. I have some followup changes to remove `initializer` and `global` packages totally.

PR #114, opened 2018-07-18
test: add unit tests for dfget/cli

lowzj:


lowzj:
Test coverage is decreased because that unit tests for `cli.initialize` using by `exec.Cmd` are not counted in test coverage.

PR #121, opened 2018-07-26
refactor: add file util functions in file_util.go

zousheng:
implement some of file util functions in file_util.go.

### DIFF
--- a/dfget/util/file_util.go
+++ b/dfget/util/file_util.go
@@ -17,9 +17,16 @@
 package util
 
 import (
+	"bufio"
+	"crypto/md5"
 	"fmt"
+	"io"
 	"os"
+	"path/filepath"
 )
+
+// BufferSize define the buffer size when reading and writing file
+const BufferSize = 8 * 1024 * 1024
 
 // CreateDirectory creates directory recursively.
 func CreateDirectory(dirPath string) error {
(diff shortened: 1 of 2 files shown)


dfget/util/file_util.go
```diff
@@ -35,33 +37,75 @@ func CreateDirectory(dirPath string) error {
 
 // DeleteFile deletes a file not a directory.
 func DeleteFile(filePath string) error {
-	return nil
+	return os.Remove(filePath)
```
lowzj: `os.Remove` will delete a file or a directory, but this function `DeleteFile` just deletes a file not a directory.

dfget/util/file_util.go
```diff
@@ -35,33 +37,75 @@ func CreateDirectory(dirPath string) error {
...
+	return os.OpenFile(path, flag, perm)
 }
 
 // Link creates a hard link pointing to src named linkName.
 func Link(src string, linkName string) error {
-	return nil
+
+	if PathExist(linkName) {
+		if err := DeleteFile(linkName); err != nil {
+			fmt.Printf("failed to delelte old link %s", linkName)
```
lowzj: in the `util` package, we don't need to print any information, just return the result and error to the caller.

dfget/util/file_util.go
```diff
@@ -35,33 +37,75 @@ func CreateDirectory(dirPath string) error {
...
 }
 
 // OpenFile open a file. If the file isn't exist, it will create the file.
 // If the directory isn't exist, it will create the directory.
 func OpenFile(path string, flag int, perm os.FileMode) (*os.File, error) {
-	return nil, nil
+	if PathExist(path) {
+		return os.OpenFile(path, flag, perm)
+	}
+	pathDir := filepath.Dir(path)
```
lowzj: you can use `CreateDirectory` to create the parent directory of the file.

dfget/util/file_util.go
```diff
@@ -35,33 +37,75 @@ func CreateDirectory(dirPath string) error {
...
+		}
+	}
+
+	return os.Link(src, linkName)
+
 }
 
 // CopyFile copies the file src to dst.
 func CopyFile(src string, dst string) error {
+	data, err := ioutil.ReadFile(src)
```
lowzj: This implementation is not suitable for reading and writing large files.

dfget/util/file_util.go
```diff
@@ -35,33 +37,75 @@ func CreateDirectory(dirPath string) error {
...
 func DeleteFile(filePath string) error {
-	return nil
+	return os.Remove(filePath)
 }
 
 // DeleteFiles deletes all the given files.
 func DeleteFiles(filePaths ...string) {
+	if len(filePaths) > 0 {
+		for _, f := range filePaths {
+			if err := os.Remove(f); err != nil {
```
lowzj: you can use `DeleteFile` in the for loop.

dfget/util/file_util.go
```diff
@@ -35,33 +37,75 @@ func CreateDirectory(dirPath string) error {
...
+		fmt.Printf("failed to read file %s", dst)
+		return err
+	}
 	return nil
 }
 
 // MoveFile moves the file src to dst.
 func MoveFile(src string, dst string) error {
-	return nil
+	if err := DeleteFile(dst); err != nil {
```
lowzj: It will return an error when the `dst` is not exist and `os.Rename` could not be executed.

dfget/util/file_util.go
zhouhaibing089: Besides, the intention should be more clear: what if the provided filepath is a directory, are you going to delete all files under that directory?

zousheng: Thanks for advice, will fix it

lowzj:
merged, the travis-ci is canceled because that there is a lot of other jobs queued.

Issue #125, opened 2018-08-02
docker Verifying Checksum error when download  images from source


nonly:
### error info
5d1e1cf08c34: Already exists 
c6f6a9a081c7: Verifying Checksum 
8c8e72264313: Verifying Checksum 
59ff1dbb3c48: Verifying Checksum 
3df20762c818: Waiting 
50ac5ba0faac: Waiting 
5972742b20a6: Waiting 
bbd696e7f20d: Waiting 
fcd7d5affc42: Waiting 
cd2032e0b412: Waiting 
a9b1d28ec99a: Waiting 
b068a343d85e: Waiting 
ff8fa04f82fa: Waiting 
1887236555a6: Waiting 
c068e0014379: Waiting 
997b3b3d74b6: Waiting 
11058049e349: Waiting  

### reason
i suppose failed to 307 Temporary Redirect when download blobs from docker distribution with HTTP host header.

### possible solution
df-daemon should discard HTTP host header.

PR #126, opened 2018-08-02
bugfix: fix back source failed

nonly:
#125  [df-daemon] discard HTTP host header for backing to source successfully


lowzj:
merged into branch: `master`, `0.2.x`.
branch `master` is being refactoring, it's not recommended to use.

PR #130, opened 2018-08-13
refactor: add http_util.go and SupernodeAPI

lowzj:


dfget/core/api/api_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/go-check/check"
+)
+
+func Test(t *testing.T) {
+	check.TestingT(t)
```
zhouhaibing089: Is this the suggested approach for unit testing in this project?

dfget/core/api/supernode_api.go
```diff
@@ -0,0 +1,90 @@
...
+func NewSupernodeAPI() *SupernodeAPI {
+	return &SupernodeAPI{
+		Scheme:      "http",
+		ServicePort: 8002,
+		Timeout:     5 * time.Second,
+	}
+}
+
+// SupernodeAPI implements the communication between supernode and dfget.
+type SupernodeAPI struct {
```
zhouhaibing089: Maybe define this as an interface?

dfget/core/api/supernode_api.go
```diff
@@ -0,0 +1,90 @@
...
+func (api *SupernodeAPI) Register(ip string, req *types.RegisterRequest) (
+	resp *types.RegisterResponse, e error) {
+	var (
+		code int
+		body []byte
+	)
+	url := fmt.Sprintf("%s://%s:%d%s",
+		api.Scheme, ip, api.ServicePort, PeerRegisterPath)
+	if code, body, e = util.PostJSON(url, req, api.Timeout); e != nil {
+		return nil, e
```
zhouhaibing089: As you are using named return, just `return` should be enough.

dfget/core/api/supernode_api_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+type SupernodeAPITestSuite struct {
+	mock   *mockHTTPClient
+	origin util.SimpleHTTPClient
+}
+
+func (s *SupernodeAPITestSuite) SetUpSuite(c *check.C) {
+	s.origin = util.DefaultHTTPClient
+
+	s.mock = &mockHTTPClient{}
```
zhouhaibing089: Maybe I am missing something, it seems this client is not well initialized.

dfget/core/api/supernode_api.go
```diff
@@ -0,0 +1,90 @@
...
+		Timeout:     5 * time.Second,
+	}
+}
+
+// SupernodeAPI implements the communication between supernode and dfget.
+type SupernodeAPI struct {
+	Scheme      string
+	ServicePort int
+	Timeout     time.Duration
+}
```
zhouhaibing089: Maybe adding a new field `Client util.SimpleHTTPClient` here. This makes it easier to do unit test (no need to override variable `util.DefaultHTTPClient`).

dfget/core/api/api_test.go
lowzj: Yes, go-check provides more handy functions for unit testing.

dfget/core/api/supernode_api.go
lowzj: You're right, it will be more convenient for unit testing in feature development.

dfget/core/api/supernode_api.go
lowzj: I just want the return statement to be more clear.

dfget/core/api/supernode_api.go
```diff
@@ -0,0 +1,99 @@
...
+/* the url paths of supernode APIs*/
+const (
+	PeerRegisterPath      = "/peer/registry"
+	PeerPullPieceTaskPath = "/peer/task"
+	PeerReportPiecePath   = "/peer/piece/suc"
+	PeerServiceDownPath   = "/peer/service/down"
+)
+
+// NewSupernodeAPI creates a new instance of SupernodeAPI with default value.
+func NewSupernodeAPI() SupernodeAPI {
```
zhouhaibing089: It might be better to allow initialization with parameters.

dfget/core/api/supernode_api.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/alibaba/Dragonfly/dfget/types"
+	"github.com/alibaba/Dragonfly/dfget/util"
+)
+
+/* the url paths of supernode APIs*/
+const (
+	PeerRegisterPath      = "/peer/registry"
+	PeerPullPieceTaskPath = "/peer/task"
+	PeerReportPiecePath   = "/peer/piece/suc"
+	PeerServiceDownPath   = "/peer/service/down"
```
zhouhaibing089: By the way, is there any reason to export these api paths?

dfget/core/api/api_test.go
zhouhaibing089: I'm not sure how much adoption that go-check has. I always find that simple table driven unit tests are enough.

dfget/core/api/supernode_api.go
lowzj: I think that it could write another function to initialize with customized parameters, and this function just creates an api with default value.

dfget/core/api/supernode_api.go
lowzj: It seems to be better not to export them.

dfget/core/api/supernode_api.go
zhouhaibing089: That works, too

dfget/core/api/api_test.go
lowzj: All of them are allowed to write unit-test in this project. IMOP, `go-check` is much more handy and clear than `Golang testing` in some scenarios, such as `util_test.go`, `rate_limiter_test.go`. And I didn't like to write `if statement` to check whether the result is expected in unit testing.

zhouhaibing089:
lgtm

Issue #142, opened 2018-09-21
Feature : Df-daemon support multi-registrys


mushixun:
In the actual production environment, there will be multiple registrys providing services. These 
registrys provide different types of image downloads according to requirements. However, when a 
computing node downloads images through df-daemon, only one Registry address can be specified, 
and another Registry is required. The df-daemon must be reconfigured and restarted. This is 
unacceptable for the production environment. I hope df-daemon provides support for multiple 
Registry. If the first Registry configured cannot find the image, it will use another Registry.

mushixun:
@zhouhaibing089 OK！There are a lot of registry in our company, computing nodes are mixed to run different applications, and the images of these applications may be distributed in different registry, and the registry parameter of df-daemon can only point to 1 registry, to point to another registry only modify the configuration and restart the df-daemon, so I implemented the issue.
sample : df-daemon --registry registry.example.com

allencloud:
Should this be included in dfget side? @mushixun @zhouhaibing089 @fengzixu

lijianfeng1993:
I have the same requirement.

mushixun:
@allencloud I think to do this in dfget is better

zhouhaibing089:
Could we have a proposal first?

Starnop:
Has been solved. The related PR : #410 #486

PR #143, opened 2018-09-26
feature: Df-daemon support multi-registries

mushixun:
Provide df-daemon rest_api for --registry parameter,this can change the registry_address online.
sample:
curl -i http://df-daemon_ip:65001/args?registry=registry.example.com

### DIFF
--- a/dfdaemon/handler/param_handler.go
+++ b/dfdaemon/handler/param_handler.go
@@ -15,6 +15,7 @@
 package handler
 
 import (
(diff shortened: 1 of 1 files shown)


zhouhaibing089:
Could you explain how would this help #142, what would be the new user experience?

mushixun:
@zhouhaibing089 My PR need I to do something else？

lowzj:
There are pull requests to support multi-registries: #389 #410

PR #153, opened 2018-10-10
fix fmt error

silenceshell:

fix go fmt error

NONE

No

NA

NA

### DIFF
--- a/dfget/config/config_test.go
+++ b/dfget/config/config_test.go
@@ -148,8 +148,8 @@ func (suite *ConfigSuite) TestCheckURL(c *check.C) {
 		"127.0.0.1:8080":       true,
 		"127.0.0.1:8080/我":     true,
 		"127.0.0.1:8080/我?x=1": true,
-		"a.b":            true,
-		"www.taobao.com": true,
+		"a.b":                  true,
+		"www.taobao.com":       true,
 		"https://github.com/alibaba/Dragonfly/issues?" +
 			"q=is%3Aissue+is%3Aclosed": true,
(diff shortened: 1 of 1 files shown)


allencloud:
Thanks a lot for your contribution. @silenceshell 
I am afraid that you need to gofmt your code. 
```
./dfget/config/config_test.go
CHECK: please format Go code with 'gofmt -s -w .'
Makefile:31: recipe for target 'check' failed
make: *** [check] Error 1
delete /tmp/dragonfly/build
```

silenceshell:
I checked the CI log and found Dragonfly use golang 1.9, while I am using golang 1.11

It seems that since golang 1.11 `go fmt` changes the alignment format. Dragonfly specify its requirement as below.

```
Requirements: go1.7+ and the go cmd must be in environment variable named PATH.
```

Is it better to specify requirement as golang 1.7 ~ 1.10 ?  We can add a check step for golang in `build/clinet/run.sh`.

allencloud:
> I checked the CI log and found Dragonfly use golang 1.9, while I am using golang 1.11

For this part, do we need to add a requirement to set down the golang version.

Currently, PouchContainer plans to upgrade to golang 1.10.4. While we are not planning to use the newest version of golang. @silenceshell

silenceshell:
Thanks for your comment @allencloud !

You mean we DO need a requirement to set down the golang version? Because it is confused for people who uses golang 1.11 and above in the future.

My suggestion is that we should add a check step for golang version in build/clinet/run.sh.

PR #174, opened 2018-10-25
feature: add blocking queue

lowzj:
Add a general implementation of blocking queue to use easily instead of directly using buffered channels.

The buffered channels can only implement blocking queue with a fixed capacity and there're many redundant codes when we directly use it.

So this pull request defines a general blocking queue interface and implements 2 kinds of blocking queue:
* `infiniteQueue`: we can use it when we don't care about the capacity of queue beforehand. It uses `List` to store elements, uses a `unsafe.Pointer(point to a chan)` and `sync.Mutex` for concurrency controlling.
    ```go
    // create a blocking queue with an infinite capacity
    q := NewQueue(0)
    ```
* `finiteQueue`: we can use it when we want a fixed capacity queue. It's implemented by buffered channels.
    ```go
    // create a blocking queue with fixed capacity
    q := NewQueue(10)
    ```

### DIFF
--- a/dfget/util/queue.go
+++ b/dfget/util/queue.go
@@ -1,0 +1,209 @@
+/*
+ * Copyright 1999-2018 Alibaba Group.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


allencloud:
Could you add more details in the pull request description? @lowzj 
Otherwise others could not review your code easily. Thanks a lot.

PR #179, opened 2018-10-26
fix http cache miss for servers which supports etag only

silenceshell:

I used harbor as our docker registry, with Dragonfly as a p2p proxy, and I found that supernode has never hit the cache in `./repo/download`. 

It seems that supernode always downloads the blob file again from harbor, even the same file has exist in the `./repo/download/{taskid}/files`

In the method `CacheDetectorImpl.parseBreakNum`, supernode only uses the `Last-Modified` to check whether the cache has expired.

However, Harbor only returns the `ETag` header in the response message. The `ETag` is the sha256 digest of the blob. 

This PR adds support for ETag identifier with http cache. `parseBreakNum` will return 0 when both `Last-Modified` and `Etag` check failed.

NA

NA

1. use Harbor as docker registry(I don't think harbor is nessary, however I have not tried), and Dragonfly as a p2p proxy.
2. `docker pull` a image from harbor on a node. check the `app.log` on supernode, will find that supernode download the blob file from harbor.
3. wait 3-6 minites, check `data-gc.log` on supernode, when it says supernode has gc all count, `docker pull` the image again.
4. check the `app.log` on supernode, will get `cache full hit for taskId: ...`. That means the cache hits.

NA

### DIFF
--- a/src/supernode/src/main/java/com/alibaba/dragonfly/supernode/service/cdn/Downloader.java
+++ b/src/supernode/src/main/java/com/alibaba/dragonfly/supernode/service/cdn/Downloader.java
@@ -230,7 +230,7 @@ public class Downloader implements Callable<Boolean> {
                         }
                     }
                     if (code == checkCode) {
-                        fileMetaDataService.updateLastModified(taskId, conn.getLastModified());
+                        fileMetaDataService.updateLastModifiedAndETag(taskId, conn.getLastModified(), conn.getHeaderField("ETag"));
                         isConnected = true;
                         break;
                     }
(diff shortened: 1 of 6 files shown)


silenceshell:
@lowzj Thanks for your review. I have updated the code, now only send one request to check. Would you please help to review again?

src/supernode/src/main/java/com/alibaba/dragonfly/supernode/service/cdn/Downloader.java
```diff
@@ -231,6 +231,7 @@ private HttpURLConnection createConn(String taskId, String fileUrl, String[] hea
                     }
                     if (code == checkCode) {
                         fileMetaDataService.updateLastModified(taskId, conn.getLastModified());
+                        fileMetaDataService.updateETag(taskId, conn.getHeaderField("ETag"));
```
lowzj: This could be set together with `lastModified` because of they are used together to do the same thing always.

silenceshell: Thanks, this is much better. I have updated the code, would you please review again?

lowzj:
lgtm

PR #195, opened 2018-10-31
docs: make swagger.yml to define SuperNode's API

allencloud:

make swagger.yml to define SuperNode's API

Swagger is helpful tool to make API definitions.

What I did:

* define Dragonfly supernode's API via swagger;
* generate api structs via swagger command which is also an added file ./hack/generate-swagger-models.sh
* vendor some packages which is used in the generated struct file.

fix https://github.com/alibaba/Dragonfly/issues/184

no need 

no

no


lowzj:
lgtm

PR #199, opened 2018-11-01
chore: fix go vet check failure and misspell of shell scripts

lowzj:

The `"${packages}"` will be treated as one whole string parameter by `go vet`. Actually we want `go vet` receives all files not one file. A workaround is:

```bash
echo "${packages}" | xargs go vet 2>&1
```


allencloud:
LGTM

PR #204, opened 2018-11-02
docs: define preheat api using swagger

lowzj:

define preheat api using swagger


apis/swagger.yml
```diff
@@ -539,14 +539,29 @@ definitions:
         type: "string"
         description: |
           ID of preheat task.
+      status:
+        type: "string"
+        description: "the status of preheat task"
```
allencloud: Could you list the enum value of the status field.:

* waiting
* running
* failed
* success

And we should add more details about that.
Since when a preheat task is finished, I think after some time this object should be GCed, otherwise it will lead to memory leak.

allencloud:
```
apis/types/error_response.go:14:62: "occured" is a misspelling of "occurred"
./apis/swagger.yml:585:51: "occured" is a misspelling of "occurred"
```
need update.

apis/swagger.yml
```diff
@@ -539,14 +539,37 @@ definitions:
...
+      startTime:
+        type: "string"
+        format: "date-time"
+        description: "the preheat task start time"
+      finishTime:
+        type: "string"
+        format: "date-time"
+        description: "the preheat task finish time"
```
allencloud: If  it has finished, what time will it be?

allencloud:
LGTM

PR #218, opened 2018-11-07
docs: use html label to format swagger output

lowzj:


lowzj:
closed. move it to #216

PR #226, opened 2018-11-10
docs: remove binary in package

allencloud:

remove binary in package. 

In the source code of Dragonfly, I do not think we should vendor this binary file.
In the future we could make this binary stored in the release link or somewhere else.

none

no need.

none

none


xiang90:
lgtm

PR #267, opened 2018-12-18
docs: update copyright header of source file

allencloud:

update copyright header of source file from `Copyright 1999-2018 Alibaba Group.` to `Copyright The Dragonfly Authors.`

fixes https://github.com/dragonflyoss/Dragonfly/issues/248

no need

none

none

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright 1999-2018 Alibaba Group.
+# Copyright The Dragonfly Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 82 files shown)


lowzj:
The bad link problem is caused by the change of domain `d7y.io`, I will fix it later.

zhouhaibing089:
Also a heads up here, do we want to enforce header presence for every source files?

allencloud:
> Also a heads up here, do we want to enforce header presence for every source files?

You mean not only the Go source file, but also any other files? @zhouhaibing089

zhouhaibing089:
@allencloud: And also a check in CI jobs(if we haven't done so already)..

Issue #252, opened 2018-11-30
[Question] dfget fails to pull images


shenlanse:
## Question

After installed all component of dragonfly and finished all configuration according to this [document](https://d7y.io/quick_start/), I failed to pull images via dragonfly with great possibility.

when failed, check the `dfclient.log`:

```
[2018-11-30 16:14:54,289] INFO sign:8797-1543565691.582 lineno:64 : do register to 127.0.0.1,remainder:['127.0.0.1']
[2018-11-30 16:14:54,771] INFO sign:8626-1543565671.974 lineno:110 : pull piece task result:{u'msg': u'client sucCount:2,cdn status:RUNNING,cdn sucCount:2', u'code': 602} and sleep 1.142 ...
[2018-11-30 16:14:54,916] INFO sign:8625-1543565671.979 lineno:110 : pull piece task result:{u'msg': u'client sucCount:0,cdn status:RUNNING,cdn sucCount:0', u'code': 602} and sleep 1.890 ...
[2018-11-30 16:14:55,921] INFO sign:8626-1543565671.974 lineno:60 : local http result:20971520 for path:/rate/ and cost:0.001
[2018-11-30 16:14:55,957] INFO sign:8626-1543565671.974 lineno:110 : pull piece task result:{u'msg': u'client sucCount:3,cdn status:RUNNING,cdn sucCount:3', u'code': 602} and sleep 1.064 ...
[2018-11-30 16:14:56,813] INFO sign:8625-1543565671.979 lineno:110 : pull piece task result:{u'msg': u'client sucCount:0,cdn status:RUNNING,cdn sucCount:0', u'code': 602} and sleep 1.013 ...
[2018-11-30 16:14:57,026] INFO sign:8626-1543565671.974 lineno:110 : pull piece task result:{u'msg': u'client sucCount:3,cdn status:RUNNING,cdn sucCount:3', u'code': 602} and sleep 1.577 ...
[2018-11-30 16:14:57,832] INFO sign:8625-1543565671.979 lineno:110 : pull piece task result:{u'msg': u'client sucCount:0,cdn status:RUNNING,cdn sucCount:0', u'code': 602} and sleep 1.743 ...
[2018-11-30 16:14:58,610] INFO sign:8626-1543565671.974 lineno:110 : pull piece task result:{u'msg': u'client sucCount:3,cdn status:RUNNING,cdn sucCount:3', u'code': 602} and sleep 1.613 ...
[...]

PR #277, opened 2018-12-24
fix: unreachable supernode ip in p2p-network

lowzj:

Make it possible for users to set the supernode's ip that we advertiese to other peers in the p2p-network, the property is: `advertieseIp`

```bash
java -Dsupernode.advertieseIp=10.1.1.1 -jar supernode.jar
```

By default, the first non-loop address of the host is advertied.

fixes: #252


Starnop:
LGTM.

PR #278, opened 2018-12-25
feature: update API definition to be more accurate

allencloud:

update API deinition to be more accurate

none

just definition of API, no need to test.

none

none


apis/swagger.yml
```diff
@@ -266,13 +271,20 @@ paths:
         - name: id
           in: path
           required: true
-          description: "ID of piece"
+          description: "ID of task"
           type: string
+        - name: num
```
Starnop: The field `num` means that the pieceNum of task, in the meantime, we use `num` as the request number of pieces of task in API ` /tasks/{id}/pieces`, shall we make a difference?

apis/swagger.yml
```diff
@@ -369,7 +381,10 @@ definitions:
     properties:
       ID:
         type: "string"
-        description: "Peer ID of dfget client. Every peer has a unique ID among peer network."
+        description: |
```
Starnop: `id` is not required to create a Peer, it should be returned by Supernode.

Starnop:
We need a collection of API related to `PeerTask`

allencloud:
> a collection of API related to PeerTask

I think `PeerTask` is a struct in supernode, why to expose this in API? @Starnop

apis/swagger.yml
allencloud: What is the num format like? Could you give us an example? @lowzj

Starnop:
> > a collection of API related to PeerTask
> 
> I think `PeerTask` is a struct in supernode, why to expose this in API? @Starnop

A `PeerTask` means a download process, for operation usage, we need to know which download processes are executing, the processes status, and so on. So we need an API to **get/list** `PeerTasks`. As for **create/update/delete** operations, we should handle them with `Task` related interfaces.

apis/swagger.yml
Starnop: We can use `pieceRange` here to identify a piece, and the type is string , an example: `0-45565`.

Starnop:
We need to update the struct of pieceInfo which include the following fields：
1. pieceRange string
2. pieceNum int
3. pieceSize int32
4. pieceMd5 string
5. cid string
6. peerIP string
7. peerPort int
8. path string
9. status string

@lowzj anything else should be added ?

apis/swagger.yml
lowzj: `pieceRange` is better. `Range` is a http header, for example: `Range: bytes=0-45565`.

lowzj:
> We need to update the struct of pieceInfo which include the following fields：
> 
> 1. pieceRange string
> 2. pieceNum int
> 3. pieceSize int32
> 4. pieceMd5 string
> 5. cid string
> 6. peerIP string
> 7. peerPort int
> 8. path string
> 9. status string
> 
> @lowzj anything else should be added ?

The `PieceUpdateRequest` need a parameter `srcCid`.

apis/swagger.yml
```diff
@@ -266,13 +271,20 @@ paths:
...
+            the range number of specific piece in the task.
+          type: integer
         - name: "PieceUpdateRequest"
           in: "body"
-          description: "request body which contains task update information"
+          description: |
+            request body which contains task update information.
           schema:
-            $ref: "#/definitions/PieceUpdateRequest"
+            $ref: "#/definitions/PieceInfo"
```
lowzj: Use `PieceUpdateRequest` instead of `PieceInfo`, the properties are:
```golang
// the downloader cid
srcCid string,
// the uploader cid
dstCid string,
// the peer task request status: START(700), RUNNING(701), FINISH(702)
status int,
// the peer task request result: FAIL(500), SUCCESS(501), INVALID(502), SEMISUC(503)
result int,
```

Starnop:
LGTM.

allencloud:
I still need to do some update. Please do not merge this now.

Starnop:
Can we move forward? ^_^

allencloud:
I have updated all the demand we discussed.  PTAL @Starnop @lowzj

Starnop:
> > We need to update the struct of pieceInfo which include the following fields：
> > 
> > 1. pieceRange string
> > 2. pieceNum int
> > 3. pieceSize int32
> > 4. pieceMd5 string
> > 5. cid string
> > 6. peerIP string
> > 7. peerPort int
> > 8. path string
> > 9. status string
> > 
> > @lowzj anything else should be added ?
> 
> The `PieceUpdateRequest` need a parameter `srcCid`.

@lowzj  Maybe we don't need the field of `status`, WDYT?

lowzj:
> > > ...
> > > 9. status string
> > > 
> > > @lowzj anything else should be added ?
> > 
> > 
> > The `PieceUpdateRequest` need a parameter `srcCid`.
> 
> @lowzj Maybe we don't need the field of `status`, WDYT?

agree.

allencloud:
> Should we unify the status as string type. Now the type of status passed betweendfget and supernode is int.

Yes. I think we should use string. @lowzj 
I think in the swagger, it has already been string.

lowzj:
LGTM

PR #280, opened 2018-12-26
fix: remove redundant log and output correct file length

lowzj:

The `dfget` will output twice result message and the downloaded file's length is not correct.

```bash
--2018-12-26 19:51:37--  https://www.taobao.com
dfget version:0.3.0
workspace:/Users/zj/.small-dragonfly sign:51733-1545825097.015
client:127.0.0.1 connected to node:127.0.0.1
start download by dragonfly
download SUCCESS(0) cost:0.052s length:-1 reason:0
download SUCCESS(0) cost:0.052s length:-1 reason:0
```

### DIFF
--- a/cmd/dfget/app/root.go
+++ b/cmd/dfget/app/root.go
@@ -35,7 +35,6 @@ var rootCmd = &cobra.Command{
 		if err != nil {
 			os.Exit(err.Code)
 		}
-		util.Printer.Println(resultMsg(cfg.Ctx, time.Now(), err))
 		os.Exit(0)
 	},
 }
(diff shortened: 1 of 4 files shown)


Starnop:
Through offline communication, the value of `ctx.RV.FileLength` returned by [supernode](https://github.com/dragonflyoss/dragonfly/blob/8b18c016e815ded09e0ad3ed39c72e3992ff14de/dfget/core/core.go#L141). However, Supernode gets the value from HTTP header of the source file and returns -1 if the value does not exist.  So we get the incorrect fileLength here.

In addition, if the supernode returns without downloading the file, it will not be able to get this value, So we should double check that `fileLength` is correct by `dfget`.

Starnop:
LGTM.

PR #286, opened 2018-12-29
refactor: update context to config in dfget

Starnop:

In general, we use ctx as the variable name of package `context` of golang, in the meantime, we don't have to create a global variable for config, passing it by argument is enough. So this pull request has dong two things as follows:
1. update the name `ctx` -> `cfg`
2. delete the global variable of `Context`

None.

updated.

After merged, we can use go `context` as normal.

### DIFF
--- a/cmd/dfget/app/root.go
+++ b/cmd/dfget/app/root.go
@@ -8,7 +8,7 @@ import (
 	"strings"
 	"time"
 
-	cfg "github.com/dragonflyoss/Dragonfly/dfget/config"
+	"github.com/dragonflyoss/Dragonfly/dfget/config"
 	"github.com/dragonflyoss/Dragonfly/dfget/core"
 	"github.com/dragonflyoss/Dragonfly/dfget/errors"
 	"github.com/dragonflyoss/Dragonfly/dfget/util"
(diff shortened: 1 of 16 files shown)


allencloud:
I totally agree with the update. But I will review the pr later today.

cmd/dfget/app/root.go
```diff
@@ -23,118 +23,128 @@ var (
...
 	},
 }
 
 func init() {
 	initFlags()
 }
 
+// runDfget do some init operation.
+func runDfget(args []string) {
+	// initialize logger and get properties
```
fengzixu: If you want to add comments for your functions, please obey this rule:

https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

dfget/core/downloader/p2p_downloader.go
```diff
@@ -121,20 +121,20 @@ func (p2p *P2PDownloader) Run() error {
 				p2p.finishTask(response, clientWriter)
 				return nil
 			} else {
-				p2p.Ctx.ClientLogger.Warnf("request piece result:%v", response)
+				p2p.Cfg.ClientLogger.Warnf("request piece result:%v", response)
```
fengzixu: Same problem as above.

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

fengzixu:
LGTM

allencloud:
Thanks for your review. @fengzixu 
LGTM as well.

PR #295, opened 2019-01-03
refactor: api/types/*.go code review, modify functions with return va…

songshuone:
…lues

Signed-off-by: Wu Peng 1399094187@qq.com

### DIFF
--- a/apis/types/peer_create_request.go
+++ b/apis/types/peer_create_request.go
@@ -62,51 +62,38 @@ func (m *PeerCreateRequest) Validate(formats strfmt.Registry) error {
 	return nil
 }
 
-func (m *PeerCreateRequest) validateIP(formats strfmt.Registry) error {
+func (m *PeerCreateRequest) validateIP(formats strfmt.Registry) (err error) {
 
 	if swag.IsZero(m.IP) { // not required
-		return nil
+		return
 	}
-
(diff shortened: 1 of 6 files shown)


apis/types/peer_create_request.go
```diff
@@ -62,51 +62,38 @@ func (m *PeerCreateRequest) Validate(formats strfmt.Registry) error {
 	return nil
 }
 
-func (m *PeerCreateRequest) validateIP(formats strfmt.Registry) error {
```
fengzixu: @songshuone Please notice the comment at the top of this file.

```
 // This file was generated by the swagger tool.
 // This file was generated by the swagger tool.
 // Editing this file might prove futile when you re-run the swagger generate command
```
It seems like that we shouldn't modify it.

apis/types/peer_info.go
```diff
@@ -66,51 +66,36 @@ func (m *PeerInfo) Validate(formats strfmt.Registry) error {
 	return nil
 }
 
-func (m *PeerInfo) validateIP(formats strfmt.Registry) error {
```
fengzixu: Same problem as above

allencloud:
Thanks a lot for your work. While actually all the files in folder apis/types are auto-generated via swagger. see doc https://github.com/dragonflyoss/Dragonfly/blob/master/apis/README.md.

So I do not think we can accept this PR. While feel free to contact us if you have any questions. @songshuone

PR #298, opened 2019-01-03
fix: make invalid link in apis.md pass markdown-link-check

allencloud:
docs: auto generate Dragonfly cli/api docs via code

We found that when we added some invalid links in swagger.yml, the generated markdown documents will contain the invalid link, then the [markdown-link-check](https://github.com/tcort/markdown-link-check) tool will not pass the CI to show the failure:

```
FILE: ./docs/apis.md
[✖] http://a.b.com/fileA?user=xxx&amp;auth=yyy
[✖] http://a.b.com/fileA

ERROR: dead links found!
```

This pr tries to fix the invalid link checking to use a link trimming the prefix of `http(s)://`.

no need.

none

none


Starnop:
https://github.com/tcort/markdown-link-check/blob/09952686ad865d379383b8f34e4638f977edb951/README.md#config-file-format

Maybe you don't have to change all the links, we can ignore the link check with pattern "http://a.b.com/".

allencloud:
If we need to ignore the link check with the pattern, do we have to provide a config json file for it? @Starnop 
If the answer is yes, I think it is not a good way to add a config json file in this repo. Because it is only for CI.

Starnop:
I'm afraid it's true so far, So let's do it in this way firstly, and then we'll give the community feedback and even fix it.

Starnop:
LGTM.

allencloud:
Thanks for your review. @Starnop 
Merging...

PR #305, opened 2019-01-07
refactor: refactor dfdaemon cmd part with by spf13/cobra

### LINKED ISSUE #249: [feature request] use `spf13/cobra` to refactor dfget/dfdaemon's Command Line part
## Why you need it?

[spf13/cobra](https://github.com/spf13/cobra) is both a library for creating powerful modern CLI applications as well as a program to generate applications and command files.

We need to use cobra to help construct CLI's flag, subcommand, doc auto-generation and something else.

With cobra, the cli code would be much more clear. And cobra will encapsulate all details about the cli, but provide more concrete and clear programming interface for developers.

Many successful projects are also using it, Kubernetes, etcd, moby, PouchContainer ...

## How it could be?

none

## Other related information

none

Starnop:

1. Use`spf13/cobra` to implement the dfcaemon cmd part.
2. Use `rootCmd.Flags()` instead of `rootCmd.PersistentFlags()` with `dfget`, more info about `PersistentFlags` please refer to https://github.com/spf13/cobra#persistent-flags

fixes #249

not needed.


Starnop:
@fengzixu PTAL, THX. 😄

cmd/dfdaemon/main.go
```diff
@@ -15,40 +15,97 @@
...
+	// if CommandLine.MaxProcs <= 0, programs run with GOMAXPROCS set to the number of cores available
+	if opt.MaxProcs > 0 {
+		runtime.GOMAXPROCS(opt.MaxProcs)
+	}
+
+	logrus.Infof("start dfdaemon param:%+v", opt)
+
+	var err error
+	if opt.CertFile != "" && opt.KeyFile != "" {
+		fmt.Printf("launch dfdaemon on port:%d with https server\n", opt.Port)
```
allencloud: Why to use `fmt.Printf`?
I think if we have set the logging config, we could ignore `fmt.Printf`. WDYT? @Starnop

cmd/dfdaemon/main.go
```diff
@@ -15,40 +15,97 @@
...
+		},
+	}
 
-	initializer.Init(options)
+	setupFlags(rootCmd, opt)
 
-	// if CommandLine.MaxProcs <= 0, programs run with GOMAXPROCS set to the number of cores available
-	if options.MaxProcs > 0 {
-		runtime.GOMAXPROCS(options.MaxProcs)
+	if err := rootCmd.Execute(); err != nil {
```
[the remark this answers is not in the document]
fengzixu: You can refer to the dfget.

cmd/dfdaemon/main.go
```diff
@@ -15,40 +15,97 @@
...
+				// assume the dfget binary is at the same directory as this daemon.
+				opt.DfPath = filepath.Dir(absPath) + "/dfget"
+			}
+		}
+	}
+
+	initializer.Init(opt)
+
+	// if CommandLine.MaxProcs <= 0, programs run with GOMAXPROCS set to the number of cores available
+	if opt.MaxProcs > 0 {
```
fengzixu: The logic doesn't accord with the comment.

cmd/dfdaemon/main.go
```diff
@@ -15,40 +15,97 @@
...
+		runtime.GOMAXPROCS(opt.MaxProcs)
+	}
+
+	logrus.Infof("start dfdaemon param:%+v", opt)
```
fengzixu: I think we should obey the rule about errorString when we want to output some info.

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

cmd/dfdaemon/main.go
```diff
@@ -15,40 +15,97 @@
...
+
+	var err error
+	if opt.CertFile != "" && opt.KeyFile != "" {
+		fmt.Printf("launch dfdaemon on port:%d with https server\n", opt.Port)
+		err = http.ListenAndServeTLS(fmt.Sprintf(":%d", opt.Port),
+			opt.CertFile, opt.KeyFile, nil)
 	} else {
-		err = http.ListenAndServe(fmt.Sprintf(":%d", options.Port), nil)
+		fmt.Printf("launch dfdaemon on port:%d with http server\n", opt.Port)
+		err = http.ListenAndServe(fmt.Sprintf(":%d", opt.Port), nil)
```
fengzixu: I think we should give the clear IP addr, such as 0.0.0.0 or 127.0.0.1.

cmd/dfdaemon/main.go
fengzixu: Yeap. Using log tool is better.

cmd/dfdaemon/main.go
```diff
@@ -15,40 +15,97 @@
...
+	fs.StringVar(&o.DfPath, "dfpath", "", "dfget path")
+	fs.StringVar(&o.RateLimit, "ratelimit", "", "net speed limit,format:xxxM/K")
+	fs.StringVar(&o.URLFilter, "urlfilter", "Signature&Expires&OSSAccessKeyId", "filter specified url fields")
+	fs.StringVar(&o.Registry, "registry", "https://index.docker.io", "registry addr(https://abc.xx.x or http://abc.xx.x) and must exist if dfdaemon is used to mirror mode")
+	fs.StringVar(&o.DownRule, "rule", "", "download the url by P2P if url matches the specified pattern,format:reg1,reg2,reg3")
+	fs.BoolVar(&o.Notbs, "notbs", true, "not try back source to download if throw exception")
+}
+
+// start to run dfdaemon server.
+func runDaemon() {
```
fengzixu: I think that we should move the operation about initing config into the single function and process it ahead of the runDaemon.

cmd/dfdaemon/main.go
```diff
@@ -15,40 +15,97 @@
...
-	"github.com/sirupsen/logrus"
-
 	"github.com/dragonflyoss/Dragonfly/cmd/dfdaemon/options"
 	"github.com/dragonflyoss/Dragonfly/dfdaemon/initializer"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+var (
```
fengzixu: `help` command should be implemented.

cmd/dfdaemon/main.go
Starnop: May be an example is better. THX.

cmd/dfdaemon/main.go
Starnop: As for `dfdaemon`, the default implementation is enough. 😄

cmd/dfdaemon/app/root.go
```diff
@@ -0,0 +1,107 @@
...
+
+	initializer.Init(opt)
+
+	// if Options.MaxProcs <= 0, programs run with GOMAXPROCS set to the number of cores available.
+	if opt.MaxProcs > 0 {
+		runtime.GOMAXPROCS(opt.MaxProcs)
+	}
+}
+
+// Execute will process dfget.
```
fengzixu: Hah, please revise here. 😆

fengzixu: Except for here, LGTM

LGTM

Starnop: So careless. 😓

cmd/dfdaemon/app/root.go
```diff
@@ -0,0 +1,107 @@
...
+var rootCmd = &cobra.Command{
+	Use:   "dfdaemon",
+	Short: "The dfdaemon is a proxy that intercepts image download requests.",
+	Long:  "The dfdaemon is a proxy between container engine and registry used for pulling images.",
+	Run: func(cmd *cobra.Command, args []string) {
+		runDaemon()
+	},
+}
+
+func init() {
```
zhouhaibing089: Nit: I do not see any problem if we put `setupFlags` right before `runDaemon` at L28, or is there any? (If we do not have to use `init`, we don't).

cmd/dfdaemon/app/root.go
```diff
@@ -0,0 +1,107 @@
...
+var (
+	opt           = &options.Options{}
+	defaultDFRepo = filepath.Join(os.Getenv("HOME"), ".small-dragonfly/dfdaemon/data/")
+)
+
+var rootCmd = &cobra.Command{
+	Use:   "dfdaemon",
+	Short: "The dfdaemon is a proxy that intercepts image download requests.",
+	Long:  "The dfdaemon is a proxy between container engine and registry used for pulling images.",
+	Run: func(cmd *cobra.Command, args []string) {
```
zhouhaibing089: Can we use `RunE` which can propagate errors to `Execute`?

cmd/dfdaemon/options/options.go
```diff
@@ -1,12 +1,5 @@
 package options
```
zhouhaibing089: can we move this package to `cmd/dfdaemon/app/options`?

cmd/dfdaemon/app/root.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// dfget download config
+	fs.StringVar(&o.DFRepo, "localrepo", defaultDFRepo, "temp output dir of dfdaemon")
+	fs.StringVar(&o.CallSystem, "callsystem", "com_ops_dragonfly", "caller name")
+	fs.StringVar(&o.DfPath, "dfpath", "", "dfget path")
+	fs.StringVar(&o.RateLimit, "ratelimit", "", "net speed limit,format:xxxM/K")
+	fs.StringVar(&o.URLFilter, "urlfilter", "Signature&Expires&OSSAccessKeyId", "filter specified url fields")
+	fs.StringVar(&o.Registry, "registry", "https://index.docker.io", "registry addr(https://abc.xx.x or http://abc.xx.x) and must exist if dfdaemon is used to mirror mode")
+	fs.StringVar(&o.DownRule, "rule", "", "download the url by P2P if url matches the specified pattern,format:reg1,reg2,reg3")
+	fs.BoolVar(&o.Notbs, "notbs", true, "not try back source to download if throw exception")
```
zhouhaibing089: Instead of doing all of this here, can we call `opt.AddFlags(` instead? The flag binding should reside in `options` package ideally.

cmd/dfdaemon/app/root.go
fengzixu: Can you explain the reason why we should avoid using `init`?

cmd/dfdaemon/app/root.go
Starnop: Yeah, got it.

cmd/dfdaemon/options/options.go
Starnop: Done.

cmd/dfdaemon/app/root.go
Starnop: Done.

cmd/dfdaemon/app/root.go
Starnop: Maybe it's not bad to use `init` according to [cobra example](https://github.com/spf13/cobra/blob/master/cobra/cmd/root.go#L42).

zhouhaibing089: Well, I am not saying we should avoid using `init`, but should avoid abusing it. Basically, `init` is used for side effects(plugin registration, global initialization, etc). Another argument could be: What if we do not want to run this piece of code in unit tests(I guess we certainly do not want to run this code in uni tests)?

allencloud: I think both are OK. While the way @zhouhaibing089 suggested is much straight-forward.

cmd/dfdaemon/options/options_test.go
```diff
@@ -1,58 +0,0 @@
-package options
```
zhouhaibing089: Could we move this to `cmd/dfdaemon/app/options`, too? (rather than delete it)

allencloud:
LGTM

zhouhaibing089:
Ah, Could we add back the unit test for `cmd/dfdaemon/app/options`?

Issue #306, opened 2019-01-07
[feature request] use "spf13/cobra" to auto generate command line markdown document


allencloud:
## Why you need it?

How to generate user-friendly document is very important work we have to do in Dragonfly project. While I think the best way to make it is to follow the rule **CODE IS DOCUMENT(代码即文档)**. Under such kind of policy, when every update occurs to the CLI-related code, the project has abilities to auto generate the corresponding document. Then the developers can spend less time to construct the document, instead they could have more time to polish the CLI related code when programming which is also reasonable.

We have already supported using `spf13/cobra` to implement COMMAND LINE framework in https://github.com/dragonflyoss/Dragonfly/pull/257. With this work, we can use https://github.com/spf13/cobra/blob/master/doc/md_docs.md to add another command to generate all the CLI document.

After adding the command which is used to generate CLI document, we can make @pouchrobot to automatically generate doc every day when it has update to the cli part. This is quite similar to [PouchContainer](https://github.com/alibaba/pouch).

## How it could be?

use "spf13/cobra" to auto generate command line markdown document

## Other related information

none

asadsm:
We have a cluster of machines and we're using NFS to make sure each machine in the cluster can access data from across the cluster. The problem is that we don't want to overload the network bus so in some cases, it's better to copy over data to local disk from the cluster to avoid network congestion. If DragonFly could provide this functionality, of P2P copy where each machine can act as the source and sink of the data it would be able to provide an excellent alternative to Bittorrent.

allencloud:
Could you move the comment to issue #219 , sorry I mistake the link just now. In addition, could you comment under issue #219, like the following way:

```
Orgnization: Alibaba
Location: Hangzhou, China
Contact: abc@d.com
Scenario: We have a cluster of machines and we're using NFS to make sure each machine in the cluster can access data from across the cluster. The problem is that we don't want to overload the network bus so in some cases, it's better to copy over data to local disk from the cluster to avoid network congestion. If DragonFly could provide this functionality, of P2P copy where each machine can act as the source and sink of the data it would be able to provide an excellent alternative to Bittorrent.
```

PR #257, opened 2018-12-10
refactor: use cobra to implement the dfget client

fengzixu:
Signed-off-by: fengzixu <hnustphoenix@gmail.com>

Related to: https://github.com/dragonflyoss/Dragonfly/issues/249

### Ⅰ. Describe what this PR did

Using cobra to implement the dfget client

### Ⅱ. Does this pull request fix one issue?

https://github.com/dragonflyoss/Dragonfly/issues/249

### Ⅲ. Why don't you add test cases (unit test/integration test)? (你真的觉得不需要加测试吗？)
I will add ut before the sign "WIP" was removed.

### Ⅳ. Describe how to verify it

I run three command:

1. `./dfget -u 'https://raw.githubusercontent.com/dragonflyoss/Dragonfly/master/docs/images/logo/dragonfly-linear.png'    ` Downloading files successfully.
2. `./dfget version` Showing the version of dfget.
3.  `./dfget`  Showing the message about help.

### Ⅴ. Special notes for reviews

There are two problems I want to discuss with developers in the community:

1. When I use cobra to implement the dfget client in cmd/dfget, should I remove the /dfget/cli ? Or import /dfget/cli in cmd/dfget/root.go and use the format cli.xxx to use the function about dfget client.
2. Whether existing approaches which implemented the dfget client in this PR is OK ? If it is accessible, I will add ut for it and delete useless code in /dfget/cli

PR #312, opened 2019-01-07
feature: add gen-doc command to auto gen dfget cli docs

allencloud:

How to generate user-friendly document is very important work we have to do in Dragonfly project. While I think the best way to make it is to follow the rule **CODE IS DOCUMENT(代码即文档)**. Under such kind of policy, when every update occurs to the CLI-related code, the project has abilities to auto generate the corresponding document. Then the developers can spend less time to construct the document, instead they could have more time to polish the CLI related code when programming which is also reasonable.

We have already supported using `spf13/cobra` to implement COMMAND LINE framework in https://github.com/dragonflyoss/Dragonfly/pull/257. With this work, we can use https://github.com/spf13/cobra/blob/master/doc/md_docs.md to add another command to generate all the CLI document.

After adding the command which is used to generate CLI document, we can make @pouchrobot to automatically generate doc every day when it has update to the cli part. This is quite similar to [PouchContainer](https://github.com/alibaba/pouch).

fix https://github.com/dragonflyoss/Dragonfly/issues/306

no need 

none

none


allencloud:
After PR https://github.com/pouchcontainer/pouchrobot/pull/100 is merged, we could make @pouchrobot to support auto generate cli documentation.

Starnop:
Please rebase the master branch after [PR](https://github.com/dragonflyoss/Dragonfly/pull/305) merged.

allencloud:
PTAL @Starnop

cmd/dfget/app/gen_doc.go
```diff
@@ -0,0 +1,39 @@
...
+		SilenceErrors: true,
+		SilenceUsage:  true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return genDocCommand.runGenDoc(args)
+		},
+	}
+	rootCmd.AddCommand(genDocCommand.cmd)
+}
+
+func (g *GenDocCommand) runGenDoc(args []string) error {
```
Starnop: Could you please implement the related part of `dfdaemon` by he way?

allencloud: Yes, I will do that in a separate PR.

docs/cli_reference/dfget.md
```diff
@@ -1,83 +1,43 @@
...
-  --version, -v         version
-  --node NODE, -n NODE  specify nodes
-  --console             show log on console
-  --header HEADER       http header, e.g. --header="Accept: *" --header="Host:
-                        abc"
-  --dfdaemon            caller is from df-daemon
+dfget [flags]
 ```
 
-## Files
```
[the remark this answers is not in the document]
allencloud: Yes, we can split that from the dfget.md.

allencloud:
I am wondering if we could add a new subcommand of `dfget pull` for dfget. Then it would be much clearer for the usage of dfget. @lowzj @Starnop

PR #321, opened 2019-01-12
bugfix: update rootCmd of dfget from PersistentFlags to Flags

Starnop:

In fact, we should use `Flags()` instead of `PersistentFlags()`. A 'persistent flag' means that this flag will be available to the command it's assigned to as well as every command under that command. For more info about `PersistentFlags`, please refer to https://github.com/spf13/cobra#persistent-flags

None.

not needed.


Starnop:
Just a small change, PTAL. @allencloud

PR #326, opened 2019-01-15
refactor: client launches an independent peer server when downloading files in p2p pattern

lowzj:

* client launches an independent peer server when downloading files in p2p pattern
* add `server` command to `dfget`

### DIFF
--- a/cmd/dfget/app/server.go
+++ b/cmd/dfget/app/server.go
@@ -18,17 +18,26 @@ package app
 
 import (
 	"fmt"
+	"path/filepath"
 
-	"github.com/dragonflyoss/Dragonfly/dfget/config"
+	"github.com/dragonflyoss/Dragonfly/dfget/core/uploader"
+	"github.com/dragonflyoss/Dragonfly/dfget/util"
 	"github.com/spf13/cobra"
 )
 
 var serverCmd = &cobra.Command{
-	Use:           "server",
-	Short:         "Launch a peer server to upload files.",
+	Use:   "server",
+	Short: "Launch a peer server for uploading files.",
 	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Printf("Server: %v\n", config.Ctx)
-		fmt.Printf("Server: %v\n", config.Ctx.RV.String())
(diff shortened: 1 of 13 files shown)


cmd/dfget/app/server.go
```diff
@@ -0,0 +1,60 @@
...
+	Short: "Launch a peer server for uploading files.",
+	Run: func(cmd *cobra.Command, args []string) {
+		logPath := filepath.Join(cfg.WorkHome, "logs")
+		cfg.ServerLogger = util.CreateLogger(logPath,
+			"dfserver.log", "INFO", cfg.Sign)
+		if port, err := uploader.LaunchPeerServer(cfg); err == nil {
+			fmt.Println(port)
+			uploader.WaitForShutdown()
+		} else {
+			fmt.Println(err.Error())
```
allencloud: I think we should exit with a non-0 exit code here, while `fmt.Println` will never return a non-zero exit code.

In addition, I wish that we could first check the case of `err != nil`. Because this could make the exception handling in an indent loop, but the main logic could remain in the main code block.

cmd/dfget/app/server.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func initServerFlags() {
+	serverCmd.PersistentFlags().StringVar(&cfg.RV.SystemDataDir, "data", cfg.RV.SystemDataDir,
+		"the directory which stores temporary files for p2p uploading")
+	serverCmd.PersistentFlags().StringVar(&cfg.WorkHome, "home", cfg.WorkHome,
+		"the work home of dfget server")
+	serverCmd.PersistentFlags().StringVar(&cfg.RV.LocalIP, "ip", "",
+		"the ip that server will listen on")
+	serverCmd.PersistentFlags().IntVar(&cfg.RV.PeerPort, "port", 0,
```
allencloud: Is it reasonable to config the default port to be 0?
How about a fixed port number?

lowzj: `0` means that `dfget server` will auto-generate a port and attempt to listen and serve on this port. It will retry at most 10 times if `dfget server` starts failed.

allencloud:
LGTM

PR #332, opened 2019-01-16
bugfix: write temporary file when cannot hard link

lowzj:

It  now doesn't write data to file correctly when create hard link failed.


Starnop:
LGTM.

PR #336, opened 2019-01-18
refator: use error instead of panic

### LINKED ISSUE #193: Should we use panic or return error?
## Question

Now, we use panic and recover to handle the error. However, `return error` can alse do that. This is a classic question in the `go` language. `Should we use panic or return error?`

There is a discussion [here ](https://stackoverflow.com/questions/44504354/should-i-use-panic-or-return-error) which may help us.

And I agree with the following sentence:

> Generally speaking, in go, if you're not sure, return an error. Panics should be pretty rare.

Just  a case as follows:
```
# dfget-go --url "http://test" --output ./resource.png --node "127.0.0.1"
panic: invalid url: http://test [recovered]
	panic: (*logrus.Entry) (0x78bbc0,0xc42008c870)

goroutine 1 [running]:
github.com/alibaba/Dragonfly/vendor/github.com/sirupsen/logrus.Entry.log(0xc42008c690, 0xc420081dd0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/tmp/dragonfly/build/src/github.com/alibaba/Dragonfly/vendor/github.com/sirupsen/logrus/entry.go:126 +0x318
github.com/alibaba/Dragonfly/vendor/github.com/sirupsen/logrus.(*Entry).Panic(0xc42008c820, 0xc420183d28, 0x1, 0x1)
	/tmp/dragonfly/build/src/github.com/alibaba/Dragonfly/vendor/github.com/sirupsen/logrus/entry.go:196 +0xaa
github.com/alibaba/Dragonfly/vendor/github.com/sirupsen/logrus.(*Logger).Panic(0xc42008c690, 0xc420055d28, 0x1, 0x1)
	/tmp/dragonfly/build/src/github.com/alibaba/Dragonfly/vendor/github.com/sirupsen/logrus/logger.go:244 +0x6d
github.com/alibaba/Dragonfly/dfget/config.AssertContext.func1(0xc420166000)
	/tmp/dragonfly/build/src/github.com/alibaba/Dragonfly/dfget/config/config.go:281 +0x87
panic(0x728f40, 0xc42007ff30)
	/usr/local/go/src/runtime/panic.go:502 +0x229
github.com/alibaba/Dragonfly/dfget/util.PanicIfError(0x7e17e0, 0xc42007ff10, 0x79e00b, 0xb)
	/tmp/dragonfly/build/src/github.com/alibaba/Dragonfly/dfget/util/util.go:77 +0x105
github.com/alibaba/Dragonfly/dfget/config.AssertContext(0xc420166000)
	/tmp/dragonfly/build/src/github.com/alibaba/Dragonfly/dfget/config/config.go:285 +0x121
github.com/alibaba/Dragonfly/dfget/cli.initialize()
	/tmp/dragonfly/build/src/github.com/alibaba/Dragonfly/dfget/cli/cli.go:50 +0x3b
github.com/alibaba/Dragonfly/dfget/cli.Run()
	/tmp/dragonfly/build/src/github.com/alibaba/Dragonfly/dfget/cli/cli.go:37 +0x26
main.main()
	/tmp/dragonfly/build/src/github.com/alibaba/Dragonfly/cmd/dfget/main.go:28 +0x20
```
When I use a invalid URL and get panic, it's not make sense for end user.

WDYT ?

Starnop:

I submitted this pull request for the following reasons: 

+ For a project with `Golang`, it's not suitable to use a lot of `panic` . 
+ In the mean time, it's not user-friendly to print the `Stack information` when something goes wrong.
+ For error handling, we should normalize that.
+ By the way, refactor the code which is not reasonable.

fixes  #193

updated.

### DIFF
--- a/cmd/dfget/app/root.go
+++ b/cmd/dfget/app/root.go
@@ -13,6 +13,7 @@ import (
 	"github.com/dragonflyoss/Dragonfly/dfget/errors"
 	"github.com/dragonflyoss/Dragonfly/dfget/util"
 
+	errHandler "github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
@@ -41,22 +42,34 @@ func init() {
 
 // runDfget do some init operations and start to download.
 func runDfget() error {
-	// initialize logger and get properties
-	initLog()
+	// initialize logger
+	if err := initClientLog(); err != nil {
+		util.Printer.Println(fmt.Sprintf("init log error: %v", err))
+		return err
+	}
+
+	// get config from property files
 	initProperties()
 
-	// check the legitimacy of parameters
+	if err := transParams(); err != nil {
+		util.Printer.Println(err.Error())
+		return err
+	}
+
 	checkParameters()
 	cfg.ClientLogger.Infof("get cmd params:%q", os.Args)
 
-	config.AssertConfig(cfg)
+	if err := config.AssertConfig(cfg); err != nil {
+		util.Printer.Println(fmt.Sprintf("assert context error: %v", err))
+		return err
+	}
 	cfg.ClientLogger.Infof("get init config:%v", cfg)
 
 	// enter the core process
 	err := core.Start(cfg)
 	util.Printer.Println(resultMsg(cfg, time.Now(), err))
 	if err != nil {
-		return err
+		os.Exit(err.Code)
 	}
 	return nil
 }
@@ -68,6 +81,7 @@ func checkParameters() {
 	}
 }
 
+// load config from property files.
 func initProperties() {
 	properties := config.NewProperties()
 	for _, v := range cfg.ConfigFiles {
@@ -94,17 +108,27 @@ func initProperties() {
 	if cfg.ClientQueueSize == 0 {
 		cfg.ClientQueueSize = properties.ClientQueueSize
 	}
+}
 
+// transParams trans the user-friendly parameter formats
+// to the format corresponding to the `Config` struct.
+func transParams() error {
 	cfg.Filter = transFilter(filter)
 
 	var err error
-	cfg.LocalLimit, err = transLimit(localLimit)
-	util.PanicIfError(err, "convert locallimit error")
-	cfg.TotalLimit, err = transLimit(totalLimit)
-	util.PanicIfError(err, "convert totallimit error")
+	if cfg.LocalLimit, err = transLimit(localLimit); err != nil {
+		return errHandler.Wrapf(errors.ErrConvertFailed, "locallimit: %v", err)
+	}
+
+	if cfg.TotalLimit, err = transLimit(totalLimit); err != nil {
+		return errHandler.Wrapf(errors.ErrConvertFailed, "totallimit: %v", err)
+	}
+
+	return nil
 }
 
-func initLog() {
+// initClientLog init client logger.
+func initClientLog() error {
 	var (
 		logPath  = path.Join(cfg.WorkHome, "logs")
 		logLevel = "info"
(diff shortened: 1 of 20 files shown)


dfget/config/config.go
```diff
@@ -240,35 +243,45 @@ func NewConfig() *Config {
...
-		cfg.WorkHome = path.Join(currentUser.HomeDir, ".small-dragonfly")
-		cfg.RV.MetaPath = path.Join(cfg.WorkHome, "meta", "host.meta")
-		cfg.RV.SystemDataDir = path.Join(cfg.WorkHome, "data")
-		cfg.RV.FileLength = -1
-	} else {
-		panic(fmt.Errorf("get user error: %s", err))
+	currentUser, err := user.Current()
+	if err != nil {
+		util.Printer.Println(fmt.Sprintf("get user error: %s", err))
+		os.Exit(CodeGetUserError)
```
lowzj: I don't suggest to use `os.Exit` in inner packages. It cannot recover any more and not easy to test.

dfget/core/core.go
```diff
@@ -86,13 +81,23 @@ func prepare(cfg *config.Config) (err error) {
 
 	rv.RealTarget = cfg.Output
 	rv.TargetDir = path.Dir(rv.RealTarget)
-	panicIf(util.CreateDirectory(rv.TargetDir))
+	if err := util.CreateDirectory(rv.TargetDir); err != nil {
```
lowzj: use `=` instead of `:=`

dfget/core/core.go
```diff
@@ -86,13 +81,23 @@ func prepare(cfg *config.Config) (err error) {
...
+		return err
+	}
 	cfg.RV.TempTarget, err = createTempTargetFile(rv.TargetDir, cfg.Sign)
-	panicIf(err)
+	if err != nil {
```
lowzj: It could be written in one line

dfget/core/core.go
```diff
@@ -115,21 +120,21 @@ func launchPeerServer(cfg *config.Config) (err error) {
 }
 
 func registerToSuperNode(cfg *config.Config, register regist.SupernodeRegister) (
```
lowzj: The `panics` in this function will be captured and ignore, the `p2p` pattern will be  degraded to `source` pattern, `dfget` will download the file from source server. But the new code returns a none-nil error to caller, this will break the downloading flow in `core.Start`.

dfget/errors/errors.go
```diff
@@ -47,3 +66,35 @@ type DFGetError struct {
...
+func IsNilError(err error) bool {
+	if err == nil {
+		return true
+	}
+	return false
+}
+
+// IsInvalidValue check the error is the value is invalid or not.
+func IsInvalidValue(err error) bool {
+	fmt.Printf("err: %v", err)
```
lowzj: delete this line

dfget/util/log_test.go
```diff
@@ -123,15 +113,17 @@ func (suite *DFGetUtilSuite) TestAddConsoleLog(c *check.C) {
...
+	logger, err := CreateLogger(tmpPath, tmpFileName, level, sign)
 	return logger, tmpFile, r, err
 }
 
-func cleanTempFile(file *os.File, err error) {
-	if err == nil {
-		file.Close()
-		os.Remove(file.Name())
-	}
+func cleanTempFile(file *os.File) {
```
lowzj: The file is `nil` when an error occurred.
It's better to retain the parameter `err` even though you use `c.Assert` to check it before calling this function. It requires less pre-conditions and keeps up robustness for other callers to use it conveniently.

dfget/config/config.go
Starnop: It is unreasonable to rely heavily on the user's home directory. We should make it possible to customize by parameters. WDYT? If that, I will finish it in the later pull request and  record a `TODO` task temporarily.

cmd/dfget/app/root.go
```diff
@@ -94,34 +108,58 @@ func initProperties() {
...
+// to the format corresponding to the `Config` struct.
+func transParams() error {
 	cfg.Filter = transFilter(filter)
 
-	var err error
-	cfg.LocalLimit, err = transLimit(localLimit)
-	util.PanicIfError(err, "convert locallimit error")
-	cfg.TotalLimit, err = transLimit(totalLimit)
-	util.PanicIfError(err, "convert totallimit error")
+	lLimit, err := transLimit(localLimit)
```
lowzj: use named return value, these codes could be written:

```go
if cfg.LocalLimit, err = transLimit(localLimit); err != nil {
    ...
}
```

Starnop: Done.

lowzj:
LGTM

PR #340, opened 2019-01-21
fix: ci passed but code checks and unit-test failed in actual

lowzj:
Signed-off-by: lowzj <zj3142063@gmail.com>

### Ⅰ. Describe what this PR did

The building scripts always return `0` even though they executes failed.

### Ⅱ. Does this pull request fix one issue?

related: #255

PR #343, opened 2019-01-21
fix failed unit tests

Starnop:

Previously, CI will looks good even if the unit test did not pass.  So there were a lot of tests that didn't actually pass before. And update them in this pull request. As for fix the bug of CI, please refer to #340 

None.

### DIFF
--- a/dfget/config/config.go
+++ b/dfget/config/config.go
@@ -261,12 +261,6 @@ func NewConfig() *Config {
 
 // AssertConfig checks the config and return errors.
 func AssertConfig(cfg *Config) (err error) {
-	defer func() {
-		if err != nil {
-			cfg.ClientLogger.Errorf("assert context error: %v", err)
-		}
-	}()
-
 	if util.IsNil(cfg) {
 		return errors.Wrap(errType.ErrNotInitialized, "runtime config")
 	}
@@ -280,7 +274,7 @@ func AssertConfig(cfg *Config) (err error) {
 	}
 
 	if err := checkOutput(cfg); err != nil {
-		return err
+		return errors.Wrapf(errType.ErrInvalidValue, "output: %v", err)
 	}
 	return nil
 }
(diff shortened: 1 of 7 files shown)


dfget/config/config.go
```diff
@@ -276,7 +270,7 @@ func AssertConfig(cfg *Config) (err error) {
 	}
 
 	if err := checkURL(cfg); err != nil {
-		return errors.Wrapf(errType.ErrInvalidValue, "url: %v", err)
+		return errors.Wrapf(errType.ErrInvalidValue, "url %+v", err)
 	}
 
 	if err := checkOutput(cfg); err != nil {
```
lowzj: We should wrap this error. I think that it's also an invalid value error.

dfget/config/config_test.go
```diff
@@ -95,16 +96,14 @@ func (suite *ConfigSuite) TestAssertConfig(c *check.C) {
...
 		checkFunc func(err error) bool
 	}{
 		{checkFunc: errors.IsNotInitialized},
-		{clog: clog, checkFunc: errors.IsNotInitialized},
-		{clog: clog, slog: clog, checkFunc: errors.IsInvalidValue},
-		{clog: clog, slog: clog, url: "http://a.b", checkFunc: errors.IsNilError},
-		{clog: clog, slog: clog, url: "http://a.b", output: "/root", checkFunc: errors.IsInvalidValue},
+		{clog: clog, checkFunc: errors.IsInvalidValue},
+		{clog: clog, url: "http://a", checkFunc: errors.IsInvalidValue},
+		{clog: clog, url: "http://a.b.com", output: "/tmp/output", checkFunc: errors.IsNilError},
```
lowzj: We should not change the output `/root`

Starnop:
All mentioned above has been updated. PTAL.

lowzj:
LGTM

PR #345, opened 2019-01-23
fix: don't write service file if downloading pattern isn't p2p

lowzj:

When downloading pattern is not `p2p`, `dfget` should not write service file for uploading, because they're useless.

This pull request makes `dfget` only write downloading data to target directory. It will reduce write disk times.


Starnop:
LGTM.

Starnop:
LGTM.

Issue #44, opened 2018-04-04
[bug] cannot use dfget to download an empty file


lowzj:
dfclient.log
```
2019-01-23 18:49:53.725 INFO sign:14434-1548240593.723 : get cmd params:["./bin/darwin_amd64/dfget" "-u" "http://127.0.0.1:8001/a.test" "-o" "/tmp/a.test" "--console"]
2019-01-23 18:49:53.726 INFO sign:14434-1548240593.723 : get init config:{"url":"http://127.0.0.1:8001/a.test","output":"/tmp/a.test","pattern":"p2p","node":["127.0.0.1"],"console":true,"clientQueueSize":6,"startTime":"2019-01-23T18:49:53.722690597+08:00","sign":"14434-1548240593.723","user":"zj","workHome":"/Users/zj/.small-dragonfly","configFile":["/etc/dragonfly.yaml","/etc/dragonfly.conf"]}
--2019-01-23 18:49:53--  http://127.0.0.1:8001/a.test
dfget version:0.3.0
workspace:/Users/zj/.small-dragonfly sign:14434-1548240593.723
2019-01-23 18:49:53.726 INFO sign:14434-1548240593.723 : target file path:/tmp/a.test
2019-01-23 18:49:53.727 INFO sign:14434-1548240593.723 : runtimeVariable: {"MetaPath":"/Users/zj/.small-dragonfly/meta/host.meta","SystemDataDir":"/Users/zj/.small-dragonfly/data","DataDir":"/Users/zj/.small-dragonfly/data","RealTarget":"/tmp/a.test","TargetDir":"/tmp","TempTarget":"/tmp/dfget-14434-1548240593.723.tmp-108158694","Cid":"127.0.0.1-14434-1548240593.723","TaskURL":"http://127.0.0.1:8001/a.test","TaskFileName":"a.test-14434-1548240593.723","LocalIP":"127.0.0.1","PeerPort":0,"FileLength":-1,"DataExpireTime":180000000000,"ServerAliveTime":300000000000}
2019-01-23 18:49:53.728 INFO sign:14434-1548240593.723 : local http result: err:dial tcp4 127.0.0.1:0: connect: can't assign requested address, port:0 path:/check/
2019-01-23 18:49:53.857 INFO sign:14434-1548240593.723 : local http result:a.test-14434-1548240593.723 err:<nil>, port:25801 path:/check/
[...]

PR #351, opened 2019-01-24
fix: [0.2.x] cannot use dragonfly to download an empty file

lowzj:

Dragonfly cannot download an empty file because that supernode don't write and cache data if the file length less and equals zero.

merge to branch: `0.2.x`

related: #44


chenchaobing:
LGTM

PR #353, opened 2019-01-25
feature: implement the rate limit feature

Starnop:

It's  an important feature to limit the download/upload speed for Dragonfly. More info about the advanced features about Dragonfly, please refer to [features](https://github.com/dragonflyoss/Dragonfly#features).

As you can see, dfget limits the speed with two parameters:
+ locallimit: rate limit about a single download task
+ totallimit: rate limit about the whole host
 
You can get more info about that [here](https://github.com/dragonflyoss/Dragonfly/blob/master/docs/cli_reference/dfget.md#options).

As for a download task, dfget will calculate the rate limit dynamically before download a piece from other peers **through the native uploader server**. Proceed as follows:
1. get locallimit
2. send a request to the **native uploader server** with: 
  + url: http://IP:Port/rate/{taskFileName}
  + header："rateLimit"=locallimit

As for a uploader server, the uploader will provide an API for parse rate limit. The behaviors of this handler are as follows:
1. Return the ratelimit obtained from the header when **totalLimit** less than or equals zero.
2. Calculate a new value based on all the download tasks and totalLimit based on the formula as follows:
  + Formula: `newRateLimit=(ratelimit*totalLimit+total-1)/total`

Where the `total` represents the sum of the rate limits of all download tasks of the native host.

None.

Added.

Pass all unit tests with CI.

### DIFF
--- a/dfget/core/downloader/p2p_downloader.go
+++ b/dfget/core/downloader/p2p_downloader.go
@@ -18,8 +18,10 @@ package downloader
 
 import (
 	"bytes"
+	"fmt"
 	"math/rand"
 	"os"
+	"strconv"
 	"time"
 
 	"github.com/dragonflyoss/Dragonfly/dfget/config"
@@ -51,7 +53,6 @@ type P2PDownloader struct {
 	pieceSizeHistory [2]int32
 	queue            util.Queue
 	clientQueue      util.Queue
-	writerDone       chan struct{}
 
 	clientFilePath  string
 	serviceFilePath string
@@ -62,6 +63,8 @@ type P2PDownloader struct {
 	// not in: the range hasn't been processed
 	pieceSet map[string]bool
 	total    int64
+
+	rateLimiter *util.RateLimiter
 }
 
 func (p2p *P2PDownloader) init() {
@@ -77,12 +80,13 @@ func (p2p *P2PDownloader) init() {
 	p2p.queue.Put(NewPieceSimple(p2p.taskID, p2p.node, config.TaskStatusStart))
 
 	p2p.clientQueue = util.NewQueue(config.DefaultClientQueueSize)
-	p2p.writerDone = make(chan struct{})
 
 	p2p.clientFilePath = helper.GetTaskFile(p2p.taskFileName, p2p.Cfg.RV.DataDir)
 	p2p.serviceFilePath = helper.GetServiceFile(p2p.taskFileName, p2p.Cfg.RV.DataDir)
 
 	p2p.pieceSet = make(map[string]bool)
+
+	p2p.rateLimiter = util.NewRateLimiter(int32(p2p.Cfg.LocalLimit), 2)
 }
 
 // Run starts to download the file.
@@ -203,8 +207,37 @@ func (p2p *P2PDownloader) pullPieceTask(item *Piece) (
 	return res, err
 }
 
-func (p2p *P2PDownloader) pullRate(data *types.PullPieceTaskResponseContinueData) {
+// getPullRate get download rate limit dynamically.
+func (p2p *P2PDownloader) getPullRate(data *types.PullPieceTaskResponseContinueData) {
+	var localRate int
 
+	if p2p.Cfg.LocalLimit > 0 {
+		localRate = p2p.Cfg.LocalLimit
+	} else {
+		localRate = data.DownLink * 1024
+	}
+
+	// Calculate the download speed limit
+	// that the current download task can be assigned
+	// by the uploader server.
+	// TODO: Reduce the frequency of requests.
+	url := fmt.Sprintf("http://%s:%d%s%s", p2p.Cfg.RV.LocalIP, p2p.Cfg.RV.PeerPort, config.LocalHTTPPathRate, p2p.taskFileName)
+	headers := make(map[string]string)
+	headers["rateLimit"] = strconv.Itoa(localRate)
+	resp, err := util.Do(url, headers, util.DefaultTimeout)
+	if err != nil {
+		p2p.Cfg.ClientLogger.Errorf("failed to pullRate: %v", err)
+		p2p.rateLimiter.SetRate(util.TransRate(localRate))
+		return
+	}
+
+	reqRate, err := strconv.Atoi(resp)
+	if err != nil {
+		p2p.Cfg.ClientLogger.Errorf("failed to parse rate from resp %s: %v", resp, err)
+		p2p.rateLimiter.SetRate(util.TransRate(localRate))
+		return
+	}
+	p2p.rateLimiter.SetRate(util.TransRate(reqRate))
 }
 
 func (p2p *P2PDownloader) startTask(data *types.PullPieceTaskResponseContinueData) {
@@ -215,6 +248,7 @@ func (p2p *P2PDownloader) startTask(data *types.PullPieceTaskResponseContinueDat
 		cfg:         p2p.Cfg,
 		queue:       p2p.queue,
 		clientQueue: p2p.clientQueue,
+		rateLimiter: p2p.rateLimiter,
 	}
 	powerClient.Run()
 }
@@ -296,7 +330,7 @@ func (p2p *P2PDownloader) processPiece(response *types.PullPieceTaskResponse,
 		}
 		if !ok {
 			p2p.pieceSet[pieceRange] = false
-			p2p.pullRate(pieceTask)
+			p2p.getPullRate(pieceTask)
 			go p2p.startTask(pieceTask)
 			hasTask = true
 		}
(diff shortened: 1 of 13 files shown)


dfget/core/downloader/p2p_downloader.go
```diff
@@ -203,18 +205,46 @@ func (p2p *P2PDownloader) pullPieceTask(item *Piece) (
...
+		localRate = p2p.Cfg.LocalLimit
+	} else {
+		localRate = data.DownLink * 1024
+	}
+
+	// Calculate the download speed limit
+	// that the current download task can be assigned
+	// by the uploader server.
+	// TODO: Reduce the frequency of requests.
+	url := fmt.Sprintf("http://%s:%d%s%s", p2p.Cfg.RV.LocalIP, p2p.Cfg.RV.PeerPort, config.LocalHTTPPathRate, p2p.taskFileName)
```
lowzj: How about to create an interface of dfget http server just like `SupernodeAPI`? It makes the code more readable and reusable.

dfget/core/downloader/power_client.go
```diff
@@ -59,54 +62,66 @@ func (pc *PowerClient) Run() (err error) {
...
-			return err
-		}
-		// TODO handle read timeout
+	// send download request
+	url := fmt.Sprintf("http://%s:%d%s", dstIP, peerPort, pc.pieceTask.Path)
+	headers := make(map[string]string)
+	headers["Range"] = pc.pieceTask.Range
+	headers["pieceNum"] = strconv.Itoa(pc.pieceTask.PieceNum)
+	headers["pieceSize"] = fmt.Sprint(pc.pieceTask.PieceSize)
+	startTime := time.Now().Unix()
```
lowzj: just use:
```go
startTime := time.Now()
```

calculate the time cost:
```go
time.Now().Sub(startTime).Seconds()
```

dfget/core/uploader/uploader.go
```diff
@@ -401,33 +410,85 @@ func (ps *peerServer) uploadHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	// define a function that Range will call it sequentially
+	// for each key and value present in the map
+	f := func(key, value interface{}) bool {
+		task, ok := value.(*taskConfig)
+		if !ok {
+			return true
+		}
+		total += task.rateLimit
+		return true
```
lowzj:
```go
		if task, ok := value.(*taskConfig); ok {
			total += task.rateLimit
		}
		return true
```

dfget/core/uploader/uploader_util.go
```diff
@@ -47,6 +48,19 @@ func getTaskFile(taskFileName string) (*os.File, error) {
...
 	taskPath := helper.GetServiceFile(taskFileName, tc.dataDir)
+
+	// Pre-check the length of the file as expected.
+	fileInfo, err := os.Stat(taskPath)
+	if err != nil {
+		return nil, err
+	}
+	fileSize := fileInfo.Size()
+	fmt.Println(fileSize)
+	fmt.Printf("%+v", params)
```
lowzj: delete these 2 lines

dfget/core/uploader/uploader_util.go
```diff
@@ -47,6 +48,19 @@ func getTaskFile(taskFileName string) (*os.File, error) {
...
+	if fileSize-params.start < params.pieceLen {
```
lowzj: The condition here is not sufficient, such as this case:
```
fileSize = 10
params.sart = -2
params.pieceLen = 11
```

Starnop: In fact, we will check the legality of range before that. 

https://github.com/dragonflyoss/Dragonfly/blob/master/dfget/core/uploader/uploader_util.go#L59

dfget/core/downloader/p2p_downloader.go
Starnop: May be do that with the later pull request? 
At the same time, I also want to define some constants  about http headers, etc.

dfget/core/downloader/power_client.go
```diff
@@ -59,54 +61,61 @@ func (pc *PowerClient) Run() (err error) {
...
 	}
 
-	piece := NewPiece(pc.taskID, pc.node, pc.pieceTask.Cid, pc.pieceTask.Range, config.ResultFail, config.TaskStatusRunning)
+	// TODO handle read timeout
+	readFinish := time.Now()
+
+	// Verify md5 code
+	realMd5 := limitReader.Md5()
+	if realMd5 != pieceMD5 {
+		pc.cfg.ClientLogger.Errorf("piece range:%s error,realMd5:%s,expectedMd5:%s,dstIp:%s,total:%d", pc.pieceTask.Range, realMd5, pieceMD5, dstIP, total)
```
lowzj: It will be logged in defer scope again.

dfget/core/downloader/p2p_downloader.go
```diff
@@ -203,18 +203,45 @@ func (p2p *P2PDownloader) pullPieceTask(item *Piece) (
...
 	powerClient := &PowerClient{
 		taskID:      p2p.taskID,
 		node:        p2p.node,
 		pieceTask:   data,
 		cfg:         p2p.Cfg,
 		queue:       p2p.queue,
 		clientQueue: p2p.clientQueue,
+		rateLimit:   rateLimit,
 	}
 	powerClient.Run()
```
lowzj: How about does `PowerClient.Run` return `Piece` and `error`? Then we don't need to pass `queue` and `clientQueue` to `PowerClient`. We can process them in this function according to the result of `PowerClient.Run`.

dfget/core/uploader/uploader_util.go
```diff
@@ -104,14 +116,19 @@ func transFile(f *os.File, w http.ResponseWriter, start, readLen int64) error {
 			return nil
 		}
 
-		if int64(num) > remain {
-			w.Write(buf[:remain])
-		} else {
-			w.Write(buf[:num])
+		length := int64(num)
```
lowzj: * remove line `// TODO: limit the read rate.`
* these codes don't support to upload empty file:
```go
		if num == 0 {
			if total == 0 {
				return fmt.Errorf("content is empty")
			}
			return nil
		}
```

dfget/util/http_util.go
```diff
@@ -114,6 +115,36 @@ func Get(url string, timeout time.Duration) (int, []byte, error) {
...
+// Do performs the given http request and fills the given http response.
+// When timeout <= 0, it will block until receiving response from server.
+func Do(url string, headers map[string]string, timeout time.Duration) (string, error) {
+	// init request and response
+	req := fasthttp.AcquireRequest()
+	req.SetRequestURI(url)
+	for k, v := range headers {
+		req.Header.Add(k, v)
+	}
+	resp := fasthttp.AcquireResponse()
```
lowzj: release the `req` and `resp`:
```go
req := fasthttp.AcquireRequest()
defer fasthttp.ReleaseRequest(req)
...
resp := fasthttp.AcquireResponse()
defer fasthttp.ReleaseResponse(resp)
```

dfget/core/downloader/p2p_downloader.go
lowzj: ok

dfget/core/downloader/power_client.go
```diff
@@ -54,59 +56,65 @@ func (pc *PowerClient) Run() (err error) {
 
 	defer func() {
 		if err != nil {
-			pc.cfg.ClientLogger.Errorf("read piece cont error:%s from dst:%s", err, dstIP)
+			pc.cfg.ClientLogger.Errorf("failed to read piece cont from dst:%s ,error:%s", err, dstIP)
 			// TODO handle dst_ip == self.node
```
lowzj: remove this line: `// TODO handle dst_ip == self.node`

dfget/core/downloader/p2p_downloader.go
Starnop: Maybe it's not a good idea. It is not always necessary to put the returned piece both into `quene` and `clientQueue`.

dfget/core/downloader/power_client.go
```diff
@@ -54,59 +56,64 @@ func (pc *PowerClient) Run() (err error) {
...
-		total, err := pieceCont.ReadFrom(reader)
-		pc.cfg.ClientLogger.Infof("get pieceCont total: %d", total)
-		if err != nil {
-			return err
-		}
-		// TODO handle read timeout
+	// send download request
+	url := fmt.Sprintf("http://%s:%d%s", dstIP, peerPort, pc.pieceTask.Path)
+	headers := make(map[string]string)
+	headers["Range"] = pc.pieceTask.Range
```
lowzj: `headers["Range"] = pc.pieceTask.Range' -> `headers["Range"] = "bytes=" + pc.pieceTask.Range`

PR #358, opened 2019-01-29
fix: optimize the judgment of local cache expiration

lowzj:

* increase read timeout to 2 seconds
* not expire if get response from remote server timeout

### DIFF
--- a/src/supernode/src/main/java/com/dragonflyoss/dragonfly/supernode/common/util/HttpClientUtil.java
+++ b/src/supernode/src/main/java/com/dragonflyoss/dragonfly/supernode/common/util/HttpClientUtil.java
@@ -27,7 +27,6 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.HttpsURLConnection;
 import javax.net.ssl.SSLContext;
@@ -85,7 +84,7 @@ public class HttpClientUtil {
                 fillHeaders(conn, headers);
                 conn.setUseCaches(false);
                 conn.setConnectTimeout(2000);
-                conn.setReadTimeout(1000);
+                conn.setReadTimeout(2000);
                 if (lastModified > 0) {
                     conn.setIfModifiedSince(lastModified);
                 }
@@ -106,7 +105,10 @@ public class HttpClientUtil {
                 }
                 break;
             } catch (Exception e) {
-                logger.warn("url:{} isExpired error", fileUrl, e);
+                logger.warn("url:{} isExpired error:%s", fileUrl, e.getMessage(), e);
+                if (times == 0) {
+                    return false;
+                }
             } finally {
                 closeConn(conn);
                 conn = null;


src/supernode/src/main/java/com/dragonflyoss/dragonfly/supernode/common/util/HttpClientUtil.java
```diff
@@ -84,8 +82,9 @@ public static boolean isExpired(String fileUrl, long lastModified, String eTag,
                 conn = openConnection(url);
                 fillHeaders(conn, headers);
                 conn.setUseCaches(false);
+                //conn.setRequestMethod("HEAD");
                 conn.setConnectTimeout(2000);
-                conn.setReadTimeout(1000);
+                conn.setReadTimeout(2000);
```
Starnop: How did you get this value? I think it is unreasonable if it is based on experience. 
Could we make it configurable? 
+ If that, we have to do that without hesitation . 
+ If not, let's take a step back and set a constant.

WDYT?

src/supernode/src/main/java/com/dragonflyoss/dragonfly/supernode/common/util/HttpClientUtil.java
```diff
@@ -84,8 +82,9 @@ public static boolean isExpired(String fileUrl, long lastModified, String eTag,
                 conn = openConnection(url);
                 fillHeaders(conn, headers);
                 conn.setUseCaches(false);
+                //conn.setRequestMethod("HEAD");
```
Starnop: Delete it or record a TODO item?

src/supernode/src/main/java/com/dragonflyoss/dragonfly/supernode/common/util/HttpClientUtil.java
```diff
@@ -126,6 +128,7 @@ public static long getContentLength(String fileUrl, String[] headers, boolean df
                 conn = openConnection(url);
                 conn.setUseCaches(false);
                 fillHeaders(conn, headers);
+                //conn.setRequestMethod("HEAD");
```
Starnop: So do with above.

src/supernode/src/main/java/com/dragonflyoss/dragonfly/supernode/common/util/HttpClientUtil.java
lowzj: Keep the same with the value in function `getContentLength`. I have no plan to make it configurable in java-version. Maybe we could do that when we refactor supernode with golang.

Starnop: Then set a constant is better. WDYT?

src/supernode/src/main/java/com/dragonflyoss/dragonfly/supernode/common/util/HttpClientUtil.java
```diff
@@ -28,16 +34,8 @@
...
-import javax.net.ssl.HostnameVerifier;
-import javax.net.ssl.HttpsURLConnection;
-import javax.net.ssl.SSLContext;
-import javax.net.ssl.SSLSession;
-import javax.net.ssl.TrustManager;
-import javax.net.ssl.X509TrustManager;
-
 import com.dragonflyoss.dragonfly.supernode.common.exception.AuthenticationRequiredException;
 import com.dragonflyoss.dragonfly.supernode.common.exception.UrlNotReachableException;
-
```
Starnop: Why delete it?

src/supernode/src/main/java/com/dragonflyoss/dragonfly/supernode/common/util/HttpClientUtil.java
lowzj: I think that this work can be done in supernode go-version, and java-version we just do some bugfix work.

chenchaobing: > I think that this work can be done in supernode go-version, and java-version we just do some bugfix work.

agree

chenchaobing:
LGTM

PR #359, opened 2019-01-29
feature: report downloaded piece's status to supernode

lowzj:

* report the downloaded piece's status to supernode, then other peers can download from this peer
* fix the incorrect `dataDir` of downloading task in dfget server

### DIFF
--- a/dfget/config/constants.go
+++ b/dfget/config/constants.go
@@ -92,6 +92,7 @@ const (
 	StrRange         = "Range"
 	StrContentLength = "Content-Length"
 	StrContentType   = "Content-Type"
+	StrUserAgent     = "User-Agent"
 
 	StrTaskFileName = "taskFileName"
 	StrClientID     = "cid"
@@ -106,6 +107,26 @@ const (
 	StrBytes = "bytes"
 )
 
+/* piece meta */
+const (
+	// PieceHeadSize every piece starts with a piece head which has 4 bytes,
+	// its value is:
+	//    real data size | (piece size << 4)
+	// And it's written with big-endian into the first four bytes of piece data.
+	PieceHeadSize = 4
+
+	// PieceTailSize every piece ends with a piece tail which has 1 byte,
+	// its value is: 0x7f
+	PieceTailSize = 1
+
+	// PieceMetaSize piece meta is constructed with piece head and tail,
+	// its size is 5 bytes.
+	PieceMetaSize = PieceHeadSize + PieceTailSize
+
+	// PieceTailChar the value of piece tail
+	PieceTailChar = byte(0x7f)
+)
+
 /* others */
 const (
 	DefaultTimestampFormat = "2006-01-02 15:04:05"
@@ -114,8 +135,8 @@ const (
 	ServerPortLowerLimit = 15000
 	ServerPortUpperLimit = 65000
 
-	RangeNotExistDesc = "range not satisfiable"
-	AddrUsedDesc      = "address already in use"
+	RangeNotSatisfiableDesc = "range not satisfiable"
+	AddrUsedDesc            = "address already in use"
 
 	PeerHTTPPathPrefix = "/peer/file/"
 	CDNPathPrefix      = "/qtdown/"
--- a/dfget/core/api/download_api.go
+++ b/dfget/core/api/download_api.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/dragonflyoss/Dragonfly/dfget/config"
 	"github.com/dragonflyoss/Dragonfly/dfget/util"
+	"github.com/dragonflyoss/Dragonfly/version"
 )
 
 // DownloadRequest wraps the request which is sent to peer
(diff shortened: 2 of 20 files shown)


dfget/core/downloader/power_client.go
```diff
@@ -254,9 +258,24 @@ func (cw *ClientWriter) write(piece *Piece, startTime time.Time) error {
...
 }
 
+func (cw *ClientWriter) sendSuccessPiece(piece *Piece, cost time.Duration) {
+	cw.api.ReportPiece(piece.SuperNode, &types.ReportPieceRequest{
+		TaskID:     piece.TaskID,
+		Cid:        cw.Cfg.RV.Cid,
+		DstCid:     piece.DstCid,
+		PieceRange: piece.Range,
+	})
+	if cost.Seconds() > 2.0 {
```
Starnop: Just curious.  What is the necessity of this judgment?

dfget/core/downloader/power_client.go
```diff
@@ -158,6 +161,7 @@ type ClientWriter struct {
 	targetQueue  util.Queue
 	targetWriter *TargetWriter
 
+	api api.SupernodeAPI
```
Starnop: Is it necessary? In fact, I think it's more reasonable to set `api.SupernodeAPI` as a global variable, what do you think?

dfget/core/downloader/power_client.go
lowzj: output which piece download takes more time

dfget/core/downloader/power_client.go
lowzj: It will be easier for testing to set `api.SupernodeAPI` as a field of `ClientWriter` than as a global variable.

dfget/util/log.go
```diff
@@ -35,7 +35,6 @@ func CreateLogger(logPath string, logName string, logLevel string, sign string)
 	// parse log level
 	level, err := log.ParseLevel(logLevel)
 	if err != nil {
-		// TODO: print a warning.
```
Starnop: I think we should print a warning here? WDYT?

dfget/config/constants.go
```diff
@@ -106,6 +107,14 @@ const (
 	StrBytes = "bytes"
 )
 
+/* piece meta */
+const (
+	PieceHeadSize = 4
```
Starnop: Could you please add comments here to make it clear?

dfget/core/api/download_api.go
```diff
@@ -54,6 +55,7 @@ func (d *downloadAPI) Download(ip string, port int, req *DownloadRequest) (*http
 	headers[config.StrRange] = config.StrBytes + "=" + req.PieceRange
 	headers[config.StrPieceNum] = strconv.Itoa(req.PieceNum)
 	headers[config.StrPieceSize] = fmt.Sprint(req.PieceSize)
+	headers[config.StrUserAgent] = "dfget/" + version.DFGetVersion
```
Starnop: Regardless of `Supernode`, uploader does not parse this header  currently.

dfget/core/downloader/p2p_downloader/client_writer.go
```diff
@@ -49,18 +51,21 @@ type ClientWriter struct {
...
 
-	Cfg *config.Config
+	api api.SupernodeAPI
+	cfg *config.Config
 }
 
 // NewClientWriter creates and initialize a ClientWriter instance.
-func NewClientWriter(taskFileName, cid, clientFilePath, serviceFilePath string, clientQueue util.Queue, Cfg *config.Config) (*ClientWriter, error) {
+func NewClientWriter(taskFileName, cid, clientFilePath, serviceFilePath string,
+	clientQueue util.Queue, api api.SupernodeAPI, Cfg *config.Config) (*ClientWriter, error) {
```
Starnop: s/Cfg/cfg?

dfget/core/api/download_api.go
lowzj: `User-Agent` is a standard http header, it's convenient for us to know where the request comes from.

dfget/util/log.go
lowzj: `dfget` only use `debug` or `info`, the code here just in case.

dfget/config/constants.go
lowzj: done

dfget/core/downloader/back_downloader/back_downloader.go
```diff
@@ -41,6 +40,8 @@ type BackDownloader struct {
 	Total   int64
 	Success bool
 
+	cfg *config.Config
```
allencloud: Does `BackDownloader` will use all fields in `config.Config`?

I do not think that it is a good way to use `config.Config` in `BackDownloader`. Since other fields in `BackDownloader`, like `URL`, `MD5` and so on are also in `config.Config`. Then it would be mixture when find a parameter which is both in `BackDownloader` and `BackDownloader.cfg`. I do not think it is a good habit to programme.

While for here, I think it is good to do like this:

* make clear that which parameters BackDownloader needs exactly;
* try to encapsulate all the parameter above into some field like `BackDownloaderConfig`;
* do a thoroughly initialization of  `BackDownloaderConfig` at the very beginning of `BackDownloader`.

I think data structure's purity is quite important for readiness of code. @lowzj

lowzj: I agree that data structure's purity is important for reading. Actually, the fields in `BackDownloaderConfig` are the public fields in `BackDownloader`, and they are initialized from `config.Config` in constructor `NewBackDownloader`.

Maybe we could open an issue and another pull request to do this.

allencloud: Yes, I will put forward an issue soon.

allencloud:
LGTM

PR #360, opened 2019-01-29
refactor: untify the api of uploader

Starnop:

The uploader provides a few http APIs. we splice the url directly where it needed to be called previously. This is a very rough approach, we should define it uniformly.

None.

Help wanted.

### DIFF
--- a/dfget/core/api/uploader_api.go
+++ b/dfget/core/api/uploader_api.go
@@ -1,0 +1,94 @@
+/*
+ * Copyright The Dragonfly Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package api
+
+import (
+	"fmt"
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/dragonflyoss/Dragonfly/dfget/config"
+	"github.com/dragonflyoss/Dragonfly/dfget/util"
+)
+
+// UploaderAPI defines the communication methods between dfget and uploader.
+type UploaderAPI interface {
+	// ParseRate sends a request to uploader to calculate the ratelimit dynamically
+	// for the speed limit of the whole host machine.
+	ParseRate(ip string, port int, req *ParseRateRequest) (string, error)
+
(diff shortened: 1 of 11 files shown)


dfget/core/api/uploader_api_types.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"time"
+)
+
+// CommonURL wraps the basic fields needed for a request.
+type CommonURL struct {
+	IP      string
+	Port    int
+	Timeout time.Duration
+}
```
lowzj: I think that it may be not a good idea to wrap these fields for a request. Actually, these fields are not parameter of uploader server's APIs. It's better to set them as fields of uploaderAPI struct.

dfget/core/api/uploader_api_types.go
```diff
@@ -0,0 +1,69 @@
...
+
+// ParseRateRequest wraps the request which is sent to uploader
+// in order to calculate the rate limit dynamically.
+type ParseRateRequest struct {
+	CommonURL
+
+	TaskFileName string
+	RateLimit    string
```
lowzj: It's better to set `RateLimit`'s type as `int`.

dfget/core/api/uploader_api.go
```diff
@@ -0,0 +1,121 @@
...
+// uploaderAPI is an implementation of interface UploaderAPI.
+type uploaderAPI struct {
+}
+
+// NewUploaderAPI returns a new UploaderAPI.
+func NewUploaderAPI() UploaderAPI {
+	return &uploaderAPI{}
+}
+
+func (u *uploaderAPI) Download(req *DownloadRequest) (*http.Response, error) {
```
lowzj: I think that it's not a api of uploader server. It seems that dfget will download pieces only from uploader.

Starnop: Done.

dfget/core/api/uploader_api_types.go
Starnop: Done.

dfget/core/downloader/power_client.go
```diff
@@ -71,13 +71,15 @@ func (pc *PowerClient) Run() (err error) {
...
 
 	// send download request
-	url := fmt.Sprintf("http://%s:%d%s", dstIP, peerPort, pc.pieceTask.Path)
-	headers := make(map[string]string)
-	headers["Range"] = "bytes=" + pc.pieceTask.Range
-	headers["pieceNum"] = strconv.Itoa(pc.pieceTask.PieceNum)
-	headers["pieceSize"] = fmt.Sprint(pc.pieceTask.PieceSize)
+	req := &api.DownloadRequest{
+		Path:       pc.pieceTask.Path,
+		PieceRange: config.StrBytes + "=" + pc.pieceTask.Range,
```
lowzj: construct `PieceRange` in download implementation

Starnop: Done.

lowzj:
LGTM

Issue #347, opened 2019-01-23
Plan of dfget's refactoring plan


allencloud:
## Background

This is one part of [Golang Refactoring Plan of the Whole Dragonfly Project](https://github.com/dragonflyoss/Dragonfly/issues/346)

dfget is the client of Dragonfly's Client/Server Architecture. It is called by dfdaemon and used to interact with supernode of Dragonfly.

Although dfget is built by golang currently, the interact way with user and its calling API is not so ready in terms of scalability. So we split the dfget refactoring with the following details.

## Refactoring Details

* API client refactoring
    * Golang SDK of Dragonfly's API
* CLI client refactoring
    * cobra command adoption of `dfget`
    * detailed description of every essential cobra command's fields
* base and advanced features
    * support pull images from multiple remote registries;
    * support downloading an empty file;
    * private image support with no invasiveness;
    * HTTPS connection support for registry.

fengzixu:
Let me do it. If someone wants to participate in this issue together with me, please contact me.

fengzixu:
I found that you submit the PR: https://github.com/dragonflyoss/Dragonfly/pull/353. Is it conflict with the refactor of dfget ? @Starnop @lowzj . For the refactor of dfget, do you have any requirement? @allencloud

Starnop:
@fengzixu The task list has been updated. ASAP. This is also an important feature of  the refactor of dfget.

Starnop:
> Let me do it. If someone wants to participate in this issue together with me, please contact me.

So glad for your active participation. However, there is a small advice :
This issue is a task list. I think you should focus on one task at a time. 
In addition, you should assign yourself before you focus on completing an issue.
WDYT? 😄

fengzixu:
Sure. I think that before we want to begin to resolve one of those tasks, we must comment in this issue. @Starnop

fengzixu:
I would like to handle the task "unify the same field type definition in different places" @Starnop @lowzj

lowzj:
I will open a pull request to implements this feature: `HTTPS connection support for registry`, it's actually a feature of `dfdaemon`.

PR #362, opened 2019-01-30
refactor: reorganize the dfget downloader files

Starnop:

I submitted this PR for the following reasons:

+ I found that the file of `p2p downloader` was too big to write unit tests in a PR when I tried to do that. The functions inside are obviously divided into multiple parts, so why don't we organize the code with dividing multiple files?
+ In addition, we put the `back downloader` and `p2p downloader` in the same folder, which is unclear.

So I decide to reorganize the files with downloader.

None.

 I will do that in the next pull request  for the simplicity of each PR.

Pass the CI.


dfget/core/downloader/downloader.go
```diff
@@ -98,7 +58,8 @@ func DoDownloadTimeout(downloader Downloader, timeout time.Duration) error {
 	return err
 }
 
-func convertHeaders(headers []string) map[string]string {
+// ConvertHeaders conver headers from array type to map type for http request.
```
lowzj: `conver` -> `converts`

dfget/core/downloader/p2p_downloader/client_writer.go
```diff
@@ -0,0 +1,169 @@
...
+	"os"
+	"time"
+
+	"github.com/dragonflyoss/Dragonfly/dfget/config"
+	"github.com/dragonflyoss/Dragonfly/dfget/core/helper"
+	"github.com/dragonflyoss/Dragonfly/dfget/util"
+)
+
+// ClientWriter writes a file for uploading and a target file.
+type ClientWriter struct {
```
allencloud: We need to add more description for every fields in struct definition.

Starnop: Maybe a a separate pull request. This pull request is only to reorganize the files.

lowzj:
LGTM

dfget/core/downloader/p2p_downloader/client_writer.go
Starnop: Has been recorded in the issue #347.

Issue #364, opened 2019-01-31
[bugfix] Dragonfly bind an already in used port make dfdaemon.log file size huge


godliness:
### Ⅰ. Issue Description

Dragonfly bind an already in used port make dfdaemon.log file size huge

### Ⅱ. Describe what happened

Dragonfly bind an already in used port make dfdaemon.log file size huge,  and dragonfly write into dfdaemon.log file frequently, below is the log content:
```
time="2019-01-28 12:27:00" level=info msg="start dfdaemon param:&{DfPath:/software/servers/df-client/dfget DFRepo:/data/dragonfly/.small-dragonfly/dfdaemon/data/ RateLimit:1000M CallSystem:com_ops_dragonfly URLFilter:Signature&Expires&OSSAccessKeyId Notbs:true MaxProcs:4 Version:false Verbose:false Help:false HostIP:127.0.0.1 Port:65001 Registry:http://repo.image.local DownRule: CertFile: KeyFile:}"
time="2019-01-28 12:27:00" level=fatal msg="listen tcp :65001: bind: address already in use"
time="2019-01-28 12:27:00" level=info msg=init...
time="2019-01-28 12:27:00" level=info msg="dfget version:0.2.0\n"
time="2019-01-28 12:27:00" level=info msg="init finish"
time="2019-01-28 12:27:00" level=info msg="start dfdaemon param:&{DfPath:/software/servers/df-client/dfget DFRepo:/data/dragonfly/.small-dragonfly/dfdaemon/data/ RateLimit:1000M CallSystem:com_ops_dragonfly URLFilter:Signature&Expires&OSS
```
![image](https://user-images.githubusercontent.com/10972761/52026775-0269c200-2544-11e9-87fb-cc966507a8eb.png)

But there is another confused thing is that dragonfly actually already bind on port 65001, but it still output the message `time="2019-01-28 12:27:00" level=fatal msg="listen tcp :65001: bind: address already in use"` 

![image](https://user-images.githubusercontent.com/10972761/52027024-dc90ed00-2544-11e9-9ebf-7ff2fb7213d1.png)

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1. 
2.
3.

### Ⅵ. Environment:

- dragonfly version: 0.2.1
- OS (e.g. from /etc/os-release):
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
[...]

PR #369, opened 2019-02-01
bugfix: fix rotating log makes dfdaemon.log file huge size

godliness:

When the size of dfdaemon.log file large than 20M, the size of log file will continuous growth up.
Use Mmap to copy the file which need be truncated.

fix https://github.com/dragonflyoss/Dragonfly/issues/364

Run this pr in my local machine, when dfdaemon.log large than 20M, truncated it into 10M successfully.

/cc @lowzj 
/cc @allencloud

### DIFF
--- a/dfdaemon/initializer/initializer.go
+++ b/dfdaemon/initializer/initializer.go
@@ -18,7 +18,6 @@ package initializer
 
 import (
 	"fmt"
-	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


dfdaemon/initializer/initializer.go
```diff
@@ -102,29 +101,33 @@ func cleanLocalRepo(options *options.Options) {
...
-				// move the pointer to head
-				logFile.Seek(0, 0)
-				count, _ := io.Copy(logFile, transFile)
-				logFile.Truncate(count)
-				log.SetOutput(logFile)
-				transFile.Close()
+			truncateSize := logSizeLimit/2 - 1
+			mem, err := syscall.Mmap(int(logFile.Fd()), 0, int(stat.Size()), syscall.PROT_READ|syscall.PROT_WRITE, syscall.MAP_SHARED)
+			if err != nil {
+				log.Errorf("failed to map file and memory: %s", err)
```
lowzj: These logs here will be discarded.

godliness: @lowzj 
 thx! fixed.

dfdaemon/initializer/initializer.go
```diff
@@ -102,32 +101,37 @@ func cleanLocalRepo(options *options.Options) {
...
+func rotateLog(logFile *os.File) {
 	logSizeLimit := int64(20 * 1024 * 1024)
 	for {
 		time.Sleep(time.Second * 60)
-		stat, err := os.Stat(logFilePath)
+		stat, err := logFile.Stat()
 		if err != nil {
-			log.Errorf("failed to stat %s: %s", logFilePath, err)
+			// make sure this error could be printed when logFile is broken
+			fmt.Fprintf(os.Stderr, "failed to stat: %s\n", err)
```
allencloud: I am not sure this error could be printed in stderr, since the fd stdout and stderr of process has been redirect to the logfile. Am I right? @godliness @mdianjun @lowzj @Starnop

dfdaemon/initializer/initializer.go
```diff
@@ -102,32 +101,37 @@ func cleanLocalRepo(options *options.Options) {
...
 			continue
 		}
 		// if it exceeds the 20MB limitation
 		if stat.Size() > logSizeLimit {
 			log.SetOutput(ioutil.Discard)
+			// make sure set log output back to logFile when error raised
+			defer log.SetOutput(logFile)
```
allencloud: Since the function `rotateLog` has a loop without `break` or `return`, the for loop is infinite. If the function is infinite without return, I am wondering if the `defer log.SetOutput(logFile)` will be triggered.

dfdaemon/initializer/initializer.go
godliness: @allencloud 

Thanks, i think when we use `log.SetOutput(out io.Writer)` which means write log into the "out", the out could be a logfile and also could be std.out or std.err, here we write the log into std.err that i think it could be printed.

dfdaemon/initializer/initializer.go
godliness: @allencloud 

Yes, you're right, it won't be triggered cause this function is  infinite, so i removed this loop from this function and make the function return err so that it could be easy to collect err and printed them.

dfdaemon/initializer/initializer.go
```diff
@@ -148,7 +146,15 @@ func initLogger() {
 		if logFile, err := os.OpenFile(logFilePath, os.O_CREATE|os.O_RDWR|os.O_APPEND, 0644); err == nil {
 			logFile.Seek(0, 2)
 			log.SetOutput(logFile)
-			go rotateLog(logFile, logFilePath)
+			go func(logFile *os.File) {
+				ticker := time.NewTicker(60 * time.Second)
+				for range ticker.C {
+					if err := rotateLog(logFile); err != nil {
+						// write error log into os.Stderr directly to avoid writing to a broken log file.
+						fmt.Fprintf(os.Stderr, "failed to rotate log %s: %v\n", logFile.Name(), err)
```
lowzj: I see that function `rotateLog` ensures that it will reset `logFile` as the output of log, so maybe we can just use `log` here?

godliness: @lowzj 

OK, i just afraid when we failed to rotate the log, so weird thing happened on the logFile that will lead to write the log into log file failed, but i'm not sure what the weird things are.

godliness: @lowzj 

Alternatively, what about we use MultiWriter?

```
// log both file and console
wrt := io.MultiWriter(os.Stdout, f)
// set output
log.SetOutput(wrt)
```

lowzj: @godliness I think that it's a good option for user to set when we want to debug or test locally. It's better to do this in a new pull request.

lowzj:
LGTM

PR #372, opened 2019-02-02
refactor: directly config logrus' default Logger rather than create a new one

allencloud:

In the component of dfget, the Logger is regarded one field of the configuration, and it will be transferred to other places where logger recording is needed.

Since when dfget process is started to the status of its exit, the log configuration will not change, then I think we can directly use the default `logrus.Logger` in pkg `logrus`. And we only need to configure the default logger's LogLevel, Formatter and Output via:

* logrus.SetOutput()
* logrus.SetLevel()
* logrus.SetFormatter()

no, just refactoring

no 

no

none

### DIFF
--- a/cmd/dfget/app/root.go
+++ b/cmd/dfget/app/root.go
@@ -73,13 +73,13 @@ func runDfget() error {
 	}
 
 	checkParameters()
-	cfg.ClientLogger.Infof("get cmd params:%q", os.Args)
+	logrus.Infof("get cmd params:%q", os.Args)
 
 	if err := config.AssertConfig(cfg); err != nil {
 		util.Printer.Println(fmt.Sprintf("assert context error: %v", err))
 		return err
 	}
-	cfg.ClientLogger.Infof("get init config:%v", cfg)
+	logrus.Infof("get init config:%v", cfg)
 
 	// enter the core process
 	err := core.Start(cfg)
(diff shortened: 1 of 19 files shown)


dfget/util/log.go
```diff
@@ -30,69 +27,29 @@ import (
 // DefaultLogTimeFormat defines the timestamp format.
 const DefaultLogTimeFormat = "2006-01-02 15:04:05.000"
 
-// CreateLogger creates a logger.
-func CreateLogger(logPath string, logName string, logLevel string, sign string) (*log.Logger, error) {
```
lowzj: I think that we should retain this function and create a new function to initialize the `logrus` default log instant. And in future, we should share this log package with `dfget`, `dfdaemon`, `supernode`.
`supernode` need to write different log files in different modules, we may use it when refactoring `supernode`.

cmd/dfget/app/root.go
```diff
@@ -145,22 +146,48 @@ func transParams() error {
 
 // initClientLog init client logger.
 func initClientLog() error {
-	var (
-		logPath  = path.Join(cfg.WorkHome, "logs")
-		logLevel = "info"
-	)
+	// set the log level
+	logLevel := logrus.InfoLevel
```
lowzj: create a function to initialize logrus, then the `dfget` and `dfget server` could reuse this function to init log.

allencloud:
> It's better to retain the test file: dfget/util/log_test.go.

I have recalled the deleted file. Thanks for your review. @lowzj

lowzj:
LGTM

Issue #385, opened 2019-02-12
[feature request] dfdaemon support forwarding requests of some trusted hosts


godliness:
---
name: Feature Request
about: Suggest your excellent idea for this project

---

## Why you need it?

If we only use `registry-mirrors ` to support request forwarding, then all the yaml of running pods need change their image field. eg: from repo.image.local/test:0.0.1 to 127.0.0.1:65001/test:0.0.1 that is not be accepted for platform users; so we try to use `http_proxy` in docker.service file directly, but there is still some problems that we only support "127.0.0.1" and "localhost" and "hostIP" that dfdaemon running on.

## How it could be?

What about we support multi trusted hosts which could be passed by command line params which could be named "trust-hosts" or something else?

## Other related information

Related to issue: 

https://github.com/dragonflyoss/Dragonfly/issues/264

https://github.com/dragonflyoss/Dragonfly/issues/289

Referenced:

https://www.bladewan.com/2018/08/18/docker_image_p2p/
![image](https://user-images.githubusercontent.com/10972761/52609090-0e943e80-2eb7-11e9-8ef5-4dfe6096a869.png)

Any suggestions？

/cc @allencloud 
/cc @lowzj

lowzj:
closed because that pr: #389 is merged

PR #389, opened 2019-02-12
feature: add trust hosts params for dfdaemon

godliness:

As dfdaemon be used to http_proxy mode, all the requests with host
are not local ip of dfdaemon then client will got a 403 error code.

Add a trust-hosts command line params and determine the request wether
be trusted or not, if it's trusted we directly forward them.

Related to issue:
https://github.com/dragonflyoss/Dragonfly/issues/385

None

1. `./dfdaemon --trust-hosts repo.image.local,ui.repo.image.local,t1.repo.image.local`

2. inscure-registry be configured` ["repo.image.local","ui.repo.image.local","t1.repo.image.local", "t2.repo.image.local"]`

3. docker.service file be configured `Environment="HTTP_PROXY=http://127.0.0.1:65001"`

4. docker pull repo.image.local/nginx which download image successfully

5. docker pull t2.repo.image.local/nginx which not in the `--trust-hosts` 403 error be raised.

--

/cc @allencloud  @lowzj

### DIFF
--- a/cmd/dfdaemon/app/options/options.go
+++ b/cmd/dfdaemon/app/options/options.go
@@ -74,6 +74,10 @@ type Options struct {
 
 	// Key file path.
 	KeyFile string
+
+	// TrustHosts includes the trusted hosts which dfdaemon forward their
+	// requests directly when dfdaemon is used to http_proxy mode.
+	TrustHosts []string
 }
 
 // NewOption returns the default options.
@@ -115,5 +119,5 @@ func (o *Options) AddFlags(fs *flag.FlagSet) {
 	fs.StringVar(&o.Registry, "registry", "https://index.docker.io", "registry addr(https://abc.xx.x or http://abc.xx.x) and must exist if dfdaemon is used to mirror mode")
 	fs.StringVar(&o.DownRule, "rule", "", "download the url by P2P if url matches the specified pattern,format:reg1,reg2,reg3")
 	fs.BoolVar(&o.Notbs, "notbs", true, "not try back source to download if throw exception")
-
+	fs.StringSliceVar(&o.TrustHosts, "trust-hosts", o.TrustHosts, "list of trusted hosts which dfdaemon forward their requests directly, comma separated.")
 }
--- a/dfdaemon/global/global.go
+++ b/dfdaemon/global/global.go
@@ -51,6 +51,10 @@ type CommandParam struct {
 	// Registry addr and must exist if dfdaemon is used to mirror mode,
 	// format: https://xxx.xx.x:port or http://xxx.xx.x:port.
 	Registry string
+
+	// TrustHosts includes the trusted hosts as keys of the map which dfdaemon forward their
+	// requests directly when dfdaemon is used to http_proxy mode.
+	TrustHosts map[string]string
 }
 
 var (
--- a/dfdaemon/handler/root_handler.go
+++ b/dfdaemon/handler/root_handler.go
@@ -36,7 +36,10 @@ func Process(w http.ResponseWriter, r *http.Request) {
 			r.URL.Host = r.Header.Get("Host")
 		}
 		if r.URL.Host == "" {
+			// if host is still empty, we need skip to forward it.
+			w.WriteHeader(http.StatusForbidden)
 			log.Errorf("url host is empty")
+			return
 		}
 	}
 	r.Host = r.URL.Host
@@ -57,6 +60,7 @@ func Process(w http.ResponseWriter, r *http.Request) {
 	targetURL.RawQuery = ""
 
 	hostIP := util.ExtractHost(r.URL.Host)
+	trustHost := global.CommandLine.TrustHosts[hostIP]
 	switch hostIP {
 	case "127.0.0.1", "localhost", global.CommandLine.HostIP:
 		if len(global.CommandLine.Registry) > 0 {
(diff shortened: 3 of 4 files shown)


cmd/dfdaemon/app/options/options.go
```diff
@@ -115,5 +119,5 @@ func (o *Options) AddFlags(fs *flag.FlagSet) {
 	fs.StringVar(&o.Registry, "registry", "https://index.docker.io", "registry addr(https://abc.xx.x or http://abc.xx.x) and must exist if dfdaemon is used to mirror mode")
 	fs.StringVar(&o.DownRule, "rule", "", "download the url by P2P if url matches the specified pattern,format:reg1,reg2,reg3")
 	fs.BoolVar(&o.Notbs, "notbs", true, "not try back source to download if throw exception")
-
+	fs.StringSliceVar(&o.TrustHosts, "trust-hosts", o.TrustHosts, "List of trusted hosts which dfdaemon forward their requests directly, comma separated.")
```
fengzixu: lowercase "List"

dfdaemon/handler/root_handler.go
```diff
@@ -65,13 +65,17 @@ func Process(w http.ResponseWriter, r *http.Request) {
 		} else {
 			log.Warnf("registry not config but url host is %s", hostIP)
 		}
+	case global.CommandLine.TrustHosts[hostIP]:
```
fengzixu: In my opinion, we should pick up the value from the map before the controlling-stream enters the `swich-case`. Use approaches as below:
```
value, ok := m[xxx]
```
But, if you can find some evidence to prove that there is no risk in here, you can tell me.

dfdaemon/initializer/initializer.go
```diff
@@ -229,6 +229,13 @@ func initParam(options *options.Options) {
 		}
 	}
 
+	parsedTrustHosts := make(map[string]string)
```
fengzixu: Did you test the more than two trust hosts through command parameters? If you did, please complete the description of this PR.

fengzixu:
You can modify the title of this PR according to the doc of contributing in the root-dir.

dfdaemon/handler/root_handler.go
godliness: Thanks for the comments @fengzixu 

I thinks this should be fine, if the key of this map is not exist, the default value is `""`, so the value of `global.CommandLine.TrustHosts[hostIP]` is `""` here.

But there is an extreme case which is what if hostIP is `""`?  I think this should be ok, cause if the `hostIP` is `""` so that the request could not be forwarded at all.

dfdaemon/initializer/initializer.go
godliness: @fengzixu  

Added more than two trust hosts and test it, please see the description of this PR.

dfdaemon/handler/root_handler.go
fengzixu: 1. The standard way of getting value from a map in Golang is using `value, ok := m[key]`. I think we should obey this rule. 

2. As you said,  the request wasn't be forwarded when `global.CommandLine.TrustHosts[hostIP]` is `""`. But it makes the code less readable. As a reviewer, I don't know why you assign the null values to `targetURL.Host` and `targetURL.Scheme` respectively and what happens if `targetURL.Host` and `targetURL.Scheme` are `""` . However, previous logic looks like so clear:  if we don't find a valid hostIP, we will output warning info and throw the error. Although there is perhaps no impact after you modified,  we shouldn't do anything except logging if `global.CommandLine.TrustHosts[hostIP]` is `""`.

godliness: @fengzixu 

 Thanks,  I've move "pick up map" into "default case" , and added "return" if `r.URL.Host` is "" in 
 previous code.

godliness: @fengzixu 

Thanks for you review, i've fixed it.

1. Added "return" and `w.WriteHeader(http.StatusForbidden)` when `r.URL.Host` is still "" in previous code which already logging the host is empty condition.
2. Move "pick up map" before `switch-case` and added `case trustHost` in `switch-case`.

Please have a look, thx.

zhouhaibing089:
> As dfdaemon be used to http_proxy mode, all the requests with host
are not local ip of dfdaemon then client will got a 403 error code.

I am not sure whether it is really compatible with an http_proxy, most of the http_proxies I see is simply a tcp server, and accepts `CONNECT` requests.

If it is, then perhaps we should document this clearly somewhere.

dfdaemon/global/global.go
```diff
@@ -51,6 +51,10 @@ type CommandParam struct {
 	// Registry addr and must exist if dfdaemon is used to mirror mode,
 	// format: https://xxx.xx.x:port or http://xxx.xx.x:port.
 	Registry string
+
+	// TrustHosts includes the trusted hosts as keys of the map which dfdaemon forward their
+	// requests directly when dfdaemon is used to http_proxy mode.
+	TrustHosts map[string]string
```
zhouhaibing089: Not related to the change here:

It seems incorrect to repeat the same parameter in options and here, if possible, can we stop adding new fields here as I see, we may have to remove this variable sooner or later.

godliness: @zhouhaibing089 

Thanks, 
I have no idea about where should i add the new field before the refactor of this.
Refactor of removing global file should be another PR.

godliness:
@zhouhaibing089 

Tcp is transparent for us, cause dfdaemon is a http server i think, why we need care about it?

godliness:
@lowzj  Any suggestions?

godliness:
Does this PR could be merged？@lowzj

lowzj:
merged via command line

PR #391, opened 2019-02-13
build: populate version with ldflags

zhouhaibing089:
1. the version can be dynamically retrieved with git.
2. the previous BUILD ldflags seems not used.

### DIFF
--- a/hack/build-client.sh
+++ b/hack/build-client.sh
@@ -3,6 +3,8 @@ DFDAEMON_BINARY_NAME=dfdaemon
 DFGET_BINARY_NAME=dfget
 PKG=github.com/dragonflyoss/Dragonfly
 BUILD_IMAGE=golang:1.10.4
+VERSION=$(git describe --tags | cut -c 2-)
+LDFLAGS="-X ${PKG}/version.Version=${VERSION}"
 
 curDir=$(cd "$(dirname "$0")" && pwd)
 cd "${curDir}" || return
@@ -10,7 +12,6 @@ BUILD_SOURCE_HOME=$(cd ".." && pwd)
 
 . ./env.sh
 
-BUILD=$(git rev-parse HEAD)
 BUILD_PATH=bin/${GOOS}_${GOARCH}
 USE_DOCKER=${USE_DOCKER:-"0"}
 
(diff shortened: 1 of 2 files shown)


hack/build-client.sh
```diff
@@ -3,14 +3,15 @@ DFDAEMON_BINARY_NAME=dfdaemon
 DFGET_BINARY_NAME=dfget
 PKG=github.com/dragonflyoss/Dragonfly
 BUILD_IMAGE=golang:1.10.4
+VERSION=$(git describe --tags)
```
lowzj: remove the prefix letter `v` of tag name

luisdavim: maybe we could also populate `main.Build` with `git rev-parse --short HEAD`

zhouhaibing089: @lowzj: Updated.

@luisdavim: `git describe --tags` won't show the short hash of commit if this commit itself is already pointing to a tag. In another hand, if the current commit is not a tag, it actually helps if the short hash appears in version information.

lowzj: There's a problem here, now branch `master` is always an unreleased developing branch, and all tags are created based on other released version branches like `0.1.x`, `0.2.x`. So when we use `git describe --tags` on `master`, it's always `v0.0.1-xxx`(`v0.0.1` is the previous tags created on `master`) but `v0.3-xxx`(`0.3` is currently unreleased developing version) is expected.

luisdavim: You should create tags from master before cutting the release branch

lowzj: For example, I create the tag `v0.3.0` from master and cut the release branch `0.3.x`. Then I start to develop features of the next version `0.4` on master. During this time, there are some bugs of released version `0.3.0` reported. We have to fix them, merge them into `master`, backport them to branch `0.3.x`, and create a new tag `v0.3.1` from branch `0.3.x`.

Then I execute `make build` and `dfget version` on master, the output is `0.3.0-xxx`, but actually `0.4-xxx` is what I want.

zhouhaibing089: @lowzj: To be fair, `git describe --tags` is the only way which works across branches. My suggestion would be:

1. Suppose we are cutting a new release `0.4` from `master` branch.
1. we can create a new tag on master branch: `v0.5.0-alpha.0`.

Now even we haven't create any new tags on `0.4` branch, the version information would still be something like `v0.4.0.alpha.0-xxx`. Later on, when we are confident to release a stable branch, we can create a new tag `v0.4.0` on `0.4` branch, then.

lowzj: Agree, we should add a scripts to cut new branches automatically.

How about this:
1. `version.Version` only has the major and minor version such as `0.4`, it represents the version of this branch
2. when we cut a new branch `0.4` from `master`, then we should upgrade the `version.Version` on `master` to the next developing version `0.5`, the new version must be bigger than the old one.
3. add an initialization function to determine which version should be use: the one from `ldflags` and the one from `version.Version`, such as:

    ```go
    func init() {
    	if Version > DFDaemonVersion {
    		DFDaemonVersion = Version
    	}
    	if Version > DFGetVersion {
    		DFGetVersion = Version
    	}
    }
    ```

PR #410, opened 2019-02-21
feature: dfdaemon supports proxing https registries

### LINKED ISSUE #205: [feature request] dfdaemon support private registry with HTTPS protocol
---
name: Feature Request
about: Suggest your excellent idea for this project

---

## Why you need it?

It doesn't work when I pull private container images from an image registry with HTTPS protocol by using dragonfly.

The error outputted by dockerd:

```
denied: requested access to the resource is denied
unauthorized: authentication required
```

## How it could be?

Actually, the `dfdaemon` is just a proxy between dockerd and registry. The `dfdaemon` captures every request send from dockerd to registry, filters out all the image layer downloading requests and use `dfget` to download them.

We should investigate HTTPS proxy implementation and make it possible for `dfdaemon` to get http data from tcp data, analysis them and download image layers correctly.

lowzj:

---

This pull request implements a way to support registries which use `https` schema in `dfdaemon`'s proxy pattern:
* configure docker daemon
    * set `dfdaemon` host as `http_proxy`
    * set your registry hosts as `insecure-registries`
    * bypass proxy for your registry auth hosts
* `dfdaemon` get the http requests, change them to `https` schema and send https request to the real registries

fixes #205 
related #347 

Here's a example to pull images from a https registry(`registry.cn-hangzhou.aliyuncs.com`) by using proxy pattern:

* create dfdaemon's config file: `/etc/dragonfly/dfdaemon.yml`, and start `dfdaemon`
  ```yaml
  registries:
      - regx: "^registry.cn-hangzhou.aliyuncs.com$"
        schema: https
        host: registry.cn-hangzhou.aliyuncs.com
  ```
* change your docker daemon config file: `/etc/docker/daemon.json`, and restart `docker daemon`
  ```json
  {
      "insecure-registries" : [
          "registry.cn-hangzhou.aliyuncs.com"
      ],
      "proxies": {
          "default": {
              "httpProxy": "http://127.0.0.1:65001",
          }
      }
  }
  ```
* pull a image by dragonfly
  ```shell
  docker pull registry.cn-hangzhou.aliyuncs.com/dragonflyoss/supernode:0.2.1
  ```

@Starnop @godliness

### DIFF
--- a/dfdaemon/config/config.go
+++ b/dfdaemon/config/config.go
@@ -1,0 +1,181 @@
+/*
+ * Copyright The Dragonfly Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package config
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"io/ioutil"
+	"regexp"
+	"strings"
+
+	"gopkg.in/yaml.v2"
+)
+
+// -----------------------------------------------------------------------------
+// Properties
+
+// NewProperties create a new properties with default values.
+func NewProperties() *Properties {
+	return &Properties{}
+}
+
+// Properties holds all configurable properties of dfdaemon.
+// The default path is '/etc/dragonfly/dfdaemon.yml'
+// For examples:
+//     registries:
+//         - regx: (^localhost$)|(^127.0.0.1$)
+//           schema: http
+//           host: a.com
+//         - regx: ^reg.com:1001$
+//           schema: http
+//           host: reg.com
+//         - regx: ^reg.com:1002$
+//           schema: https
+//           host: reg.com
+//           certs: ['/etc/ssl/reg.com/server.crt']
+type Properties struct {
+	// Registries the more front the position, the higher priority.
+	// You could add an empty Registry at the end to proxy all other requests
+	// with those origin schema and host.
+	Registries []*Registry `yaml:"registries"`
+}
+
+// Load loads properties from config file.
+func (p *Properties) Load(path string) error {
+	if err := p.loadFromYaml(path); err != nil {
+		return err
+	}
+	var tmp []*Registry
+	for _, v := range p.Registries {
+		if v != nil {
+			if err := v.init(); err != nil {
+				return err
+			}
+			tmp = append(tmp, v)
+		}
+	}
+	p.Registries = tmp
+	return nil
+}
+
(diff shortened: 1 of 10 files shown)


dfdaemon/handler/root_handler.go
```diff
@@ -17,79 +17,104 @@
...
+	if global.UseHTTPS {
+		r.URL.Scheme = "https"
+	} else {
+		r.URL.Scheme = "http"
+	}
+	return nil
+}
+
+func sendResponse(w http.ResponseWriter, code int, resp string) {
+	w.WriteHeader(http.StatusBadRequest)
```
godliness: The params `code` of function `sendResponse` will always be re-assigned to `http.StatusBadRequest`?

dfdaemon/config/config.go
```diff
@@ -0,0 +1,123 @@
...
+
+// -----------------------------------------------------------------------------
+// Registry
+
+// NewRegistry create and init registry proxy with the giving values.
+func NewRegistry(schema, host, cert, regx string) (*Registry, error) {
+	reg := &Registry{
+		Schema: schema,
+		Host:   host,
+		Cert:   cert,
```
godliness: This param looks like never be used in this PR?

dfdaemon/handler/root_handler.go
lowzj: done

dfdaemon/config/config.go
godliness: This param looks like never be used in this PR?

godliness: This param looks like never be used in this PR?

lowzj: Yes, this param should be provided when remote https host is self-certificated. But now it works fine because that we don't verify the server's certificate. see code: [transport.go#63](https://github.com/dragonflyoss/Dragonfly/blob/master/dfdaemon/handler/transport.go#L63).

I will fix it:
* verify certificate if the `cert` is not empty
* don't verify certificate if the `cert` is empty

lowzj: done

dfdaemon/config/config.go
```diff
@@ -0,0 +1,149 @@
...
+func (r *Registry) Match(s string) bool {
+	return r.compiler != nil && r.compiler.MatchString(s)
+}
+
+// TLSConfig returns a initialized tls.Config instance.
+func (r *Registry) TLSConfig() *tls.Config {
+	return r.tlsConfig
+}
+
+func (r *Registry) init() error {
```
Starnop: Shall we have a validation for the parameters，such as "schema". And the test should contain the corresponding test instance.

godliness:
LGTM

dfdaemon/config/config.go
lowzj: done

dfdaemon/config/config.go
```diff
@@ -0,0 +1,166 @@
...
+// Properties
+
+// NewProperties create a new properties with default values.
+func NewProperties() *Properties {
+	return &Properties{}
+}
+
+// Properties holds all configurable properties of dfdaemon.
+// The default path is '/etc/dragonfly/dfdaemon.yml'
+type Properties struct {
```
Starnop: Could you please write a demo here and update the docs ASAP.

lowzj: I have add an example here.

PR #415, opened 2019-02-22
fix:  the speed will be slower and slower when multiple dfgets are executed consecutively on the same machine

lowzj:
…cuted consecutively on the same machine

Bugfix: the speed will be slower and slower when multiple dfgets are executed consecutively on the same machine.

### DIFF
--- a/dfget/core/uploader/peer_server.go
+++ b/dfget/core/uploader/peer_server.go
@@ -249,6 +249,7 @@ func (ps *peerServer) oneFinishHandler(w http.ResponseWriter, r *http.Request) {
 	if v, ok := ps.syncTaskMap.Load(taskFileName); ok {
 		task := v.(*taskConfig)
 		task.taskID = taskID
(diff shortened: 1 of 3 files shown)


dfget/core/uploader/peer_server.go
```diff
@@ -249,12 +249,14 @@ func (ps *peerServer) oneFinishHandler(w http.ResponseWriter, r *http.Request) {
...
 		task := v.(*taskConfig)
 		task.taskID = taskID
+		task.rateLimit = 0
 		task.cid = cid
 		task.superNode = superNode
 		task.finished = true
 	} else {
 		ps.syncTaskMap.Store(taskFileName, &taskConfig{
 			taskID:    taskID,
+			rateLimit: 0,
```
Starnop: The default value is zero. So it's not necessary.

Starnop:
LGTM.

PR #419, opened 2019-02-22
feature: add scripts to build rpm and deb packages for dfdaemon and dfget

### LINKED ISSUE #183: [question] Do we have a plan to support apt-get/yum installation
### Ⅰ. Issue Description

Currently dragonfly could be installed via docker images. While I think dragonfly is a very essential system software for distribution. 

How about supporting dragonfly installed via apt-get/yum? It would be very convenient for us to install that without any other dependencies.

Dragonfly Package

* supernode: supernode + ngnix
* dfget:  a binary
* daemon : a binary

inoc603:

This PR add scripts to build rpm and deb packages for dfdaemon and dfget, with [fpm](https://github.com/jordansissel/fpm).

fixes #183

I think since people may want to use dfget outside a docker container, we should provide easy installation.

This PR only adds a packaging script. 

First `make build-client USE_DOCKER=1 GOOS=linux` to build the binaries for **linux**, then

```
make release USE_DOCKER=1 VERSION=0.3.0
```

This will build rpm and deb packages for dfdaemon and dfget to `release/`. If you don't want to use docker, you can [install fpm](https://fpm.readthedocs.io/en/latest/installing.html) on your local machine, and drop `USE_DOCKER=1`. 

You can also `make rpm VERSION=0.3.0` or `make deb VERSION=0.3.0` to build rpm or deb packages separately.

You can verify the installation in docker containers.

```
docker run --rm -it -v $(pwd):$(pwd) -w $(pwd) centos:7 \
    bash -c "ls -l release/ && rpm -i release/*.rpm && ls -l /opt/dragonfly/df-client"

docker run --rm -it -v $(pwd):$(pwd) -w $(pwd) ubuntu:14.04 \
    bash -c "ls -l release/ && dpkg -i release/*.deb && ls -l /opt/dragonfly/df-client"
```

It seems how to version the package is still under discussion in #391 , so currently I think it's better to set the version explicitly. If `VERSION` is not set, the default version will be `0.0.$(date +%s)`.

The new supernode written in go doesn't seem production ready, and I can't find a build script for it, so I didn't package it. I'm not sure whether we should include the java supernode in the distribution, so it's also not packaged. However it's easy to include it if we've decided to do so.

The packages still needs more metadata, like description, maintainers, etc.

Currently we only build packages for `amd64`, other architecture can be added if necessary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,3 +66,16 @@ docs:
 	@echo "Begin to generate docs of API/CLI"
 	./hack/generate-docs.sh
 .PHONY: docs
+
+rpm:
+	./hack/package.sh rpm
+.PHONY: rpm
+
+deb:
+	./hack/package.sh deb
+.PHONY: deb
+
+dist:
+	./hack/package.sh
+.PHONY: dist
+
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,7 @@ USE_DOCKER?=0 # Default: build components in the local environment.
 clean:
 	@echo "Begin to clean redundant files."
 	@rm -rf ./bin
+	@rm -rf ./release
 .PHONY: clean
 
 build-dirs:
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,7 @@ deb:
 	./hack/package.sh deb
 .PHONY: deb
 
-dist:
+release:
 	./hack/package.sh
-.PHONY: dist
+.PHONY: release
 
--- a/hack/build-supernode.sh
+++ b/hack/build-supernode.sh
@@ -10,8 +10,8 @@ USE_DOCKER=${USE_DOCKER:-"0"}
 
 build-supernode-docker() {
     docker run --rm -it \
-        -v ${BUILD_SOURCE_HOME}:${BUILD_SOURCE_HOME} \
-        -w ${BUILD_SOURCE_HOME} \
+        -v "${BUILD_SOURCE_HOME}:${BUILD_SOURCE_HOME}" \
+        -w "${BUILD_SOURCE_HOME}" \
         maven:3.6-jdk-8 ./hack/build-supernode.sh
 }
 
--- a/hack/build-supernode.sh
+++ b/hack/build-supernode.sh
@@ -9,10 +9,9 @@ BIN_DIR="${BUILD_SOURCE_HOME}/bin"
 USE_DOCKER=${USE_DOCKER:-"0"}
 
 build-supernode-docker() {
-    docker run --rm -it \
-        -v "${BUILD_SOURCE_HOME}:${BUILD_SOURCE_HOME}" \
-        -w "${BUILD_SOURCE_HOME}" \
-        maven:3.6-jdk-8 ./hack/build-supernode.sh
+    cd "${SUPERNODE_SOURCE_HOME}" || return
+    mvn clean package -DskipTests docker:build -e
+    cp "${SUPERNODE_BIN}/supernode.jar"  "${BIN_DIR}/${SUPERNODE_BINARY_NAME}"
 }
 
 build-supernode-local(){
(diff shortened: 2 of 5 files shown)


hack/package.sh
```diff
@@ -0,0 +1,82 @@
...
+}
+
+# TODO: Add description
+DFGET_DESCRIPTION="dfget"
+DFDAEMON_DESCRIPTION="dfdaemon"
+# TODO: Add maintainer
+MAINTAINER="dragonflyoss"
+
+build_rpm() {
+    ${FPM} -s dir -t rpm -f -p release --rpm-os=linux \
```
lowzj: Could we build `dfget` and `dfdaemon` as one package, and create soft links in `/usr/local/bin` after installing them?

lowzj:
By the way, if we want to build a rpm package on other systems, we should add the environment `GOOS=linux` into make command:
```bash
make GOOS=linux USE_DOCKER=1 build-client
make GOOS=linux USE_DOCKER=1 VERSION=0.3.0 rpm
```

hack/package.sh
inoc603: Yes. I think building them into one package is better, since there are only 2 binaries that are usually used together, and their size are not very big. I build them into 2 packages because it's not clear in #183 how many packages there should be.

How about a `df-client` package for `dfget` and `dfdaemon`，and a `df-supernode` for `supernode` only?

inoc603:
> By the way, if we want to build a rpm package on other systems, we should add the environment `GOOS=linux` into make command:
> 
> ```shell
> make GOOS=linux USE_DOCKER=1 build-client
> make GOOS=linux USE_DOCKER=1 VERSION=0.3.0 rpm
> ```

Oh, I didn't notice that `GOOS` and `GOARCH` are passed as environment variables to `docker run`. I'll fix this.

hack/build-supernode.sh
```diff
@@ -9,9 +9,10 @@ BIN_DIR="${BUILD_SOURCE_HOME}/bin"
 USE_DOCKER=${USE_DOCKER:-"0"}
 
 build-supernode-docker() {
-    cd "${SUPERNODE_SOURCE_HOME}" || return
-    mvn clean package -DskipTests docker:build -e
-    cp "${SUPERNODE_BIN}/supernode.jar"  "${BIN_DIR}/${SUPERNODE_BINARY_NAME}"
+    docker run --rm -it \
```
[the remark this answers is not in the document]
inoc603: This change is reverted in 7195ba8

Makefile
```diff
@@ -66,3 +66,16 @@ docs:
...
+	./hack/package.sh rpm
+.PHONY: rpm
+
+deb:
+	./hack/package.sh deb
+.PHONY: deb
+
+release:
+	./hack/package.sh
+.PHONY: release
```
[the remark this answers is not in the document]
inoc603: Fixed in 6ad89a6

hack/package.sh
lowzj: OK.

lowzj:
@inoc603  the document could be put into `docs/user_guide/`

hack/package.sh
inoc603: Implemented in fa5d61236fd7c5ca943acade97435c32f3410564

lowzj:
LGTM

inoc603:
@lowzj I haven't written the document for packaging the client yet, because I ran into some strange cross-compile problem on OSX. I'll address this in another issue, then open a new PR for the docs.

lowzj:
> @lowzj I haven't written the document for packaging the client yet, because I ran into some strange cross-compile problem on OSX. I'll address this in another issue, then open a new PR for the docs.

Yes, I have noticed that. The cross-compile problem on OSX is not caused by this pull request.

lowzj:
merged manually via command line.

PR #426, opened 2019-02-25
fix: dfdaemon exit when load properties failed

lowzj:

Dfdaemon exit directly when load config file failed.

### DIFF
--- a/dfdaemon/constant/code.go
+++ b/dfdaemon/constant/code.go
@@ -16,22 +16,26 @@
 
 package constant
 
-var (
+const (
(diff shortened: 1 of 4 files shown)


dfdaemon/constant/code.go
```diff
@@ -17,6 +17,8 @@
 package constant
 
 var (
+	// CodeExitConfigError represents that the config provided can not be load successfully.
```
Starnop: Please use `iota` instead of it.

lowzj: done

Starnop:
LGTM.

PR #430, opened 2019-02-28
fix: don't exit when default config file is not exist

lowzj:

Dfdaemon doesn't exit when the default config file is not exist,  make it friendly for user.


Starnop:
LGTM.

PR #436, opened 2019-03-04
refactor: Extract the common utils to the common folder

Starnop:

With the beginning of the supernode development, we found that some util methods can be shared with between components. So I extract them as a common folder. 

When a file relies on both the util methods of its own and the common, and we will use ​​`cutil` as the alias of common util package which means `the common util`.

None.

Just refactor.

Pass the CI test.


lowzj:
LGTM

PR #439, opened 2019-03-05
doc: use refers to replace documents that repeat with web sites

Starnop:

I submit this PR because we maintain a copy of the same documents on the [website](https://github.com/dragonflyoss/website).
In the mean time, in order to let the users find documents better, I use a link to point to the website.

None.

Just docs.


Starnop:
It might be better to synchronize the documents to website in some way later. So close it.

PR #440, opened 2019-03-06
refactor: extract the common const code to the common folder

Starnop:

In the process of dfget interacting with supernode, and both parties reached an agreement on the return code in order to better understand each other. So we should make codes as a public part.

 For supernode, there are three kinds of status codes: 
+ the code from dfget: PeerTaskRequestResult, PeerTaskRequestStatus
+ the result code  returned to the dfget：ResultCode
+ the status information maintained in memory：PeerPieceStatus, PeerTaskStatus, CdnStatus, PreheatTaskStatus

And the code of `PeerTaskRequestResult` and `PeerTaskRequestStatus` from dfget will converted to the status of `PeerPieceStatus` and `PeerTaskStatus` by supernode.

<image width="480" height="550" src="https://user-images.githubusercontent.com/31209634/54409805-9aea8900-4723-11e9-9ccd-9c505643514c.png" />

None.

None.

I named this file as `dfget_super_code.go`. And we can also add a file named `dfget_daemon_code.go` to define the result code between `dfget` and `dfdaemon`  in the future.

In addition, I have drawn a very simple state code transformation diagram and for reference only.

<image  src="https://user-images.githubusercontent.com/31209634/54410064-d20d6a00-4724-11e9-925f-7f2ca2b6e772.png"/>

### DIFF
--- a/common/constants/dfget_super_code.go
+++ b/common/constants/dfget_super_code.go
@@ -1,0 +1,91 @@
+/*
+ * Copyright The Dragonfly Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package constants
+
+// This file defines the code required for both dfget and supernode.
+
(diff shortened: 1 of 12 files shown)


Starnop:
Do not merge it for now.

common/constants/dfget_super_code.go
```diff
@@ -0,0 +1,61 @@
...
+const (
+	// HTTPError represents that there is an error between client and server.
+	HTTPError = -100
+	// Success represents the request is success.
+	Success = 200
+)
+
+/* the response code from supernode */
+const (
+	CodePeerFinish = 600 + iota
```
lowzj: It's better to assign value to them explicitly. It's easy for us to search error code and avoid automatically changing value when we insert a new code into them.

Starnop:
@lowzj  It's ready to be merged. PTAL. THX. 😄

lowzj:
LGTM

Issue #139, opened 2018-09-14
supernode增加存储驱动(storage driver)


nonly:
**背景**
- 当前版本的supernode是将元信息存储在本地内存中；规模比较大时本地内存占用可能过多。

**想法**
- 希望增加一些常用存储的storage driver，例如redis等；用户可以根据需求选择不同的driver存储元信息；
- 目标是：可配置，可插拔，支持多种driver

zhouhaibing089:
Looks very straightforward to replace all the `ConcurrentHashMap` with an interface.

allencloud:
First of all, thanks sincerely for constantly using and supporting Dragonfly. We will try our best to keep Dragonfly better, and keep community and eco-system growing. 

To get more feedback of Dragonfly's adoption in industry, would you mind helping comment in issue https://github.com/dragonflyoss/Dragonfly/issues/306 to give more information about your enterprise's Dragonfly usage. Thanks a lot in advance. @nonly

nonly:
@allencloud  OK, I have commented.

Starnop:
Has been implemented. So close it now.

PR #446, opened 2019-03-12
feature: add storage driver for supernode

Starnop:

The supernode will use cdn to cache the files which downloaded from the source  web server. However, using local storage is not a one-size-fits-all approach. So we define an interface `StorageDriver` that defines all the storage  driver behavior to support different storage drivers.

In addition,  we may use multiple storage drivers for different contents at the same time. So there is a `Manager` to manage multiple storage drivers.

You can understand it easily according to the follow image:

And I have implement a local driver as a demo.

#139 

Added.

### DIFF
--- a/supernode/store/local_storage.go
+++ b/supernode/store/local_storage.go
@@ -1,0 +1,285 @@
+/*
+ * Copyright The Dragonfly Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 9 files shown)


supernode/store/local_storage.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	return &fileMutex{
+		path:     filePath,
+		fileInfo: f,
+	}, nil
+}
+
+func getPrefix(str string) string {
+	return string([]byte(str)[:3])
```
lowzj: Here should check whether the length of `str` is greater than 3.

Starnop:
All mentioned above have been updated

PR #471, opened 2019-04-03
feature: make supernode configurable for dfget with dfdaemon

Starnop:

Add a `super-nodes` flag for dfdaemon.
The reason to do this is that we can only configure the supernode which used by dfget through the configuration file(`/etc/dragonfly.conf` in default). It may be more convenient  in the test environment or the scenario that someone who just want to have a try with dragonfly quickly.

None.

None.

1. Start dfdaemon with `dfdaemon --super-nodes 127.0.0.1,192.168.33.12 --registry https://index.docker.io`.
NOTE: set your own supernode IP.

2. Configure Docker Daemon.

3. Use docker client to pull a test image: `docker pull busybox:latest`.

4. Check the logs with `grep "do register to" ~/.small-dragonfly/logs/dfclient.log`
If the output of command above has content like
```
2019-04-03 08:44:01.377 INFO sign:14707-1554281041.247 : do register to one of [127.0.0.1 192.168.33.12 127.0.0.1 192.168.33.12]
```
and then it proves that this PR works.

### DIFF
--- a/cmd/dfdaemon/app/options/options.go
+++ b/cmd/dfdaemon/app/options/options.go
@@ -84,6 +84,9 @@ type Options struct {
 	// ConfigPath is the path of dfdaemon's configuration file.
 	// default value is: /etc/dragonfly/dfdaemon.yml
 	ConfigPath string
+
+	// SupernodeList specify supernode list.
+	SupernodeList []string
 }
 
 // NewOption returns the default options.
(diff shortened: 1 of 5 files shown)


cmd/dfdaemon/app/options/options.go
```diff
@@ -125,6 +128,7 @@ func (o *Options) AddFlags(fs *flag.FlagSet) {
 	fs.StringVar(&o.DownRule, "rule", "", "download the url by P2P if url matches the specified pattern,format:reg1,reg2,reg3")
 	fs.BoolVar(&o.Notbs, "notbs", true, "not try back source to download if throw exception")
 	fs.StringSliceVar(&o.TrustHosts, "trust-hosts", o.TrustHosts, "list of trusted hosts which dfdaemon forward their requests directly, comma separated.")
+	fs.StringSliceVar(&o.SupernodeList, "super-nodes", []string{"127.0.0.1"}, "specify the addresses(IP:port) of supnernodes that will be passed to dfget.")
```
lowzj: Use `node` instead of `super-nodes`, it's consistent with `dfget`.
It's better not to set dfdaemon's option `super-nodes` a default value. If the this option is not set, then `dfget` use its own nodes.

Starnop: I agree with you. And that has been updated. PATL again.

dfdaemon/downloader/dfget/dfget.go
```diff
@@ -96,6 +99,9 @@ func (dfGetter *DFGetter) parseCommand(url string, header map[string][]string, n
 		args = append(append(args, "-s"), dfGetter.rateLimit)
 		args = append(append(args, "--totallimit"), dfGetter.rateLimit)
 	}
+	if dfGetter.supernodes != nil {
```
lowzj: It's better to check the length of `supernodes` is whether greater than zero. If it's not nil and its length is equals 0, dfget will execute failed.

Starnop: DONE.

lowzj:
LGTM

PR #483, opened 2019-04-10
feature: implement the progress mgr for get/delete info

Starnop:

For `ProgreeMgr`, it will maintain the correspondence between the peers and pieces as follows:

+ ClientID->PieceNums
+ PieceNum->PeerIDs

**NOTE:**

+ A `ClientID` corresponds to a unique peer and task.
+ A `PieceNum`  corresponds to a unique piece.
+ A `PeerID` corresponds to a unique peer node.

And the information should be readable and writable. So I updated the get and delete interfaces as follows:

```
// GetPieceProgressByCID get all pieces progress with specified clientID.
// And the pieceStatus should be one of the `PieceRunning`,`PieceSuccess` and `PieceAvailable`.
GetPieceProgressByCID(ctx context.Context, taskID, clientID, pieceStatus string) (pieceNums []int, err error)

// DeletePieceProgressByCID delete the pieces progress with specified clientID.
DeletePieceProgressByCID(ctx context.Context, taskID, clientID string) (err error)

// GetPeerIDsByPieceNum gets all peerIDs with specified taskID and pieceNum.
GetPeerIDsByPieceNum(ctx context.Context, taskID string, pieceNum int) (peerIDs []string, err error)

// DeletePeerIDByPieceNum deletes the peerID wich means that
// the peer no longer provides the service for the pieceNum of taskID.
DeletePeerIDByPieceNum(ctx context.Context, taskID string, pieceNum int, peerID string) error

```

In the mean time, `ProgreeMgr` will maintain some information of a peer.
```
// PeerState maintains peer related information.
type PeerState struct {
// PeerID identifies a peer uniquely.
PeerID string

// ProducerLoad is the load of download services provided by the current node.
ProducerLoad int32

// ClientErrorCount maintains the number of times that PeerID failed to downloaded from the other peer nodes.
ClientErrorCount int32

// ServiceErrorCount maintains the number of times that the other peer nodes failed to downloaded from the PeerID.
ServiceErrorCount int32

// ServiceDownTime the down time of the peer service.
ServiceDownTime int64
}
```
So I add the get and delete interfaces as follows:
```
// GetPeerStateByPeerID gets peer state with specified peerID.
GetPeerStateByPeerID(ctx context.Context, peerID string) (*PeerState, error)

// DeletePeerStateByPeerID deletes the peerState by PeerID.
DeletePeerStateByPeerID(ctx context.Context, peerID string) error

```

None.
Added.

Use `PieceNum` instead of `PieceRange` as the key of a piece.

### DIFF
--- a/supernode/daemon/mgr/progress_mgr.go
+++ b/supernode/daemon/mgr/progress_mgr.go
@@ -4,7 +4,26 @@ import (
 	"context"
 
 	"github.com/dragonflyoss/Dragonfly/apis/types"
+	cutil "github.com/dragonflyoss/Dragonfly/common/util"
 )
+
+// PeerState maintains peer related information.
+type PeerState struct {
+	// PeerID identifies a peer uniquely.
+	PeerID string
+
+	// ProducerLoad is the load of download services provided by the current node.
+	ProducerLoad int32
+
+	// ClientErrorCount maintains the number of times that PeerID failed to downloaded from the other peer nodes.
+	ClientErrorCount int32
+
+	// ServiceErrorCount maintains the number of times that the other peer nodes failed to downloaded from the PeerID.
+	ServiceErrorCount int32
+
+	// ServiceDownTime the down time of the peer service.
+	ServiceDownTime int64
+}
 
 // ProgressMgr is responsible for maintaining the correspondence between peer and pieces.
 type ProgressMgr interface {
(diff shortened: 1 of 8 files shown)


supernode/daemon/mgr/progress_mgr.go
```diff
@@ -6,6 +6,24 @@ import (
 	"github.com/dragonflyoss/Dragonfly/apis/types"
 )
 
+// PeerState maintains peer related information.
+type PeerState struct {
```
Starnop: @lowzj Maybe move it into PeerInfo? WDYT?

lowzj:
lgtm

PR #486, opened 2019-04-11
feature: add transparent proxy to extend dfdaemon's proxy mode

### LINKED ISSUE #449: proxy模式下https registry无法pull镜像
### Ⅰ. Issue Description
Setting HTTP_PROXY, the HTTPS request will also pass through proxy, so a CONNECT request will be sent to dfdaemon, which returns 404. As for registry and auth service domain names, they can't be used without NO_PROXY filtering.

// 设置HTTP_PROXY，https请求也会经过proxy，所以会向dfdaemon发送CONNECT请求，dfdaemon返回了404。对于registry和auth服务域名一样，没有办法通过NO_PROXY过滤的情况就不能用了。 

### Ⅱ. Describe what happened
image
image

***!!!!WE STRONGLY ENCOURAGE YOU TO DESCRIBE YOUR ISSUE IN ENGLISH!!!!***

### LINKED ISSUE #478: [feature request] Extend dfdaemon's proxy mode
## Why you need it?

It would be great if we can set `HTTP_PROXY=http://dfdaemon` and speed up HTTP requests for any third party tools that respects the `HTTP_PROXY` environment variable out of the box.

A good example would be speeding up package managers like yum, apt, npm, maven and others that can use HTTP download.

The current registry proxy can be implemented with this too.

HTTPS requests are more tricky, the best we can do at the moment might be just proxying it over HTTP, without looking at the content. This will handle cases where the client needs both HTTPS and HTTP requests like #449 

## How it could be?

We need a way to describe the proxy rules like `registries` in dfdaemon's config, for example:

```yaml
proxy:
  # a regexp that will be checked against the url, probably without the scheme since
  # we can only actually handle http requests
- match: regexp
  # change the proxy target and scheme, as the current registry proxy does.
  # maybe we can also provide a way rewrite host like the nginx rewrite syntax
  change_host: example.host
  change_scheme: https
  # like the urlfilter parameter, but works for individual rules, merged with global settings
  query_filters:
    - Signature
    - Expires
    - OSSAccessKeyId
  # whether to use p2p to download, like the rule cli parameter
  p2p: true
```

We can first read the rules from config file, and maybe provide apis to dynamically change the rules later.

inoc603:

Add a transparent proxy for dfdaemon. HTTP GET requests can be proxied with dfget if it matches the proxy rule. Other methods, including CONNECT for HTTPS requests will be proxied directly.

To handle CONNECT request properly, we can't use the default http mux. So I created a new http server dedicated for proxy, the port can be configured with `--proxyPort` cli flag, with a default value of `65002`. Proxy rules can be configured in dfdaemon.yml

fixes #478 
fixes #449 

The rule matching is tested in dfdaemon/proxy/proxy_test.go

I haven't write integration test yet.

Start dfdaemon with verbose logging and the following config file:

```yaml
proxies:
- regx: /blobs/sha256
- regx: github.com
  use_https: true
```

Set HTTP_PROXY of docker daemon to `http://127.0.0.1:65002`, and pull images. If the registry is an insecure registry, requests for image layers will be fetch with dfget. Otherwise requests will be proxied directly.

Look for these log lines:\
`round trip directly: $method $url` -> direct proxy \
`round trip with dfget: $url` -> with dfget

The `use_https` can be used to proxy registries with only https service. The proxy will change the request scheme before proxying it. You can test this with curl, although this is not how the proxy is meant to be used:

`curl -I http://github.com` -> returns 301 redirect to https://github.com
`http_proxy=http://127.0.0.1:65002 curl -I http://github.com` -> returns content of https://github.com directly

Of course you can set an https registry as insecure in docker daemon's config, and add a rule with `use_http: true` to match it, then pull images from it to see if it works. 

I haven't write any documentation yet. I'll add documentation when the code is reviewed.

### DIFF
--- a/cmd/dfdaemon/app/root.go
+++ b/cmd/dfdaemon/app/root.go
@@ -25,6 +25,7 @@ import (
 	"runtime"
 
 	"github.com/dragonflyoss/Dragonfly/cmd/dfdaemon/app/options"
+	g "github.com/dragonflyoss/Dragonfly/dfdaemon/global"
 	"github.com/dragonflyoss/Dragonfly/dfdaemon/initializer"
 	"github.com/dragonflyoss/Dragonfly/dfdaemon/proxy"
 
@@ -50,20 +51,27 @@ func init() {
 	opt.AddFlags(rootCmd.Flags())
 }
 
+func runTransparentProxy(opt *options.Options) error {
+	tp, err := proxy.New(g.Properties.Proxies)
+	if err != nil {
+		return fmt.Errorf("failed to create transparent proxy: %v", err)
+	}
+	s := http.Server{
+		Addr:    fmt.Sprintf(":%d", opt.ProxyPort),
+		Handler: tp,
+	}
+	logrus.Infof("launch dfdaemon transparent proxy on %s:%d", opt.HostIP, opt.ProxyPort)
+	go logrus.Fatalf("Transparent proxy stopped: %v", s.ListenAndServe())
+	return nil
+}
+
 // start to run dfdaemon server.
 func runDaemon() error {
 	initOption(opt)
 
-	s := http.Server{
-		Addr: ":65002",
-		Handler: proxy.New([]proxy.Rule{
-			{
-				Match: "/blobs/sha256/",
-			},
-		}),
+	if err := runTransparentProxy(opt); err != nil {
+		return err
 	}
-	fmt.Println(s.Addr)
-	go fmt.Println(s.ListenAndServe())
 
 	logrus.Infof("start dfdaemon param: %+v", opt)
 
--- a/cmd/dfdaemon/app/root.go
+++ b/cmd/dfdaemon/app/root.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/dragonflyoss/Dragonfly/cmd/dfdaemon/app/options"
 	"github.com/dragonflyoss/Dragonfly/dfdaemon/initializer"
+	"github.com/dragonflyoss/Dragonfly/dfdaemon/proxy"
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -52,6 +53,17 @@ func init() {
 // start to run dfdaemon server.
 func runDaemon() error {
 	initOption(opt)
+
+	s := http.Server{
+		Addr: ":65002",
+		Handler: proxy.New([]proxy.Rule{
+			{
+				Match: "/blobs/sha256/",
+			},
+		}),
+	}
+	fmt.Println(s.Addr)
+	go fmt.Println(s.ListenAndServe())
 
 	logrus.Infof("start dfdaemon param: %+v", opt)
 
--- a/cmd/dfdaemon/app/root.go
+++ b/cmd/dfdaemon/app/root.go
@@ -61,7 +61,9 @@ func runTransparentProxy(opt *options.Options) error {
 		Handler: tp,
 	}
 	logrus.Infof("launch dfdaemon transparent proxy on %s:%d", opt.HostIP, opt.ProxyPort)
-	go logrus.Fatalf("Transparent proxy stopped: %v", s.ListenAndServe())
+	go func() {
+		logrus.Fatalf("Transparent proxy stopped: %v", s.ListenAndServe())
+	}()
 	return nil
 }
 
--- a/dfdaemon/proxy/proxy.go
+++ b/dfdaemon/proxy/proxy.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/dragonflyoss/Dragonfly/dfdaemon/config"
 	"github.com/dragonflyoss/Dragonfly/dfdaemon/handler"
+
 	"github.com/sirupsen/logrus"
 )
 
@@ -21,6 +22,8 @@ func New(rules []*config.Proxy) (*TransparentProxy, error) {
 	return proxy, nil
 }
 
+// TransparentProxy is an http proxy handler. It proxies requests with dfget
+// if any defined proxy rules is matched
 type TransparentProxy struct {
 	rules []*config.Proxy
 }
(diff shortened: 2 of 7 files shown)


dfdaemon/proxy/proxy.go
```diff
@@ -0,0 +1,117 @@
...
+	defer resp.Body.Close()
+	copyHeader(w.Header(), resp.Header)
+	w.WriteHeader(resp.StatusCode)
+	io.Copy(w, resp.Body)
+}
+
+// ShouldUseDfget returns whether we should use dfget to proxy a request. It
+// also change the scheme of the given request if the matched rule has
+// UseHTTPS = true
+func (proxy *TransparentProxy) ShouldUseDfget(req *http.Request) bool {
```
Starnop: Just curious, why use capitalization here?

dfdaemon/proxy/proxy.go
```diff
@@ -0,0 +1,117 @@
...
+import (
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"time"
+
+	"github.com/dragonflyoss/Dragonfly/dfdaemon/config"
+	"github.com/dragonflyoss/Dragonfly/dfdaemon/handler"
+	"github.com/sirupsen/logrus"
```
Starnop: Could you please add a blank line here? THX 😄

dfdaemon/proxy/proxy.go
```diff
@@ -0,0 +1,117 @@
...
+			}
+			return !rule.Direct
+		}
+	}
+	return false
+}
+
+// HandleHTTPS handles a CONNECT request and proxy an https request through an
+// http tunnel.
+func HandleHTTPS(w http.ResponseWriter, r *http.Request) {
```
Starnop: The same as above, why use capitalization here?

Starnop:
@inoc603  FYI, you can use the `git commit --amend` to modify the previous commits to avoid submitting multiple commits. 

BTW, very glad for your contribution. 😺

dfdaemon/proxy/proxy.go
inoc603: I was thinking about using this in dfdaemon/handler/root_handler.go to fix the old reverse proxy, so I made this exported. But now I think maybe it's better to leave the old reverse proxy as it is, so the user will get errors if they used the wrong ports for proxy. I'll make this un-exported.

dfdaemon/proxy/proxy.go
inoc603: I was passing this to the dfget roundtripper as a parameter, so I exported this function for possible cross package usage. But it seems unnecessary now, I'll change it.

dfdaemon/proxy/proxy.go
inoc603: Fixed in the latest commit

dfdaemon/proxy/proxy.go
inoc603: Fixed in the latest commit

dfdaemon/proxy/proxy.go
inoc603: Added in the latest commit, but the the code snippet some how is not updated or marked outdated.

inoc603:
@Starnop Where should I write the documentation on proxy usage? I remember there's a blog on d7y.io, but not any instructions in the documentation.

lowzj:
You can open a pull request on repository [dragonflyoss/website](https://github.com/dragonflyoss/website) to add documents into [docs](https://github.com/dragonflyoss/website/tree/master/docs) folder.

inoc603:
> You can open a pull request on repository [dragonflyoss/website](https://github.com/dragonflyoss/website) to add documents into [docs](https://github.com/dragonflyoss/website/tree/master/docs) folder.

Is the docs in the website repo and the docs in this repo related? Should I update the docs in this repo too? @lowzj 

Should I add a section in [docs/user_guide/download_files.md](https://github.com/dragonflyoss/Dragonfly/blob/master/docs/user_guide/download_files.md) or add new file in [docs/user_guide](https://github.com/dragonflyoss/Dragonfly/tree/master/docs/user_guide)?

Starnop:
> > You can open a pull request on repository [dragonflyoss/website](https://github.com/dragonflyoss/website) to add documents into [docs](https://github.com/dragonflyoss/website/tree/master/docs) folder.
> 
> Is the docs in the website repo and the docs in this repo related? Should I update the docs in this repo too? @lowzj
> 
> Should I add a section in [docs/user_guide/download_files.md](https://github.com/dragonflyoss/Dragonfly/blob/master/docs/user_guide/download_files.md) or add new file in [docs/user_guide](https://github.com/dragonflyoss/Dragonfly/tree/master/docs/user_guide)?

I think it's better to add a new file in [docs/user_guide](https://github.com/dragonflyoss/website/tree/master/docs/en-us/userguide). WDYT? @lowzj 

BTW, could you please submit your pull request to the repo https://github.com/dragonflyoss/website?

cmd/dfdaemon/app/root.go
```diff
@@ -49,10 +51,28 @@ func init() {
...
+	tp, err := proxy.New(g.Properties.Proxies)
+	if err != nil {
+		return fmt.Errorf("failed to create transparent proxy: %v", err)
+	}
+	s := http.Server{
+		Addr:    fmt.Sprintf(":%d", opt.ProxyPort),
+		Handler: tp,
+	}
+	logrus.Infof("launch dfdaemon transparent proxy on %s:%d", opt.HostIP, opt.ProxyPort)
+	go logrus.Fatalf("Transparent proxy stopped: %v", s.ListenAndServe())
```
lowzj: `s.ListenAndServe()` in this line will block the subsequent flows, the following codes may be what you want:

```golang
	go func() {
		logrus.Fatalf("Transparent proxy stopped: %v", s.ListenAndServe())
	}()
```

inoc603: Fixed. Some how this kind of mistakes always come back and bite me every once in a while...

lowzj:
LGTM

inoc603:
I added a new doc file to describe how to use the proxy. There are some proxy related content in other docs, like FAQ, still needs to be updated.

lowzj:
@inoc603 It's glad to see the related document to describe how to use proxy feature of dfdaemon. Could you please to open an other pull request to add documents? Then I can easily sync the documents to repo `dragonflyoss/website`.

inoc603:
Sure, I've reverted the latest commit. I'll open another PR for the docs

lowzj:
This pull request is merged by command line manually to combine the 5 commits and add detailed description of this feature.
@inoc603 Thanks for your great works!

PR #501, opened 2019-04-18
Update documentation for the new transparent proxy

### LINKED ISSUE #478: [feature request] Extend dfdaemon's proxy mode

inoc603:

Update documentation for the new proxy mode

fixes #478

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -135,11 +135,9 @@ With updating the configuration, request `docker pull mysql:5.6` will be sent to
 
 > Note: please remember restarting container engine after updating configuration.
 
-## Can I set HTTP_PROXY to be dfdaemon address?
+## Can I set dfdaemon as HTTP_PROXY?
 
-Yes, HTTP_PROXY is used to proxy every request from docker engine, including requests `docker login`, `docker pull` and `docker push`. Once it is configured, the influenced parts could be expanded much more widely. Dragonfly's feature is only taking effect on `docker pull`. Thus every `docker pull` request will be sent to dfdaemon, no matter it is an official image from docker hub or from a third-party registry. However, everyone should keep in mind one point that even if user wish to pull image `a.b.com/mysql:5.6` from registry `a.b.com`, dfdaemon will pull the image `mysql:5.6` from the configured registry address of dfdaemon, such as `c.d.com`. As a result, user pulled image `mysql:5.6` from `c.d.com` rather than `a.b.com`.
-
-In addition, with `HTTP_PROXY` enabled, Dragonfly only supports protocol scheme `HTTP`. `HTTPs` will never be supported then.
+Yes, please refer to the [proxy guide](./docs/user_guide/proxy.md).
 
 ## Do we support HA of supernode in Dragonfly
 
--- a/docs/user_guide/proxy.md
+++ b/docs/user_guide/proxy.md
@@ -1,0 +1,85 @@
+# Use dfdaemon as an HTTP proxy
+
+Dfdaemon can be used as an HTTP proxy to speed up image pulling from any registry
+as well as general HTTP downloads.
+
+Please first ensure that you know how to install and run [supernode](install_server.md)
+and [dfdaemon](install_client.md).
+
+**HTTPS support is currently very limited. All HTTPS request will be tunneled
+directly, without dfget.**
+
+## Proxy rule configuration
+
+Proxy rules are configured in `/etc/dragonfly/dfdaemon.yml`. For performance
+reason, dfdaemon will handle a request with the the first matching rule.
+
+```yaml
+proxies:
+# proxy requests directly, without dfget
+- regx: no-proxy-reg
+  direct: true
+# proxy all http image layer download requests with dfget
+- regx: blob/sha256/.*
+# change http requests to some-registry to https, and proxy them with dfget
+- regx: some-registry/
+  use_https: true
+```
+
+## Download images
+
+Add the following content to `/etc/dragonfly/dfdaemon.yml`.
+
+```yaml
+proxies:
+# proxy all http image layer download requests with dfget
+- regx: blob/sha256/.*
+```
+
+Set HTTP_PROXY for docker daemon in `/etc/systemd/system/docker.service.d/http-proxy.conf`.
+`65002` is the default proxy port for dfdaemon. This can be configured with the
+`--proxyPort` cli parameter of dfdaemon.
+
+```
+[Service]
+Environment="HTTP_PROXY=http://127.0.0.1:65002"
+```
+
+Set your registry as insecure in `/etc/docker/daemon.json`
+
+```json
+{
+  "insecure-registries": [ "your.registry" ]
+}
+```
+
+Start dfdaemon and restart docker daemon.
+
+```
+systemctl restart docker
+```
+
+Pull an image to see if it works. For registries that are not configured
+insecure, you can still pull image from it, but dfdaemon will not be able to
+speed up your downloads with dfget.
+
+```
+docker pull nginx
+docker pull your.registry/team/repo:tag
+```
+
+Then you can [check if your image is downloaded with dfget](../../FAQ.md#how-to-check-if-block-piece-is-distributed-among-dfgets-nodes).
+
+## Download files
+
+You can simply use `HTTP_PROXY` environment variable to let dfdaemon download
+requests that match the proxy rules. This works for any program that
+respects the `HTTP_PROXY` environment variable.
+
+```
+HTTP_PROXY=http://127.0.0.1:65002 curl http://github.com
+```
+
+HTTPS requests and requests that are not matched, will be proxied directly,
+and dragonfly is not able to speed up them.
+


inoc603:
I think we should be more clear about when to use dfdaemon as transparent proxy, and when to use it as a reverse proxy. Probably need a dedicated section to describe it.

docs/user_guide/proxy.md
```diff
@@ -0,0 +1,94 @@
...
+
+You can simply use `HTTP_PROXY` environment variable to let dfdaemon download
+requests that match the proxy rules. This works for any program that
+respects the `HTTP_PROXY` environment variable.
+
+```
+HTTP_PROXY=http://127.0.0.1:65002 curl http://github.com
+```
+
+HTTPS requests and reqeusts that are not matched, will be proxied directly,
```
Starnop: s/reqeusts/requests

docs/user_guide/proxy.md
```diff
@@ -0,0 +1,94 @@
...
+up image download in a registry, you have to set it as insecure in `/etc/docker/daemon.json`.
+But rest assured that the content of authentication requests will still be encrypted.
+
+```json
+{
+  "insecure-registries": [ "your.registry" ]
+}
+```
+
+This requires that your registry can handle HTTP requests for image layers,
```
Starnop: @inoc603 FYI.

```
With insecure registries enabled, Docker goes through the following steps:

First, try using HTTPS.
If HTTPS is available but the certificate is invalid, ignore the error about the certificate.
If HTTPS is not available, fall back to HTTP.
```

And I have tried it locally and it couldn't support HTTPS registry in this way. So could you please show me something that prove you downloaded an image from HTTPS registry successfully? THX. 😄

inoc603: Thanks for point that out, I find that although the registry I used to test enabled https, but it uses oss for image storage and the image layer urls it returns are http. I thought it was the `insecure-registries` option that made all the url http, clearly it was a mistake.

If the registry uses different domain names for authentication and image download, it would be easy. However that may not always be the case, if it has only one domain, currently I don't have a simple way to tell a auth request from a image download request if they are https.

inoc603: But I do suggest that we clearly separate the usage of `port` and `proxyPort`. Use `port` for reverse proxy or `registry-mirrors`, and use `proxyPort` for `HTTP_PROXY`

inoc603: We can probably use a local CA to decrypt https traffic, but that's kind of hacky. The current https support is limited to "not breaking it", but does not speed them up with dfget.

inoc603: And I think if we manage to implement the man-in-the-middle https proxy, it should be optional, and we'll have different level of https support for the users:
- only http are supported, https gets blocked
- https gets tunneled directly
- https gets decrypted and speed up with dfget

Starnop: > We can probably use a local CA to decrypt https traffic, but that's kind of hacky. The current https support is limited to "not breaking it", but does not speed them up with dfget.

We have considered this option, but it is not an ideal approach.

Starnop: > We can probably use a local CA to decrypt https traffic, but that's kind of hacky. The current https support is limited to "not breaking it", but does not speed them up with dfget.

We have considered this approach before, and if there is really no better way, I think we have to use this method. 😭

Starnop: > But I do suggest that we clearly separate the usage of `port` and `proxyPort`. Use `port` for reverse proxy or `registry-mirrors`, and use `proxyPort` for `HTTP_PROXY`

I agree with you. However, it will make uses feel confused. What do you think of that  we use `port` to for  images and use `proxyPort` for ordinary files?

Starnop: > And I think if we manage to implement the man-in-the-middle https proxy, it should be optional, and we'll have different level of https support for the users:
> 
> * only http are supported, https gets blocked
> * https gets tunneled directly
> * https gets decrypted and speed up with dfget

I will submit an issue later,  we can do more discussion at that time. And I'll let you know at the first time.

Starnop:
And then, could you please update the doc about the support related to HTTPS registry so that I can merge this PR. THX 😄

tomliusc:
Local Practice Exercise
1.  configure domain name
cat  /etc/hosts
127.0.0.1 reg.com
127.0.0.1 other.com
2. dfdaemon0.3.0 need to read dfdaemon.yml files
cat /etc/dragonfly/dfdaemon.yml
registries:
    - regx: ^reg.com$
      schema: http
      host: harbor.dcos.xixian.unicom.local
    - regx: ^other.com$
      schema: http
      host: 10.124.142.132
3.  The docker configuration file needs to configure "insecure-registries", not "registry-mirrors".
cat /etc/docker/daemon.json
    "insecure-registries":["127.0.0.1:65001"]
4. No parameters startup
dfdaemon
5. Pull operation
[root@slave03 dragonfly]# docker pull other.com:65001/common/nginx:1.92
1.92: Pulling from common/nginx
51f5c6a04d83: Pull complete 
a3ed95caeb02: Pull complete 
640c8f3d0eb2: Pull complete 
a4335300aa89: Pull complete 
Digest: sha256:311e9840c68d889e74eefa18227d0a6f995bc7a74f5453fdcd49fe3c334feb24

inoc603:
> Local Practice Exercise
> 
> 1. configure domain name
>    cat  /etc/hosts
>    127.0.0.1 reg.com
>    127.0.0.1 other.com
> 2. dfdaemon0.3.0 need to read dfdaemon.yml files
>    cat /etc/dragonfly/dfdaemon.yml
>    registries:
>    
>    * regx: ^reg.com$
>      schema: http
>      host: harbor.dcos.xixian.unicom.local
>    * regx: ^other.com$
>      schema: http
>      host: 10.124.142.132
> 3. The docker configuration file needs to configure "insecure-registries", not "registry-mirrors".
>    cat /etc/docker/daemon.json
>    "insecure-registries":["127.0.0.1:65001"]
> 4. No parameters startup
>    dfdaemon
> 5. Pull operation
>    [root@slave03 dragonfly]# docker pull other.com:65001/common/nginx:1.92
>    1.92: Pulling from common/nginx
>    51f5c6a04d83: Pull complete
>    a3ed95caeb02: Pull complete
>    640c8f3d0eb2: Pull complete
>    a4335300aa89: Pull complete
>    Digest: sha256:311e9840c68d889e74eefa18227d0a6f995bc7a74f5453fdcd49fe3c334feb24

@tomliusc Thank you. But I think you meant to put it in [this pr](https://github.com/dragonflyoss/website/pull/35)

docs/user_guide/proxy.md
inoc603: > > But I do suggest that we clearly separate the usage of `port` and `proxyPort`. Use `port` for reverse proxy or `registry-mirrors`, and use `proxyPort` for `HTTP_PROXY`
> 
> I agree with you. However, it will make users feel confused. What do you think of that we use `port` to for images and use `proxyPort` for ordinary files?

Actually we can use the same port for both, but that would require some changes on the current http api, because we can't handle CONNECT requests with the current `mux` based approach. I didn't want to change the api to much at the time I implement the transparent proxy, so I used a new port, but it's certainly possible to do so.

inoc603:
> And then, could you please update the doc about the support related to HTTPS registry so that I can merge this PR. THX 😄

@Starnop updated, PTAL

FAQ.md
```diff
@@ -180,9 +178,8 @@ It is very easy to deloy Dragonfly in Kubernetes with [Helm](https://github.com/
...
 
-We **CANNOT** pull an image from a third-party registry via Dragonfly. Images from third-party registry means that the name of image has no registry address, for example, image `a.b.com/admin/mysql:5.6` is a third-party image, and it is from registry `a.b.com`. To the opposite, `admin/mysql:5.6` is not from third-party registry, because name of the image does not contain a registry address.
-
-Because administrator needs to configure `--registry-mirrors=["http://127.0.0.1:65001"]` for container engine to proxy part of image pulling requests to `dfdaemon` which listens on 65001 locally, images not from a third-part registry will use this registry mirror. However, when user pulls image `a.b.com/admin/mysql:5.6` which contains a third-party registry address, the container engine will directly access the third-party registry to pull the image, ignoring the configured registry mirror. Then It has nothing to do with the `dfdaemon` within Dragonfly.
+Yes, this can be achieved by [using dfdaemon as docker daemon's HTTP proxy](./docs/user_guide/proxy.md#download-images).
+**However if the registry has HTTPS enabled, currently images will be pulled without dfget.**
```
Starnop: The title of this question is `Can an image from a third-party registry be pulled via Dragonfly`, so it may be paradoxical to put this sentence here. Could you please keep it as it is, and I will update it later. THX.

docs/user_guide/proxy.md
```diff
@@ -0,0 +1,86 @@
...
+
+Dfdaemon can be used as an HTTP proxy to speed up image pulling from any registry
+as well as general HTTP downloads.
+
+Please first ensure that you know how to install and run [supernode](install_server.md)
+and [dfdaemon](install_client.md).
+
+**HTTPS support is currently very limited**. All HTTPS request will be tunneled
+directly, without dfget. If you want to use HTTPS enabled registries, we suggest
+that you use HTTP for your images layers, and HTTPS for other apis like authentication.
```
Starnop: In fact, as far as I know, only `registry.**-**.aliyuncs.com` registries that has done like that, and it is unreasonable for users to modify the image registries.

docs/user_guide/proxy.md
```diff
@@ -0,0 +1,86 @@
...
+respects the `HTTP_PROXY` environment variable.
+
+```
+HTTP_PROXY=http://127.0.0.1:65002 curl http://github.com
+```
+
+HTTPS requests and requests that are not matched, will be proxied directly,
+and dragonfly is not able to speed up them.
+
+## Proxy rule configuration
```
Starnop: How about introduce the configuration first, and then introduce how to use it when downloading? Just like this

## Proxy rule configuration
## Download files with Proxy
    ### Download images
    ### Download original files

FAQ.md
inoc603: With `HTTP_PROXY` set to the transparent proxy, third party images can be pulled via dragonfly, though not necessarily with dfget. But I agree that we need better instructions for how to use dragonfly with third party registries, and I'll just revert it. It will be much easier if we solved the https hijacking problem.

docs/user_guide/proxy.md
inoc603: Sure. I removed the registry setting suggestion.

docs/user_guide/proxy.md
inoc603: I tend to put examples first and leave the more technical details last, but I think introduce the configuration first here is fine, since it's relatively short.

Starnop:
LGTM.

allencloud:
Thanks for your work. Merging...

PR #502, opened 2019-04-19
feature: register the supernode as a peer before starting the server

Starnop:

Add supernode info as the config of supernode daemon, and register the supernode as a peer before starting the daemon server.

None.

None.

### DIFF
--- a/cmd/supernode/app/root.go
+++ b/cmd/supernode/app/root.go
@@ -130,11 +130,13 @@ func runSuperNode() error {
 		return err
 	}
 
-	if err = d.Run(); err != nil {
-		logrus.Errorf("failed to run daemon: %v", err)
+	// register supernode
+	if err := d.RegisterSuperNode(); err != nil {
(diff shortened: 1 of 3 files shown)


cmd/supernode/app/root.go
```diff
@@ -130,7 +130,13 @@ func runSuperNode() error {
...
 
-	if err = d.Run(); err != nil {
+	// register supernode
+	if err := d.RegisterSuperNode(); err != nil {
+		logrus.Errorf("failed to register super node: %v", err)
+		return err
+	}
+
+	if err := d.Run(); err != nil {
 		logrus.Errorf("failed to run daemon: %v", err)
```
lowzj: There is a similar error log in function `d.Run()`, maybe it's better to remove it from `d.Run()`.

Starnop: Updated.

lowzj:
LGTM

Issue #348, opened 2019-01-23
Plan of supernode's refactoring plan


allencloud:
## Background

This is one part of [Golang Refactoring Plan of the Whole Dragonfly Project](https://github.com/dragonflyoss/Dragonfly/issues/346)

Supernode is the brain of Dragonfly. It takes over the image CDN from remote registry, and schedules peer node according the actual situation of themselves. In addition, flow control, security guarantee are also what it is responsible for. And supernode is also supposed to de designed with HA(High Availability).

### APIs

- [ ] Peer Related
    - [ ] POST /peers: register a peer node in supernode.
    - [ ] GET /peers/{id}: Get information for one single peer node;
- [ ] Task Related
    - [ ] POST /tasks: register a file/image downloading task;
    - [ ] GET /tasks/{id}/pieces: client side requests to get the pieces list of the downloading task;
    - [ ] PUT /tasks/{id}/pieces/{pieceRange}:  Update the status of each piece in the downloading task;
    - [ ] GET /tasks/{id}: Get information for one single downloading task;
    - [ ] PUT /tasks/{id}: Update the status of downloading task;
    - [ ] DELETE /tasks/{id}: Delete a downloading task and its local cache on supernode
- [ ] Preheat related
    - [ ]  POST /preheats: Creates preheat task in supernode to cache a file/image first before starting distribution
    - [ ] GET /preheats/{id}: Get the status of preheat task  
    - [ ] GET /preheats: Get preheat task list

### CLI ( Command Line )

- [ ] Use [cobra](https://github.com/spf13/cobra/) to construct supernode's CLI.
- [ ] detailed description of every essential cobra command's fields

PR #504, opened 2019-04-22
feature: add getter function for supernode cdn file path

Starnop:

Supernode will store the downloaded files as a cache with CDN. And this pull request define some file paths that CDN will use. 

fixes part of #348

Added.

### DIFF
--- a/supernode/config/global_variable.go
+++ b/supernode/config/global_variable.go
@@ -1,0 +1,37 @@
+/*
+ * Copyright The Dragonfly Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


supernode/config/global_variable.go
```diff
@@ -0,0 +1,37 @@
...
+var (
+	// HTTPSubPath is the name of the parent directory where the upload files are stored.
+	HTTPSubPath = "qtdown"
+
+	// DownloadSubPath is the name of the parent directory where the downloaded files are stored.
+	DownloadSubPath = "download"
+
+	// DownloadHome is the parent directory where the downloaded files are stored
+	// which is a relative path.
+	DownloadHome = path.Join("/repo", DownloadSubPath)
```
lowzj: Don't use `/repo` as the root directory, it should be configured by user.
There's no necessary to add `/` here, it's different in different operating system.

Starnop: Currently the baseDir is supported user-defined. As for `DownloadHome` and `UploadHome`, they will be used as the bucket. HDYT?

lowzj: It's better to remove the prefix `/` if it's a relative path.

Starnop:
Implement it in the pull request #508 , so close it now.

supernode/config/global_variable.go
Starnop: Yeah, that's what I did in the pull request #508.

PR #505, opened 2019-04-22
feature: add httpFileLength field for taskInfo

Starnop:

A task not only maintains the source file size,  but also maintains a total file size after handled into pieces. 

In fact, The existing field `fileLength` indicates that the total file size after handled into pieces. So I add a filed named `httpFileLength` to maintain the source file size.

None.

None.


lowzj:
lgtm

PR #510, opened 2019-04-26
 feature: add proxy dns server

upengs:
支持harbor https private public
docker run -d --name dfclient1 -p 65001:65001 -v /etc/dragonfly.conf:/etc/dragonfly.conf dfclient:wp --registry https://x.x.x --resolver 10.0.0.120

supornode 使用 host 网络模式
docker run -d --name supernode --restart=always --net host     dragonflyoss/supernode:0.3.0 -Dsupernode.advertiseIp=10.0.0.124

在echo "nameserver 10.0.0.120" >>/etc/resolv.conf

[root@localhost ~]# docker logout   http://127.0.0.1:65001
Removing login credentials for 127.0.0.1:65001
[root@localhost ~]# docker pull   127.0.0.1:65001/library/python
Using default tag: latest
Error response from daemon: pull access denied for 127.0.0.1:65001/library/python, repository does not exist or may require 'docker login'
[root@localhost ~]# docker login   http://127.0.0.1:65001
Username: admin
Password:
WARNING! Your password will be stored unencrypted in /root/.docker/config.json.
Configure a credential helper to remove this warning. See
https://docs.docker.com/engine/reference/commandline/login/#credentials-store

Login Succeeded
[root@localhost ~]# docker pull   127.0.0.1:65001/library/python
Using default tag: latest
latest: Pulling from library/python
Digest: sha256:ce2b6fbe7d9dc28bbd332a6b9349b7040f0ffde81f1f14e121085eacfc4c0c3c
Status: Image is up to date for 127.0.0.1:65001/library/python:latest
[root@localhost ~]# curl http://127.0.0.1:65001/env
map[dfPattern:[] home:/root/ param:{DfPath:/dfclient/dfget DFRepo:/root/.small-dragonfly/dfdaemon/data/ RateLimit:20M CallSystem:com_ops_dragonfly URLFilter:Signature&Expires&OSSAccessKeyId Notbs:true HostIP:127.0.0.1 Registry:https://x.x.x TrustHosts:map[] Resolver:10.0.0.120}]

### DIFF
--- a/dfdaemon/handler/transport.go
+++ b/dfdaemon/handler/transport.go
@@ -17,7 +17,9 @@
 package handler
 
 import (
+	"context"
 	"crypto/tls"
+	"fmt"
 	"net"
 	"net/http"
 	"os"
@@ -54,9 +56,17 @@ func NewDFRoundTripper(cfg *tls.Config) *DFRoundTripper {
 		cfg = &tls.Config{InsecureSkipVerify: true}
 	}
 
+	var resolver = net.DefaultResolver
+	if global.CommandLine.Resolver != "" {
+		resolver = &net.Resolver{PreferGo: true, Dial: func(ctx context.Context, network, address string) (conn net.Conn, e error) {
+			dialer := &net.Dialer{}
+			return dialer.DialContext(ctx, "udp", fmt.Sprintf("%s:53", global.CommandLine.Resolver))
+		}}
+	}
 	return &DFRoundTripper{
 		Round: &http.Transport{
 			DialContext: (&net.Dialer{
+				Resolver:  resolver,
 				Timeout:   10 * time.Second,
 				KeepAlive: 30 * time.Second,
 			}).DialContext,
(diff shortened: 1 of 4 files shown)


Starnop:
@upengs Could you please describe what problem this PR is trying to solve？ In my opinion, it allows users to customize the DNS resolver server and avoid DNS Spoofing problems, etc? 

> 在echo "nameserver 10.0.0.120" >>/etc/resolv.conf

In addition, there's the question, should this command be executed on the client or supernnode?

upengs:
@Starnop When using internal harbor, HTTPS domain names need to specify a well-configured DNS server, otherwise dfdaemon will fail in reverse proxy. Supernode also needs to set up a DNS server. Execute echo "nameserver 10.0.0.120" >/etc/resolv.conf in supernode

dfdaemon/handler/transport.go
```diff
@@ -54,9 +56,19 @@ func NewDFRoundTripper(cfg *tls.Config) *DFRoundTripper {
 		cfg = &tls.Config{InsecureSkipVerify: true}
 	}
 
+	var resolver *net.Resolver
```
Starnop: It's better to do like this, WDYT?   

```
var resolver = net.DefaultResolver
if global.CommandLine.Resolver != "" {
....
}
```

upengs: @Starnop   Revised

allencloud:
Could you help to sign the DCO which fails in the integration test? @upengs

upengs:
Today's test did not find this problem.

PR #515, opened 2019-05-07
feature: add a function to determine whether the resource has expired

Starnop:

None.

None.

### DIFF
--- a/common/util/http_util.go
+++ b/common/util/http_util.go
@@ -308,3 +308,15 @@ func IsSupportRange(url string, headers map[string]string) (bool, error) {
 	}
 	return false, nil
 }
+
+// GetContentLength send a head request to get file length.
+func GetContentLength(url string, headers map[string]string) (int64, int, error) {
+	// send request
+	resp, err := HTTPWithHeaders("HEAD", url, headers)
+	if err != nil {
+		return 0, 0, err
+	}
+	resp.Body.Close()
+
+	return resp.ContentLength, resp.StatusCode, nil
+}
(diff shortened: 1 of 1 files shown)


common/util/http_util.go
```diff
@@ -253,3 +259,34 @@ func CheckConnect(ip string, port int, timeout int) (localIP string, e error) {
...
+	if err != nil {
+		return true, err
+	}
+	resp.Body.Close()
+
+	if resp.StatusCode == http.StatusNotModified {
+		return false, nil
+	}
+
+	return true, nil
```
lowzj:
```go
return resp.StatusCode != http.StatusNotModified, nil
```

common/util/http_util.go
```diff
@@ -253,3 +259,34 @@ func CheckConnect(ip string, port int, timeout int) (localIP string, e error) {
...
+		headers["If-Modified-Since"] = strconv.FormatInt(lastModified, 10)
+	}
+	if !IsEmptyStr(eTag) {
+		headers["If-None-Match"] = eTag
+	}
+
+	// send request
+	resp, err := HTTPWithHeaders("HEAD", url, headers)
+	if err != nil {
+		return true, err
```
lowzj: If `HEAD` is failed, then there is a high probability that `GET` will failed. At this moment, we may use the cached data rather than downloading it again.

Starnop: Has been updated. PTAL.

lowzj:
LGTM

PR #521, opened 2019-05-09
docs: correct some abused phrase

cindy1764:
by default and in default have totally different meaning. Here in Dragonfly should we use by default.

docs: correct some abused phrase


allencloud:
LGTM

PR #528, opened 2019-05-14
bugfix: use advertise-ip as the supernode ip when specified

Starnop:

### DIFF
--- a/cmd/supernode/app/root.go
+++ b/cmd/supernode/app/root.go
@@ -97,7 +97,7 @@ func setupFlags(cmd *cobra.Command, opt *Options) {
 	flagSet.IntVar(&opt.PeerDownLimit, "down-limit", opt.PeerDownLimit,
 		"download limit for supernode to serve download tasks")
 
-	flagSet.StringVar(&cfg.AdvertiseIP, "advertise-ip", "",
+	flagSet.StringVar(&opt.AdvertiseIP, "advertise-ip", "",
 		"the supernode ip that we advertise to other peer in the p2p-network")
 }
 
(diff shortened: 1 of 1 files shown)


cmd/supernode/app/root.go
```diff
@@ -174,11 +175,11 @@ func setAdvertiseIP() error {
 	if err != nil {
 		return errors.Wrapf(errorType.ErrSystemError, "failed to get ip list: %v", err)
 	}
-	if len(ipList) != 0 {
-		cfg.AdvertiseIP = ipList[0]
+	if len(ipList) == 0 {
+		return errors.Wrapf(errorType.ErrSystemError, "get ip list empty")
```
Starnop: @lowzj  should we use `127.0.0.1` or returns error here? WDYT?

Starnop: As for Java version of Dragonfly, that's exactly what we did as a solution. And for Golang version, we should find a better way to set the IP of supernode to support for multiple networks or something else. And for now, we use either user-specified AdvertiseIP or the first non-loop address if the AdvertiseIP is empty as a temporary solution. WDYT?

lowzj: It can be set in the following order:
1. user-specified advertiseIP
2. first non-loop address
3. the parameter `superNodeIp` of the first register request from dfget

Starnop: I understand what you said. And I will do the third item in the pr #457. 
Just return here that I will do.

Starnop: Has been updated. PTAL. THX >

lowzj:
LGTM

PR #531, opened 2019-05-16
feature: add realMd5 field for taskInfo

Starnop:

We should keep two MD5 values for taskInfo.
+ Md5. It's passed by caller which means the expected md5 value of the task file. For example, dfget will catch it from dfget's CLI and pass it to supernode. 
+ RealMd5. It's the real MD5 sum value which calculated by the downloaded file. 

When supernode finishes downloading file/image from the source location, and the supernode will validate the RealMd5 value with the Md5 value to check whether the downloaded file is valid.

None.
None.


lowzj:
LGTM

PR #536, opened 2019-05-17
feature: add dfdaemon.Server

### LINKED ISSUE #534: [feature] Use the same port for reverse proxy and transparent proxy
## Why you need it?

Currently we use 2 different ports for the reverse proxy which mainly used as registry mirror, and the transparent proxy which is meant to be used for `HTTP_PROXY`.  These two proxies' function have some overlap, and it may be confusing for the user. Technically, we can use the same port for both and make configuration simpler. `/debug` and other handlers can also be served with this port.

## How it could be?

We can easily tell a proxy request from a normal request from the request URL. It's very easy to implement, but some changes need to be made about the current configuration.

I propose using reverse proxy only for the official registry mirror. As recommended in https://github.com/dragonflyoss/website/pull/35 , multiple registries are supported via `HTTP_PROXY`. So we can change the `registries` configuration back to configure the reverse mirror only, and all `HTTP_PROXY` related configuration goes to `proxies`.

```yaml
registry_mirror:
  remote: https://index.docker.io
  # optional certificates. maybe we don't need it since private registries are meant to
  # be supported with http_proxy ?
  certs: []
  # maybe we should add a regexp to filter requests?
  # or we can use blobs/sha256 by default

proxies:
  # ...

hijack_https:
  # ...
```

inoc603:

Add dfdaemon.Server and use the same port for transparent proxy and reverse proxy.

Configurations for registry mirror is changed accordingly.

fixes #534 

Use `65001` instead of `65002` for the transparent proxy.

The new way of setting registry mirror in `dfdaemon.yml`:

```yaml
registry_mirror:
  remote: https://index.docker.io
  insecure: false
  certs: []
```

I haven't change the `registries` configuration as suggested in #534. This is currently a proof of concept to demonstrate that the idea will work.

I didn't change the use of global variables in the dfdaemon package, but I think it's necessary to remove them to make reasoning about the code and writing tests easier. Maybe I'll open another PR for it.


dfdaemon/server.go
```diff
@@ -0,0 +1,122 @@
...
+type Server struct {
+	server *http.Server
+	proxy  *proxy.TransparentProxy
+}
+
+// Option is the functional option for creating a server
+type Option func(s *Server) error
+
+// WithTLSConfig sets the tls config for the server
+func WithTLSConfig(config *tls.Config) Option {
```
lowzj: It's unused in the whole project.

inoc603: I meant to use this in tests or for other more complex scenarios, but it doesn't seem very useful at the moment. I'll remove it for now

dfdaemon/proxy/proxy.go
```diff
@@ -66,15 +77,28 @@ func New(opts ...Option) (*TransparentProxy, error) {
 	return proxy, nil
 }
 
+// NewFromConfig returns a new transparent proxy from the given properties
+func NewFromConfig(c config.Properties) (*TransparentProxy, error) {
+	opts := []Option{
+		WithRules(c.Proxies),
+		WithHTTPSHosts(c.HijackHTTPS.Hosts...),
```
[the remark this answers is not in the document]
inoc603: I'll add a check before using `HijackHTTPS`. Using struct instead of pointer is also a valid choice, but I think using pointer will be easier to tell whether user has configured the field.

inoc603:
@lowzj any thoughts on the `registries` field I mentioned before?

lowzj:
@inoc603 yeah, `registry_mirror` and `proxies` are much clearer and better than before, we're very looking forward to if you want to keep working on this.

inoc603:
@lowzj the `registry` cli parameter currently has no effect, should we remove it or use it as a short hand for `registries.remote` ?

lowzj:
@inoc603 It's better to remain `registry` option, and it has high priority than config.

inoc603:
> @inoc603 It's better to remain `registry` option, and it has higher priority than config.

Sure. In order to do this, I changed the default value for the flag to an empty string. PTAL

inoc603:
> @inoc603 It's better to remain `registry` option, and it has higher priority than config.

Sure. In order to do this, I changed the default value for the flag to an empty string. PTAL

lowzj:
@inoc603  The `registry` option and `registry_mirror` config are ineffective now, maybe somewhere is wrong.

inoc603:
> @inoc603 The `registry` option and `registry_mirror` config are ineffective now, maybe somewhere is wrong.

My bad. I forgot to pass the mirror config to proxy. It should be fixed now.

lowzj:
@inoc603 Thanks for your excellent  work!! Could you rebase your code and push again?

inoc603:
@lowzj Rebased.

PR #547, opened 2019-05-24
Update Dockerfile for better caching

cd1989:

Adjust Dockerfile for better caching

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
 FROM golang:1.10.4-alpine as builder
 
 WORKDIR /go/src/github.com/dragonflyoss/Dragonfly
-COPY . /go/src/github.com/dragonflyoss/Dragonfly
-
 RUN apk --no-cache add bash make gcc libc-dev git
+
+COPY . /go/src/github.com/dragonflyoss/Dragonfly
 
 # go build dfdaemon and dfget.
 # write the resulting executable to the dir /dfclient.


Dockerfile
```diff
@@ -1,10 +1,10 @@
 FROM golang:1.10.4-alpine as builder
 
 WORKDIR /go/src/github.com/dragonflyoss/Dragonfly
-COPY . /go/src/github.com/dragonflyoss/Dragonfly
-
 RUN apk --no-cache add bash make gcc libc-dev git
 
+COPY . /go/src/github.com/dragonflyoss/Dragonfly
```
Starnop: @cd1989 Maybe there is something that I don't know. Could you please tell me that what's the difference between them?

dfdaemon/config/config.go
```diff
@@ -67,9 +67,9 @@ func NewProperties() *Properties {
 //	 # key pair used to hijack https requests
 //       cert: df.crt
 //       key: df.key
-//	 hosts:
+//	     hosts:
```
Starnop: You want to add a space symbol not a tab here? 😢

dfdaemon/config/config_test.go
```diff
@@ -195,3 +196,14 @@ func (s *ConfigTestSuite) TestRegistry_initTLSConfig(c *check.C) {
 		}
 	}
 }
+
+func TestA(t *testing.T) {
```
Starnop: Maybe it's not a good idea to use local files on your own machine.

cd1989: Sorry, my mistake, i have commit some debug code.

Dockerfile
cd1989: `COPY . /go/src/github.com/dragonflyoss/Dragonfly` layer always change, but layer `RUN apk --no-cache add bash make gcc libc-dev git` not. If this static layer is put after the dynamic one, it can't take advantage of cache of the static layer.

Starnop:
LGTM.

PR #551, opened 2019-05-27
feature: add download port flag

Starnop:

Supernode will listen on two ports:
+ The register port (default: 8002) 
+ The download port (default: 8001)

The register port is used for clients to register their own information into supernode.
The download port is used for clients to download pieces from supernode.

And we currently lack of the download port flag.

None.

None.

### DIFF
--- a/cmd/supernode/app/root.go
+++ b/cmd/supernode/app/root.go
@@ -70,6 +70,9 @@ func setupFlags(cmd *cobra.Command, opt *Options) {
 	flagSet.IntVar(&opt.ListenPort, "port", opt.ListenPort,
 		"ListenPort is the port supernode server listens on")
 
+	flagSet.IntVar(&opt.DownloadPort, "download-port", opt.DownloadPort,
+		"DownloadPort is the port for download files from supernode")
+
 	flagSet.StringVar(&opt.HomeDir, "home-dir", opt.HomeDir,
 		"HomeDir is working directory of supernode")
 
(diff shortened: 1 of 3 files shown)


supernode/daemon/daemon.go
```diff
@@ -50,7 +50,7 @@ func (d *Daemon) RegisterSuperNode() error {
 	req := &types.PeerCreateRequest{
 		IP:       strfmt.IPv4(d.config.AdvertiseIP),
```
Starnop: @lowzj  FYI. For now, we only support the local storage file server. In the future, we should make it possible to config the IP the remote file server.

cmd/supernode/app/root.go
```diff
@@ -70,6 +70,9 @@ func setupFlags(cmd *cobra.Command, opt *Options) {
 	flagSet.IntVar(&opt.ListenPort, "port", opt.ListenPort,
 		"ListenPort is the port supernode server listens on")
 
+	flagSet.IntVar(&opt.FileServerPort, "file-server-port", opt.FileServerPort,
```
lowzj: We should unify the names of supernode's ports in the documents and codes.
In docs: `registerPort(8002)`, `downloadPort(8001)`, and here they're: `ListenPort(8002)`, `FileServerPort(8001)`. Which one do you prefer?

Starnop: I prefer keep the 8002 with `ListenPort` and use `downloadPort` for 8001. WDYT?

PR #556, opened 2019-05-28
feature: update integration test with supernode from java to go

Starnop:


Starnop:
> I think that it's better to remain the java version of `supernode` until the golang version is completely ready.

It's not an interesting thing. And I will close it firstly and open it again when the golang version is ready. 😄

PR #560, opened 2019-05-28
feature: store mgr return localStorage as default StorageDriver

lowzj:

* the manager of every plugin module has the responsibility to return the default implementation
* store manager returns the `localStorage` as its default implementation

### DIFF
--- a/supernode/store/store_mgr.go
+++ b/supernode/store/store_mgr.go
@@ -18,6 +18,8 @@ package store
 
 import (
 	"fmt"
+	"path"
+	"sync"
 
 	"github.com/dragonflyoss/Dragonfly/supernode/config"
 	"github.com/dragonflyoss/Dragonfly/supernode/plugins"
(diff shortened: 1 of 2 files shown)


supernode/store/store_mgr.go
```diff
@@ -38,21 +40,52 @@ func Register(name string, builder StorageBuilder) {
 
 // Manager manage stores.
 type Manager struct {
+	cfg *config.Config
```
Starnop: HomeDir is enough instead of the entire config. WDYT?

lowzj: `HomeDir` is enough for this storage manager, but there will be several different managers required different fields of `Config` but have similar behaviors, we may extract an abstract common implementation from them. And IMOP, the initialized `config.Config` instant should be visible for these managers.

supernode/store/store_mgr.go
```diff
@@ -38,21 +40,52 @@ func Register(name string, builder StorageBuilder) {
...
+
+func (sm *Manager) getDefaultStorage() (*Store, error) {
+	if sm.defaultStorage != nil {
+		return sm.defaultStorage, nil
+	}
+
+	sm.mutex.Lock()
+	defer sm.mutex.Unlock()
+
+	if sm.defaultStorage == nil {
```
allencloud:
```
if sm.defaultStorage != nil {
    return sm.defaultStorage, nil
} 
if sm.cfg == nil {
    return nil, fmt.Errorf("cannot init local storage without home path")
}
......
return sm.defaultStorage, nil
```
How about using this to reduce ident?

lowzj: actually, the initial version is as same as you wrote above...

Starnop:
LGTM.

Issue #544, opened 2019-05-23
Bug: dfget(golang) finished successfully for resources which not exist.


xq2248:
### Ⅰ. Issue Description
Bug: dfget(golang) finished successfully for resources which not exist. Dfget generates a tiny file of 169 bytes. But we do not need it. Maybe we can reports the file does not exist or delete the tiny file after the downloading ends.

### Ⅱ. Describe what happened
We download a zip file which not exists, the command as follows:
dfget -u http://10.47.231.40/xxx.zip 
[root@node2 ~]# dfget -u http://10.47.231.40/xxx.zip -n 10.47.231.22:8005 
--2019-05-23 20:21:13--  http://10.47.231.40/xxx.zip
dfget version:0.3.0
workspace:/root/.small-dragonfly sign:18913-1558614073.266
download from source
download SUCCESS(0) cost:0.010s length:169 reason:1
[root@node2 ~]# ll xxx.zip 
-rw------- 1 root root 169 May 23 20:21 xxx.zip

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1.
2.
3.

### Ⅵ. Environment:

- dragonfly version:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

xq2248:
[root@node2 ~]# dfget -u http://10.47.231.40/xxx.zip -n 10.47.231.22         
--2019-05-27 09:47:33--  http://10.47.231.40/xxx.zip
dfget version:0.3.0
workspace:/root/.small-dragonfly sign:1342-1558921653.695
client:10.47.231.25 connected to node:10.47.231.22
start download by dragonfly
download from source
download SUCCESS(0) cost:1.355s length:169 reason:3
[root@node2 ~]# cat xxx.zip 
<html>
<head><title>404 Not Found</title></head>
<body bgcolor="white">
<center><h1>404 Not Found</h1></center>
<hr><center>nginx/1.12.1</center>
</body>
</html>

dfclient.log as follows:
2019-05-27 09:47:33.699 INFO sign:1342-1558921653.695 : use peer server on port:61405
2019-05-27 09:47:33.699 INFO sign:1342-1558921653.695 : do register to one of [10.47.231.22 10.47.231.22]
2019-05-27 09:47:33.712 INFO sign:1342-1558921653.695 : do register to 10.47.231.22, res:{"code":200,"data":{"taskId":"480b
[...]

PR #561, opened 2019-05-29
fix: BackDownloader only download file when response is less than 400

lowzj:

When the response code is greater or equals 400, it means that there's something wrong, so `BackDownloader` shouldn't download the file.

merge to branch `master` and `0.3.x`.

fixes: #544

### DIFF
--- a/dfget/core/downloader/back_downloader/back_downloader.go
+++ b/dfget/core/downloader/back_downloader/back_downloader.go
@@ -103,6 +103,10 @@ func (bd *BackDownloader) Run() error {
 	}
 	defer resp.Body.Close()
 
+	if !bd.isSuccessStatus(resp.StatusCode) {
+		return fmt.Errorf("failed to download from source, response code:%d", resp.StatusCode)
+	}
+
 	buf := make([]byte, 512*1024)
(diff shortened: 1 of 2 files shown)


dfget/core/downloader/back_downloader/back_downloader.go
```diff
@@ -129,3 +133,7 @@ func (bd *BackDownloader) Cleanup() {
 	}
 	bd.cleaned = true
 }
+
+func (bd *BackDownloader) isSuccessStatus(code int) bool {
+	return code < 400
```
Starnop: `code < 400` -> `code <= 400`?

lowzj: `400` is also an error code caused by user

Starnop: Sorry for my mistake that I consider the opposite. 😢

Starnop:
LGTM.

dfget/core/downloader/back_downloader/back_downloader.go
xq2248: This is not enough. Only 200~299  is considered as success. HTTP espone codes:
1xx : continue
2xx: success
3xx: redirect
4xx: client errors
5xx: server internal errors

PR #568, opened 2019-05-29
bugfix: add locker util and fix thread-unsafe problems for cdn

Starnop:

### DIFF
--- a/supernode/util/locker.go
+++ b/supernode/util/locker.go
@@ -1,0 +1,80 @@
+/*
+ * Copyright The Dragonfly Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


supernode/util/locker.go
```diff
@@ -0,0 +1,78 @@
...
+	defer l.Unlock()
+
+	locker, ok := l.lockerMap[key]
+	if !ok {
+		return
+	}
+
+	waitCount := locker.unlock(ro)
+	if waitCount < 1 {
+		l.syncPool.Put(locker)
```
lowzj: It may put a same locker back into the pool.

Starnop: Update. PTAL.

lowzj:
LGTM

PR #569, opened 2019-05-30
bugfix: calculate realHTTPFileLength which is different realFileLength

Starnop:

Now the writer will calculate the file Length dynamically which includes the piece header and piece tailer. However, we will use httpFileLength which is the length of the file contents to check whether file length matches the expected.

None.

None.

### DIFF
--- a/supernode/daemon/mgr/cdn/manager.go
+++ b/supernode/daemon/mgr/cdn/manager.go
@@ -3,6 +3,7 @@ package cdn
 import (
 	"context"
 	"crypto/md5"
+	"path"
 	"strconv"
 
 	"github.com/dragonflyoss/Dragonfly/apis/types"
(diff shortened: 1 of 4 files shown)


supernode/daemon/mgr/cdn/manager.go
```diff
@@ -103,18 +104,16 @@ func (cm *Manager) TriggerCDN(ctx context.Context, task *types.TaskInfo) (*types
...
 		CdnStatus:  types.TaskInfoCdnStatusSUCCESS,
-		FileLength: realFileLength,
+		FileLength: downloadMetadata.realFileLength,
 		RealMd5:    realMD5,
 	}, nil
 }
 
 // GetHTTPPath returns the http download path of taskID.
+// The returned path joined the DownloadRaw.Bucket and DownloadRaw.Key.
 func (cm *Manager) GetHTTPPath(ctx context.Context, taskID string) (string, error) {
```
Starnop: @lowzj I have updated the implementation of GetHTTPPath. Dfget will get resources from supernode instead of get them form storage directly. So the supernode should do these things:

return a path that can get info to download resources from storage
listen the download port and accept the download requests
parse the path and get resources from storage.

So, I think it's reasonable to do this. WDYT?

supernode/daemon/mgr/cdn/manager.go
```diff
@@ -103,18 +104,16 @@ func (cm *Manager) TriggerCDN(ctx context.Context, task *types.TaskInfo) (*types
...
-	info, err := cm.cacheStore.Stat(ctx, getDownloadRawFunc(taskID))
-	if err != nil {
-		return "", err
-	}
-	return info.Path, nil
+	raw := getDownloadRawFunc(taskID)
+	return path.Join("/", raw.Bucket, raw.Key), nil
```
lowzj: use `filepath.Separator` instead `/`

Starnop: In fact, it's a http path not file path.

PR #579, opened 2019-06-03
feature: add integration test for golang supernode

Starnop:


.circleci/config.yml
```diff
@@ -78,7 +81,7 @@ jobs:
           name: rm coverage.txt
           command: rm coverage.txt
 
-  api-integration-test:
+  api-integration-test-java:
```
lowzj: this job should be added into workflows

Starnop:
> 2019-06-03 16:34:30.325 INFO sign:3320-1559579670.318 : do register to 127.0.0.1, res:null error:500:{"message":"{\"Code\":501,\"Msg\":\"validation failure list:\nhostName in body must be of type hostname: \"4536bc5d05e2\": {\"Code\":2,\"Msg\":\"invalid value\"}\"}"}

The CI failed because swagger will validate hostname with regex https://github.com/go-openapi/strfmt/blob/master/default.go#L60.  And the hostname of container hosts started by circleci is a string like `4536bc5d05e2`.

So we need to do some change for the api of supernode:
+ define a self-defined regex
+ or update the type of hostname to string

NOTE: I have not found a way to change the hostname of circleci.

@lowzj Do you have any ideas?

lowzj:
@Starnop I prefer to change the type to `string`.

Starnop:
> @Starnop I prefer to change the type to `string`.

Has been updated. PTAL. THX.

PR #581, opened 2019-06-04
refactor: remove the global package in dfdaemon

inoc603:

Remove the global package in dfdaemon to make testing (which I haven't done yet) easier.

This PR is larger than I would like, since the global package is referenced in so many places.

Notable changes:
- remove dfdaemon/global and cmd/dfdaemon/app/options
- remove dfdaemon/initializer and move some of its functions to cmd/dfdaemon/app
- remove cli params "rule" and "trust-hosts"
- use [github.com/spf13/viper](https://github.com/spf13/viper) for config management and make most of the cli parameters configurable from config file

Everything should work as before. Additionally these fields from the cli parameters are now configurable through config file, thanks to [github.com/spf13/viper](https://github.com/spf13/viper)

```yaml
port: 65002
hostIp: 127.0.0.1
certpem: ""
keypem: ""
node: []
localrepo: /root/.small-dragonfly/dfdaemon/data
dfpath: /root/src/github.com/inoc603/Dragonfly/bin/linux_amd64/dfget
ratelimit: 50M
urlfilter: Signature&Expires&OSSAccessKeyId
callsystem: com_ops_dragonfly
notbs: true
verbose: true
maxprocs: 4
```


dfdaemon/config/config.go
```diff
@@ -304,3 +382,26 @@ func NewProxy(regx string, useHTTPS bool, direct bool) (*Proxy, error) {
 func (r *Proxy) Match(url string) bool {
 	return r.Regx != nil && r.Regx.MatchString(url)
 }
+
+// Error is an error with exit code
+type Error struct {
```
Starnop: Use common.errors?

cmd/dfdaemon/app/root.go
```diff
@@ -54,25 +81,40 @@ var rootCmd = &cobra.Command{
...
-func initOption(opt *options.Options) {
-	if opt.DfPath == "" {
-		if path, err := exec.LookPath(os.Args[0]); err == nil {
-			if absPath, err := filepath.Abs(path); err == nil {
-				// assume the dfget binary is at the same directory as this daemon.
-				opt.DfPath = filepath.Dir(absPath) + "/dfget"
-			}
-		}
-	}
+	rootCmd.Flags().BoolP("version", "v", false, "version")
```
Starnop: Use a variable to point to  rootCmd.Flags()?

cmd/dfdaemon/app/root.go
```diff
@@ -41,11 +41,38 @@ var rootCmd = &cobra.Command{
...
+		viper.SetConfigFile(viper.GetString("config"))
+		viper.SetConfigType("yaml")
+		if err := viper.ReadInConfig(); err != nil {
+			return err
+		}
+
+		var cfg config.Properties
 
-		s, err := dfdaemon.NewFromConfig(*g.Properties, *opt)
+		exitOnError(
```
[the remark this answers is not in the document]
inoc603: Yeah, error handling in the cmd was kind of a mess. It was a mixture of `os.Exit` and error values, and I got a little confused too. I'll try to clean that up.

dfdaemon/downloader/dfget/dfget.go
```diff
@@ -80,50 +59,45 @@ func (dfGetter *DFGetter) Download(url string, header map[string][]string, name
...
-			if value != nil && len(value) > 0 {
-				for _, oneV := range value {
-					args = append(append(args, "--header"), fmt.Sprintf("%s:%s", key, oneV))
-				}
-			} else {
-				args = append(append(args, "--header"), fmt.Sprintf("%s:%s", key, ""))
-			}
-
+	add := func(key, value string) {
+		if v := strings.TrimSpace(value); v != "" {
```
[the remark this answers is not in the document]
inoc603: Reusing functions is good, and the changed code would look like this: 

```go
        if !common.IsEmptyStr(value) {
                args = append(args, key, strings.TrimSpace(value))
        }
```

I would argue that the current implementation looks a little cleaner. And since we use the result of `strings.TrimSpace` afterwards, not only for checking if `value` is empty, it's actually fine.

Also I think we might not need to trim spaces here in the first place, because the values all come from cli parameters or config file, where extra spaces would be handled during parsing. What is the original purpose of trimming spaces here? Maybe this would be enough

```go
        if value != "" {
                args = append(args, key, value)
        }
```

dfdaemon/config/config.go
inoc603: Sure, didn't notice that we have it in the common package

cmd/dfdaemon/app/root.go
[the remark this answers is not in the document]
inoc603: Now all sub functions return error, and the exit code is handled in `Execute()`, please check. @Starnop

cmd/dfdaemon/app/root.go
inoc603: Using a variable surely make the code shorter. Fixed

cmd/dfdaemon/app/root.go
```diff
@@ -41,45 +43,107 @@ var rootCmd = &cobra.Command{
 	Long:              "The dfdaemon is a proxy between container engine and registry used for pulling images.",
 	DisableAutoGenTag: true, // disable displaying auto generation tag in cli docs
 	RunE: func(cmd *cobra.Command, args []string) error {
-		initOption(opt)
+		if viper.GetBool("version") {
```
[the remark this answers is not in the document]
inoc603: I'm all for clean code, but I don't think putting these code into one function necessarily makes it clearer. (I made some more changes, and added some tests for `cmd/dfdaemon/app`. It should be better than the last time you reviewed.)

The original `initOption(opt)` is not good mainly because it has unclear side effects and does much more than its name suggests. It didn't make the code clearer or easier to test.

My current implementation has 3 crucial part:
1. `getConfigFromViper`
2. `initDfdaemon`: set up logging, local repo, GOMAXPROC etc. (maybe we can choose a better function name)
3. `dfdaemon.NewFromConfig` and `Start`

The parts themselves are self-contained and testable.

```go
viper.SetConfigFile(viper.GetString("config"))
viper.SetConfigType("yaml")
if err := viper.ReadInConfig(); err != nil {
	return errors.Wrap(err, "read config file")
}
```

These config file related actions are only used in rootCmd once. Wrapping them into one function saves 2 lines here but is hardly any clearer.

The whole `RunE` function is short, around 30 lines, you can read it without scrolling. I would say it's reasonably clear.

P.S. I do think the logging and local repo initialization needs more work. I'd like to work on it in another PR after this.

cmd/dfdaemon/app/root.go
Starnop: The function of `rootCmd.Flags()` will return as a `*flag.FlagSet`. So it's ok to use `rf`, but maybe `flagSet` is better(that's what we usually use). WDYT?

Starnop:
Dfdaemon will get errors when the `/etc/dragonfly/dfdaemon.yml` is not exist. Should we add use the default config instead of return errors directly?

cmd/dfdaemon/app/root.go
inoc603: I think the point of using a variable here is to make the code shorter. The scope of the variable is very small which fits into one screen, and the usage is simple and consistent, so `rf` should be clear enough.

Naming a `*flag.FlagSet` instance `flagSet` is not very meaningful since there's no additional information in the name, like `var integer int`. `rootFlags`, `rootFlagSet` and `rootCmdFlagSet` may be better, but are not as short.

inoc603:
> Dfdaemon will get errors when the `/etc/dragonfly/dfdaemon.yml` is not exist. Should we add use the default config instead of return errors directly?

I noticed that as well. I think you're right, we should ignore the error that config file is not found, but only for the default config file `/etc/dragonfly/dfdaemon.yml`. If user specified a config file using `--config`, we should return the error.

inoc603:
@Starnop I've made some changes to ignore the not found error for default config file, and added tests  for it. Please review.

cmd/dfdaemon/app/root.go
```diff
@@ -41,45 +43,139 @@ var rootCmd = &cobra.Command{
...
+			fmt.Println("dfdaemon version:", version.DFDaemonVersion)
+			return nil
+		}
+
+		if err := readConfigFile(viper.GetViper(), cmd); err != nil {
+			return errors.Wrap(err, "read config file")
+		}
 
-		s, err := dfdaemon.NewFromConfig(*g.Properties, *opt)
+		cfg, err := getConfigFromViper(viper.GetViper())
```
Starnop: It will cause to panic here. In fact, it's a common mistake for Golang Devs. FYI. http://devs.cloudimmunity.com/gotchas-and-common-mistakes-in-go-golang/index.html#nil_in_nil_in_vals

And you can use `!util.IsNil(err)` to avoid this problem.

cmd/dfdaemon/app/root.go
```diff
@@ -41,45 +43,139 @@ var rootCmd = &cobra.Command{
...
 		if err != nil {
```
Starnop: It will cause to panic here. In fact, it's a common mistake for Golang Devs. FYI. http://devs.cloudimmunity.com/gotchas-and-common-mistakes-in-go-golang/index.html#nil_in_nil_in_vals

And you can use `!util.IsNil(err)` to avoid this problem.

inoc603: Oh I think it got mixed up some time during refactoring. But I think this should be fixed by changing `func (p *Properties) Validate() *dferr.Error {` to `func (p *Properties) Validate() error {`

inoc603: @Starnop I've changed `func (p *Properties) Validate() *dferr.DfError` to `func (p *Properties) Validate() error`. The panic should be fixed for now. I think we can discus the change I propose for `common/errors` package in another issue.

cmd/dfdaemon/app/root.go
```diff
@@ -41,45 +43,139 @@ var rootCmd = &cobra.Command{
...
+		if err != nil {
+			return errors.Wrap(err, "create dfdaeon from config")
+		}
 		return s.Start()
 	},
 }
 
 func init() {
-	opt.AddFlags(rootCmd.Flags())
+	self, err := filepath.Abs(os.Args[0])
```
Starnop: exec.LookPath(os.Args[0])?

dfdaemon/handler/env_handler.go
```diff
@@ -17,25 +17,15 @@
...
-	env := make(map[string]interface{})
-
-	env["dfPattern"] = global.CopyDfPattern()
-
-	env["home"] = global.HomeDir
-
-	env["param"] = global.CommandLine
-
-	w.Write([]byte(fmt.Sprintf("%+v", env)))
+	json.NewEncoder(w).Encode(viper.AllSettings())
```
Starnop: In fact, you will get an error `unsupported type: map[interface {}]interface {}` here. Please check it.

cmd/dfdaemon/app/init.go
```diff
@@ -0,0 +1,161 @@
...
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+// initDfdaemon sets up running environment for dfdaemon according to the given config
+func initDfdaemon(cfg config.Properties) error {
+	if cfg.Verbose {
+		logrus.Infoln("use verbose logging")
+		logrus.SetLevel(logrus.DebugLevel)
```
Starnop: In fact, no debug logs will be printed when I use `dfdaemon --verbose=true`. Please check it. THX>

dfdaemon/config/config.go
```diff
@@ -88,6 +80,82 @@ type Properties struct {
...
 	HijackHTTPS *HijackConfig `yaml:"hijack_https"`
+
+	// https options
+	Port    uint   `yaml:"port"`
+	HostIP  string `yaml:"hostIp"`
+	CertPem string `yaml:"certpem"`
+	KeyPem  string `yaml:"keypem"`
+
+	// dfget config
+	SuperNodes []string `yaml:"supernodes"`
```
Starnop: Maybe it's better to use "node" here which is the same as the cmd flag `rf.StringSlice("node", nil, "specify the addresses(IP:port) of supernodes that will be passed to dfget.")
`. WDYT?

cmd/dfdaemon/app/root.go
inoc603: Oh, I forgot the case where we call dfdaemon with only `dfdaemon` which searches through `PATH`. `exec.LookPath` should be used indeed.

cmd/dfdaemon/app/init.go
inoc603: I suppose this is caused by calling `dflog.InitLog(false, ...)` in `initLogger` after this. I'll fix that.

dfdaemon/config/config.go
inoc603: This is surely a bug, and the value of node cli flag will not be used. But I think `supernodes` is clearer for config file. Adding an alias in viper from `supernodes` to `node` should fix that.

dfdaemon/handler/env_handler.go
inoc603: My bad, didn't actually test these apis. Should be a simple fix.

cmd/dfdaemon/app/root.go
inoc603: Fixed

dfdaemon/handler/env_handler.go
inoc603: I added a simple recursive function to change all `map[interface{}]interface{}` to `map[string]interface{}`, should be fixed now.

cmd/dfdaemon/app/init.go
inoc603: Fixed

dfdaemon/config/config.go
inoc603: I added the alias and it now works. My opinion is to keep the `supernodes` field for config file. It's more precise and it's plural, which is more appropriate for a slice.

Starnop:
LGTM. And could you please update related docs ? THX. @inoc603

inoc603:
> LGTM. And could you please update related docs ? THX. @inoc603

Sure. Would you please open a new issue for the documentation? The changes on https and proxies can be updated together.

Starnop:
> > LGTM. And could you please update related docs ? THX. @inoc603
> 
> Sure. Would you please open a new issue for the documentation? The changes on https and proxies can be updated together. FAQ, quick_start, and user_guide all need some update, I think it's going to be quite some work.
> 
> @Starnop @lowzj Can we merge this first? So I can work on tests and docs for dfdaemon

I will submit an issue ASAP. 😄 

And as for this PR, I hope that @lowzj will have a review again. After a comment `LGTM` from him ,and then I will merge it. I also hope that this PR can be merged ASAP.

cmd/dfdaemon/app/init.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// initLogger initialize the global logrus logger
+func initLogger(cfg config.Properties) error {
+	current, err := user.Current()
+	if err != nil {
+		return errors.Wrap(err, "get current user")
+	}
+	logFilePath := filepath.Join(current.HomeDir, ".small-dragonfly/logs/dfdaemon.log")
+	logrus.Infof("use log file %s", logFilePath)
```
lowzj: IMOP, it's better to output it on debug level.

cmd/dfdaemon/app/root.go
```diff
@@ -41,45 +44,142 @@ var rootCmd = &cobra.Command{
...
+			yaml.NewEncoder(os.Stdout).Encode(cfg)
+		}
+
+		if err := initDfdaemon(*cfg); err != nil {
+			return errors.Wrap(err, "init dfdaemon")
 		}
 
+		s, err := dfdaemon.NewFromConfig(*cfg)
+		if err != nil {
+			return errors.Wrap(err, "create dfdaeon from config")
```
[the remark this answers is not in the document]
inoc603: fixed

cmd/dfdaemon/app/root.go
```diff
@@ -41,45 +44,142 @@ var rootCmd = &cobra.Command{
...
+}
+
+// getConfigFromViper returns dfdaemon config from the given viper instance
+func getConfigFromViper(v *viper.Viper) (*config.Properties, error) {
+	var cfg config.Properties
+	if err := v.Unmarshal(&cfg, func(dc *mapstructure.DecoderConfig) {
+		dc.TagName = "yaml"
+		dc.DecodeHook = decodeWithYAML(
+			reflect.TypeOf(config.Regexp{}),
+			reflect.TypeOf(config.URL{}),
```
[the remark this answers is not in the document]
inoc603: Fixed and added tests

cmd/dfdaemon/app/init.go
inoc603: Sure

cmd/dfdaemon/app/root.go
```diff
@@ -41,45 +44,142 @@ var rootCmd = &cobra.Command{
...
 
-		s, err := dfdaemon.NewFromConfig(*g.Properties, *opt)
+		cfg, err := getConfigFromViper(viper.GetViper())
 		if err != nil {
-			logrus.Fatal(err)
-			return err
+			return errors.Wrap(err, "get config from viper")
+		}
+
+		if viper.GetBool("verbose") {
```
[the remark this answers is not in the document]
inoc603: I logged each option when using them, with more readable information. But logging the whole configuration doesn't hurt anyway.

inoc603:
Rebased

PR #553, opened 2019-05-27
dfget --node flag can specify the port of supernodes.

xq2248:
<!-- 
### Ⅰ. Describe what this PR did
The --node flag of dfget only can specify the ip of supernodes, but sometimes wo need to change the port 8002 which supernodes listen on. We add ip:port syntax to the --node flag, but the old ip syntax can also be supported.

### Ⅱ. Does this pull request fix one issue?
The following commands have been tested:
dfget -u xxx -n ip1,ip2
dfget -u xxx -n ip1:port1,ip2
dfget -u xxx -n ip1:port1,ip2:port
They works.

### Ⅲ. Why don't you add test cases (unit test/integration test)? (你真的觉得不需要加测试吗？)
Yes, I have add unit test function.

### Ⅳ. Describe how to verify it
Modify the nginx listen port 8002 to 8005 at supernode, restart nginx.
Use dfget -n IP or dfget -n IP:8005 to specify the port of supernode.

PR #542, opened 2019-05-23
dfget --node flag can specify the port of supernodes

xq2248:
### Ⅰ. Describe what this PR did
The --node flag of dfget only can specify the ip of supernodes, but sometimes wo need to change the port 8002 which supernodes listen on.  We add ip:port syntax to the --node flag, but the old ip syntax can also be supported.

### Ⅱ. Does this pull request fix one issue?
fixes #473 #543

### Ⅴ. Special notes for reviews

The following commands have been tested:
dfget -u xxx -n ip1,ip2 
dfget -u xxx -n ip1:port1,ip2 
dfget -u xxx -n ip1:port1,ip2:port
They works.

Issue #473, opened 2019-04-03
[question] P2P and nginx ports


yinxianliang:
---
name: Question about Dragonfly
about: Ask whatever you want to know or confusion about Dragonfly
---
## Question

1. When P2P shares partitions, it is desirable to specify ports. There are walls between hosts in the production environment, and random ports are inconvenient.

// 1.p2p共享分块时，希望能够指定端口，生产环境主机之间有墙，随机端口不方便 

2. I don't understand that port 8001 of nginx in supernode points to repo directory and 8002 points to Tomcat service (proxy_pass http://127.0.0.1:8080;). The client's registration port is 8001, and the download port is 8002, which is not very clear (or is my setup incorrect?)

// 2.有一点我不明白，supernode中nginx的8001端口指向repo目录，8002指向了tomcat服务（proxy_pass http://127.0.0.1:8080;）。而客户端的注册端口是8001，下载端口是8002，不是很理解（还是我设置有误？） 

3. It is hoped that 8001 and 8002 ports can also be changed, and port occupancy may occur.

// 3.希望8001和8002端口也能改变，可能出现端口占用情况。

***!!!!WE STRONGLY ENCOURAGE YOU TO DESCRIBE YOUR ISSUE IN ENGLISH!!!!***

pouchrobot:
Thanks for your contribution. 🍻 @yinxianliang 
While we thought **ISSUE TITLE** could be more specific, longer than 20 chars.
Please edit issue title instead of opening a new one.
More details, please refer to https://github.com/dragonflyoss/Dragonfly/blob/master/CONTRIBUTING.md

pouchrobot:
Thanks for your contribution. 🍻 @yinxianliang 
While we thought **ISSUE TITLE** could be more specific, longer than 20 chars.
Please edit issue title instead of opening a new one.
More details, please refer to https://github.com/dragonflyoss/Dragonfly/blob/master/CONTRIBUTING.md

pouchrobot:
Thanks for your contribution. 🍻 @yinxianliang 
While we thought **ISSUE TITLE** could be more specific, longer than 20 chars.
Please edit issue title instead of opening a new one.
More details, please refer to https://github.com/dragonflyoss/Dragonfly/blob/master/CONTRIBUTING.md

Starnop:
> When P2P shares partitions, it is desirable to specify ports. There are walls between hosts in the production environment, and random ports are inconvenient.

[...]

PR #586, opened 2019-06-05
feature: parse the port of supernode from --node flag of dfget

### LINKED ISSUE #543: --node flag of dfget can not specify port of supernodes
### Ⅰ. Issue Description
The description of dfget command show that,"specify the addresses(IP:port) of supnernodes", but we can only specify ip of supernodes. That is to say, the port of supernodes must be 8002 to all the peers. We wish to fix the problem.

### Ⅱ. Describe what happened
The download will fail when supernodes listen on  ports which not equals to 8002.

The --node flag of dfget can specify ip:port of the supernodes.

dfget -u http://www.xxx.zip -n 10.47.231.22:8005
The download will fail, and  errors occurs in the dfclient.log as follows:"2019-05-23 19:03:41.639 INFO sign:15977-1558609421.526 : do register to 10.47.231.22:8005, res:null error:address 10.47.231
.22:8005:8002: too many colons in address"
 
### Ⅵ. Environment:
- dragonfly version: latest.
- OS (e.g. from /etc/os-release): Centos7
- Kernel (e.g. `uname -a`): 3.11

Starnop:

fixes #543 
fixes part of #473

Added

This PR do the same job with #542 #553. And thanks for the job by @xq2248

>merge into branch master and 0.3.x

### DIFF
--- a/dfget/config/constants.go
+++ b/dfget/config/constants.go
@@ -113,6 +113,8 @@ const (
 
 	DataExpireTime  = 3 * time.Minute
 	ServerAliveTime = 5 * time.Minute
+
+	DefaultSupernodePort = 8002
 )
 
 /* errors code */
--- a/cmd/dfget/app/root.go
+++ b/cmd/dfget/app/root.go
@@ -82,6 +82,10 @@ func runDfget() error {
 		util.Printer.Println(err.Error())
 		return err
 	}
+	if err := handleNodes(); err != nil {
+		util.Printer.Println(err.Error())
+		return err
+	}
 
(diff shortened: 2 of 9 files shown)


dfget/config/constants.go
```diff
@@ -113,6 +113,8 @@ const (
 
 	DataExpireTime  = 3 * time.Minute
 	ServerAliveTime = 5 * time.Minute
+
+	DefaultNodePort = 8002
```
lowzj: `DefaultSupernodePort`

PR #589, opened 2019-06-06
feature: update task register header to slice not map

Starnop:


Starnop:
@lowzj /request a review. 😄

PR #594, opened 2019-06-12
[WIP] docs: add proposal for decentralized distribution

zhouhaibing089:
Adds a simple doc to describe the idea to use decentralized way
for file distribution.


inoc603:
I agree that we dfdaemon should call dfget by using a library rather than `exec.Command`. Running a external command is not a very robust solution and it makes configuration more complicated.

docs/proposals/decentralized.md
```diff
@@ -0,0 +1,421 @@
...
+	s.agent.UserEvent(EventLayer, data, false)
+}
+```
+
+This event basically wants its peers to realize that there are new blocks available
+for this specific layer.
+
+#### Layer Removed
+
+This happens when peers discard this layer because of reasons liek garbage
```
SataQiu:
```suggestion
This happens when peers discard this layer because of reasons like garbage
```

docs/proposals/decentralized.md
```diff
@@ -0,0 +1,421 @@
...
+	Source  bool
+	// Blocks is the number blocks available.
+	Blocks  int64
+}
+```
+
+## Request Flow
+
+Let's say we have the states tracked in local memory, and now we have a request
+coming, the following is the persuade code on how it is handled.
```
jyizheng: persuade code -> pseudo code?

docs/proposals/decentralized.md
```diff
@@ -0,0 +1,421 @@
...
+## dfdaemon
+
+Today, dfdaemon is a long-running proxy service for docker and it calls dfget
+internally to roundtrip the http requests. This is not necessary given that dfget
+is also written in Go. A direct library call seems to be more straightforward than
+running an `exec.Command`:
+
+## dfget
+
+Once the existing dfget functionality is moved to dfdaemon, we cam make dfget
```
jyizheng: cam -> can?

zhouhaibing089: Updated, thanks.

docs/proposals/decentralized.md
zhouhaibing089: Yes, updated.

docs/proposals/decentralized.md
zhouhaibing089: Updated.

PR #496, opened 2019-04-17
move to go modules

zhouhaibing089:
Go modules has been widly adopted at this moment. In order to work
with the ecosystem well, we can move to go module as well.

### Ⅰ. Describe what this PR did

using go module.

### Ⅱ. Does this pull request fix one issue?

None

PR #597, opened 2019-06-12
chore: use mod to manage packages

wi-cuckoo:

Add two files: go.mod, go.sum. To support using go module feature

No

Nothing commit about new features

* ensure go version to support module
* run `make build-client` to get normal output

nil


Starnop:
Thanks for your contribution. Unfortunately, it's do the same job with #496. And we have not to be ready to use `go mod` until the next version is released. And that will come soon.

PR #606, opened 2019-06-14
bugfix: return http file length as the file length of register resp

Starnop:


Starnop:
@lowzj /request a review. THX>

PR #607, opened 2019-06-14
test: add unit tests for dfdaemon/config

inoc603:

Add unit tests for dfdaemon/config

go test -v ./dfdaemon/config -cover

### DIFF
--- a/dfdaemon/config/config.go
+++ b/dfdaemon/config/config.go
@@ -20,18 +20,19 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
-	"io/ioutil"
 	"net/url"
 	"os"
 	"path/filepath"
 	"regexp"
 
 	dferr "github.com/dragonflyoss/Dragonfly/common/errors"
-	"github.com/dragonflyoss/Dragonfly/common/util"
 	"github.com/dragonflyoss/Dragonfly/dfdaemon/constant"
 
 	"github.com/pkg/errors"
+	"github.com/spf13/afero"
 )
+
+var fs = afero.NewOsFs()
 
 // -----------------------------------------------------------------------------
 // Properties
@@ -347,7 +348,7 @@ func certPoolFromFiles(files ...string) (*x509.CertPool, error) {
 
 	roots := x509.NewCertPool()
 	for _, f := range files {
-		cert, err := ioutil.ReadFile(f)
+		cert, err := afero.ReadFile(fs, f)
 		if err != nil {
 			return nil, errors.Wrapf(err, "read cert file %s", f)
 		}
@@ -356,15 +357,6 @@ func certPoolFromFiles(files ...string) (*x509.CertPool, error) {
 		}
 	}
 	return roots, nil
-}
-
-// Load loads properties from config file.
-func (p *Properties) Load(path string) error {
-	if err := util.LoadYaml(path, p); err != nil {
-		return err
-	}
-
-	return nil
 }
 
 // Proxy describe a regular expression matching rule for how to proxy a request
--- a/dfdaemon/config/config_test.go
+++ b/dfdaemon/config/config_test.go
@@ -17,21 +17,23 @@
 package config
 
 import (
-	"io/ioutil"
+	"crypto/x509"
+	"encoding/json"
+	"fmt"
+	"math/rand"
 	"os"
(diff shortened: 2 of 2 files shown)


inoc603:
The unit tests are failing because I use go 1.12 on my machine and the CI uses go 1.10. I'll fix this later

dfdaemon/config/config.go
```diff
@@ -20,19 +20,20 @@ import (
...
-	"io/ioutil"
 	"net/url"
 	"os"
 	"path/filepath"
 	"regexp"
 
 	dferr "github.com/dragonflyoss/Dragonfly/common/errors"
-	"github.com/dragonflyoss/Dragonfly/common/util"
 	"github.com/dragonflyoss/Dragonfly/dfdaemon/constant"
+	"github.com/spf13/afero"
```
fengzixu: Please construct the structure of your importing packages according to the rule:

// Golang internal package

// Dragonfly internal package

// Third-party package

dfdaemon/config/config.go
```diff
@@ -347,7 +348,7 @@ func certPoolFromFiles(files ...string) (*x509.CertPool, error) {
 
 	roots := x509.NewCertPool()
 	for _, f := range files {
-		cert, err := ioutil.ReadFile(f)
+		cert, err := afero.ReadFile(fs, f)
```
fengzixu: Could you explain why do you use the function in 'spf13/afero' package rather than golang internal package? Is there any advantage?

dfdaemon/config/config_test.go
```diff
@@ -17,22 +17,23 @@
...
+	"fmt"
+	"math/rand"
 	"os"
-	"path/filepath"
+	"strings"
 	"testing"
+	"time"
 
-	"github.com/go-check/check"
+	dferr "github.com/dragonflyoss/Dragonfly/common/errors"
```
fengzixu: same problem.

dfdaemon/config/config.go
inoc603: [spf13/afero](https://github.com/spf13/afero) abstract the file system, so we can mock the file system with `afero.NewMemFs` during unit tests.

 It removes the host file system as test dependency, and it's easier to setup and teardown. Mocking cases where something does not exists is much easier.

dfdaemon/config/config_test.go
inoc603: fixed

dfdaemon/config/config.go
inoc603: fixed

inoc603:
@lowzj @Starnop The CI is complaining about a [dead link](https://circleci.com/gh/dragonflyoss/Dragonfly/6629?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) in README.md, which I didn't change. I met this problem earlier too. Triggering the CI again can fix it. Would you look into this?

```
FILE: ./README.md
[✖] https://codecov.io/gh/dragonflyoss/Dragonfly/branch/master/graph/badge.svg

ERROR: dead links found!
```

dfdaemon/config/config.go
inoc603: @fengzixu  Using the standard library is fine and unit tests can be done with temp dir. I'm ok to revert this and write tests using the the actual file system, since other parts of the code is using the fs directly. Maybe we can ask the opinion from other members? @lowzj @Starnop

lowzj: I prefer to use one testing framework in Dragonfly. Maybe `testify` is better than `go-check` used now. But recently we're going to release `0.4.0` and there's no plan to change a testing framework in this version. 
We can change the testing framework in next version after discussion. 
WDYT @inoc603 @fengzixu @Starnop

inoc603: @lowzj the `client` package uses `testify` too, and I thought we use both testing framework in Dragonfly, so I chose one from them. If we want to stick to `go-check`, we should change tests in `client` package too. The 2 frameworks are very similar in many ways, I think it's fine they co-exist in one project.

Personally I would prefer `testify`. It's more popular, the development is more active, and it's more flexible that it does not force you to organize your tests in a particular way. I use `suite.Suite` because it's similar to the way `go-check` organize tests, but you can use `testify` in standard go tests as an assertion library only.

Issue #370, opened 2019-02-02
[feature request] construct configuration data structure more decoupled in components


allencloud:
## Why you need it?

Currently, the configuration of dragonfly's components seems to be a little bit unreadable. The good data structure's purity will make contributors and developers understand what dragonfly does very quickly. Otherwise, it will take them lots of time to understand the clear path of workflow in component. Thus the goal of data structure's construction should be paid more attention on.

This issue will take the configuration as an example.

## How it could be?

**Current Situation**: 

For the configuration part, let us take the dfget as an example;

* all the whole binary will share a same configuration in code, 
* and in the configuration, it contains the command flags, details parameters in configuration and RV(runtime variables) which is quite hard to understand. 
* And in the dfget, several components share the global configuration and runtime variables which breaks the scalability due to that they are coupled by shared configurations

**New possible way**

* Split the configuration into dependent parts, command line flags is flags, configuration is configuration and maybe no need to add runtime variables which contains less detailed logistics.
* keep the configuration straightforward and layered: since dfget has several component, like task downloader, uploader and others, then we can keep the top configuration layered to be much clearer, for example:

```
type Dfget Struct {
      taskCfg TaskDownloaderConfig
      uploaderCfg UploaderConfig
      otherField []string
}

type TaskDownloaderConfig struct{
      A int
      B string
}

type UploaderConfig struct{
      C int
      D string
}
``` 
* avoid to share the top global configuration among component if possible

## Other related information

none

call for reviews. @lowzj @Starnop @fengzixu

Issue #363, opened 2019-01-30
[feature request] unify the same field type definition in different places


fengzixu:
---
name: Feature Request
about: Suggest your excellent idea for this project

---

## Why you need it?

For reaching a good result about refactoring the dfget, we should unify the type definition in different places. Besides, the operation of unifying type can reduce redundant logic of type conversion.

## How it could be?

For this feature, I think that I should know carefully code of dfget and pick up logic about using some type definition incorrectly. Finally, I will revise it.

## Other related information

This issue is the sub-question of #347

Starnop:
What is the status of it ? Could you please keep up with the latest developments with us?

fengzixu:
@Starnop Have finished. Submit it within this week.

Starnop:
@fengzixu Are you still willing to complete this job? If so, could you please do it as soon as possible? If not, please let me know. THX> 😄

fengzixu:
@Starnop Sure. Let me submit a PR within this week.

PR #612, opened 2019-06-17
refactor: unify the same field type definition in different places

fengzixu:
Signed-off-by: fengzixu <hnustphoenix@gmail.com>

1. reduce using of extra fields which were included in anthor definition

#363 

I run all unit-test in the dfget package.

For 「use int32 or int64 instead of int」, I think that using 'int' is ok for the dfget package. We don't need to revise all variables whose type is int into a wider type.

For 「unify the same field type definition in different places」, I revise the code which was written in the package called downloader. Because I consider that we should not add extra variables that are already defined elsewhere.


fengzixu:
@Starnop Request review.

Starnop:
Thanks for your contribution. However, maybe it's not a good solution. And in fact, I am trying to fix the issue #370 which is just the opposite of what you did. 😢 

And for issue issue #363, I think what should we do is just as title described. For example, we think that `int32` is enough for `pieceSize`. However, it's defined with `int64` [here](https://github.com/dragonflyoss/Dragonfly/blob/master/dfget/core/uploader/peer_server.go#L104). And that is exactly what we should make a change. 

And the other similar fields but not completely, just FYI:
+ pieceNum (expected int64)
+ port (int32/int both OK but should be unified)
+ ......

fengzixu:
> * unified

Ok. Let me fix the problem about integer type. I will give up the modification of configuration type.

Starnop:
Any updates?

fengzixu:
@Starnop Sorry. I am too busy at the current time. I will finish it asap.

Starnop:
Don't worry about that. I just want to get the latest progress.  Whatever, thanks for your work. 😄

PR #613, opened 2019-06-17
bugfix: use string as the type of lastModified

Starnop:


Starnop:
Not a good solution. So close it now.

PR #614, opened 2019-06-17
feature: add piece length for task md5 file

Starnop:


Starnop:
> reading and writing md5 file should be in the order of pieceNum.

Yes. It will be sorted [here](https://github.com/dragonflyoss/Dragonfly/blob/master/supernode/daemon/mgr/cdn/piece_md5_mgr.go#L68)

PR #619, opened 2019-06-19
fix: send http request with timeout

lowzj:

* change connection timeout from `30 seconds` to `3 seconds`
* use `GET` method instead of `HEAD` method to get `content-length`, check expiration and whether support range request. It's because that some servers may not support `HEAD` method.
* set the whole operation's timeout as `4 seconds`(including connection timeout)


Starnop:
LGTM.

Issue #634, opened 2019-06-27
Proposal: upgrade to Golang 1.12.6 and use go mod


yeya24:
## Why you need it?

Go module is widely used in many Go projects. It is faster than govendor so it is better to switch to it in dragonfly now.

Currently the golang version is 1.10.4, in order to use Go module, we have to upgrade to Go 1.12 first.

## How it could be?

Implement steps:
- upgrade to go 1.12.6
- replace govendor with go mod

yeya24:
I think we can close this now. :smile:

PR #647, opened 2019-07-02
chore: replace govendor with go mod and update docs

yeya24:
Replace govendor with go mod since we have already updated to go 1.12.6.

fix second part of #634


allencloud:
It is a big change to the project.
I am wondering if there is any possibility to check the go mod file's correctness via Makefile target?
If we can do that, we could validate the dependency's correctness.
Just my little opinion. Wish your feedback. @yeya24

yeya24:
> It is a big change to the project.
> I am wondering if there is any possibility to check the go mod file's correctness via Makefile target?
> If we can do that, we could validate the dependency's correctness.
> Just my little opinion. Wish your feedback. @yeya24

@allencloud 
Yes this validation is important. And I see this Makefile in [Prometheus](https://github.com/prometheus/promu/blob/3e43dfdc9726cee3f7c21476d9add8f13bebb645/Makefile.common#L188) is like doing what we want. Is this correct?
I change title to WIP and I will try to work on this.

inoc603:
Since we're hopping on the go mod train, maybe we can consider not checking the `vendor/` directory into git repo. `go.sum` and `go.mod` can pin dependency versions. Removing `vendor/` will prevent huge PRs when changing dependencies. Also we don't have to validate whether `vendor/` and `go.sum` matches.

Of course not checking in `vendor/` means we would spend more time downloading dependencies during CI. This can be solved by either:
- Make an image with dependencies, which is automatically updated when `go.sum` and `go.mod` changes, and do compiling in it.
- Or use a proxy that supports the [module proxy protocol](https://tip.golang.org/cmd/go/#hdr-Module_proxy_protocol), some options can be found [here](https://github.com/golang/go/wiki/Modules#are-there-always-on-module-repositories-and-enterprise-proxies)

yeya24:
@allencloud  @lowzj  WDYT?

Starnop:
@lowzj @allencloud @inoc603 Now that we have updated the go version to 1.12.6, so let's make this PR as a break change for release 0.4.1. WDYT?

Starnop:
> Since we're hopping on the go mod train, maybe we can consider not checking the `vendor/` directory into git repo. `go.sum` and `go.mod` can pin dependency versions. Removing `vendor/` will prevent huge PRs when changing dependencies. Also we don't have to validate whether `vendor/` and `go.sum` matches.
> 
> Of course not checking in `vendor/` means we would spend more time downloading dependencies during CI. This can be solved by either:
> 
> * Make an image with dependencies, which is automatically updated when `go.sum` and `go.mod` changes, and do compiling in it.
> * Or use a proxy that supports the [module proxy protocol](https://tip.golang.org/cmd/go/#hdr-Module_proxy_protocol), some options can be found [here](https://github.com/golang/go/wiki/Modules#are-there-always-on-module-repositories-and-enterprise-proxies)

I agree with that. Even though `go mod` now supports the vendor, however, it will be abandoned in the future. And as @inoc603 said, we can use docker image to provide repeatable environments.

yeya24:
Currently I met some problems in fixing ci.
- gometalinter cannot support go mod without vendor. See issue https://github.com/alecthomas/gometalinter/issues/562. Currently I have to run  go mod vendor first then run gometalinter. Maybe we can update to golang-ci in the future since gometalinter has deprecated.
- It will always fail in integration-test. I am still finding the cause.

Makefile
```diff
@@ -94,13 +100,33 @@ unit-test: build-dirs
 	./hack/unit-test.sh
 .PHONY: unit-test
 
+integration-test:
+	@go build ./test
```
Starnop: s/ `@go build ./test`/ `@go test ./test` ? Maybe it's the cause of the failure of the integration test.

Makefile
```diff
@@ -94,13 +100,33 @@ unit-test: build-dirs
...
+go-mod-tidy:
+	@echo "Begin to tidy up go.mod and go.sum"
+	@go mod tidy
+.PHONY: go-mod-tidy
+
+# we need this because currently gometalinter can't work in go mod environment.
+go-mod-vendor:
+	@echo "Begin to vendor go mod dependency"
+	@go mod vendor
+.PHONY: go-mod-tidy
```
Starnop: go-mod-vendor？

hack/env.sh
```diff
@@ -24,3 +24,6 @@ GOOS=$(go env GOOS)
 GOARCH=$(go env GOARCH)
 export GOOS
 export GOARCH
+
+export GO111MODULE=on
+export GOPROXY=https://goproxy.io
```
Starnop: Add a newline at end of file.

Makefile
yeya24: Yes, see this issue alecthomas/gometalinter#562. Currently gometalinter cannot work in go mod env without vendor. So I have to add this during ci when checking gometalinter.

hack/env.sh
yeya24: Done.

Makefile
yeya24: Done.

Makefile
yeya24: Thanks for this help!

yeya24:
@Starnop @lowzj  Can you review this again? Pass the ci now.

Starnop:
LGTM. And @lowzj  PTAL again.

Starnop:
LGTM.

allencloud:
Great work. We are moving to `go mod`. 👍 @yeya24 @Starnop

PR #655, opened 2019-07-04
docs:optimize the quick-start doc

yunfeiyangbuaa:
docker container uses utc time and the host uses cst time,so the log's time is different from the host.so maybe it is better to make time synchronization when we start a container

### DIFF
--- a/docs/quick_start/README.md
+++ b/docs/quick_start/README.md
@@ -20,10 +20,10 @@ docker run -d --name supernode --restart=always -p 8001:8001 -p 8002:8002 \
 ## Step 2：Deploy Dragonfly Client (dfclient)
 
 ```bash
-docker run -d --name dfclient -p 65001:65001 dragonflyoss/dfclient:0.4.0 --registry https://index.docker.io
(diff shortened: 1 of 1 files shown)


lowzj:
IMOP, it's unnecessary for us to add this in quick-start document.

allencloud:
Any feedback for @lowzj 's thought? @yunfeiyangbuaa

yunfeiyangbuaa:
> Any feedback for @lowzj 's thought? @yunfeiyangbuaa
it only a optimize for user to understand  the log,but it will make the quick start more  confused  to newer ,so  it may be better if we abandon this optimize

Starnop:
> it only a optimize for user to understand the log,but it will make the quick start more confused to newer ,so it may be better if we abandon this optimize


@yunfeiyangbuaa Maybe you can add it as an item of FAQ. WDYT?

yunfeiyangbuaa:
> > it only a optimize for user to understand the log,but it will make the quick start more confused to newer ,so it may be better if we abandon this optimize
> 
> @yunfeiyangbuaa Maybe you can add it as an item of FAQ. WDYT?

I will work on it. @Starnop

Starnop:
Any updates?

PR #658, opened 2019-07-05
refactor: remove code of java-version supernode

lowzj:


allencloud:
sounds good to me. The change is reasonable, while I would like to ask @Starnop to take a look at this.
He is supposed to be free this afternoon. @lowzj

Starnop:
LGTM. This will take Dragonfly to a new stage and then let's focus on the  Dragonfly with Golang and make it better. BTY,  I think we can release a new version (0.4.1) with this big change. WDYT? @lowzj @allencloud

allencloud:
I agree with 0.4.1.

PR #663, opened 2019-07-08
bugfix: update Dockerfile of supernode

Starnop:

merge to 0.4.x

### DIFF
--- a/Dockerfile.supernode
+++ b/Dockerfile.supernode
@@ -9,14 +9,15 @@ COPY . /go/src/github.com/dragonflyoss/Dragonfly
 # write the resulting executable to the dir /opt/dragonfly/df-supernode.
 RUN make build-supernode && make install-supernode
 
-FROM nginx:1.16-alpine
+FROM dragonflyoss/nginx:apline
 
 RUN apk --no-cache add ca-certificates bash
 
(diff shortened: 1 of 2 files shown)


Dockerfile.supernode
```diff
@@ -13,10 +13,11 @@ FROM nginx:1.16-alpine
 
 RUN apk --no-cache add ca-certificates bash
 
+COPY --from=builder /go/src/github.com/dragonflyoss/Dragonfly/hack/start-supernode.sh /root/start.sh
 COPY --from=builder /go/src/github.com/dragonflyoss/Dragonfly/hack/supernode-nginx.conf /etc/nginx/nginx.conf
 COPY --from=builder /opt/dragonfly/df-supernode/supernode /opt/dragonfly/df-supernode/supernode
```
lowzj: Could we merge these commands into one command to reduce the number of image layers?

Starnop: Maybe it's not a good idea.  

+ Firstly, we copy the files from different directory to different directory, and it's difficult to merge these commands into one command.
+ Secondly, we can take advantage of cache of the static layer.

readerx:
Why switch from `nginx` to `dragonflyss/nginx`?
What's the difference between `nginx` and `dragonflyss/nginx`

PR #667, opened 2019-07-11
feature: enable uploader server ip and port to be specified

Starnop:

At present, we can specify the uploader server port with command `dfget server --port`. However, you have to run `dfget server` to start a uploader server before using `dfget` to download. If you don't want to do that, and this PR will help you.

In the meantime, you can't specify the IP address of the uploader server as a peer and `dfget` will use the IP address which can success to connect to `supernode`  automatically before this PR. In fact, the address that can be connected to a supernode is not necessarily the address that can be accessed by other peer nodes.

And now, you can use command as follows to specify the ip and port of uploader server when you want to download a file with `dfget`:
`dfget -u  https://www.taobao.com --ip 192.168.33.21 --port 55555`.

None.

None.

Don't merge it until 0.4.1 release has been published.

### DIFF
--- a/cmd/dfget/app/root.go
+++ b/cmd/dfget/app/root.go
@@ -236,7 +236,11 @@ func initFlags() {
 	flagSet.BoolVar(&cfg.Verbose, "verbose", false,
 		"be verbose")
 
-	// pass to server
+	// pass to peer server which as a uploader server
+	flagSet.StringVar(&cfg.RV.LocalIP, "ip", "",
+		"IP address that server will listen on")
+	flagSet.IntVar(&cfg.RV.PeerPort, "port", 0,
+		"port number that server will listen on")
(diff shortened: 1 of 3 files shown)


cmd/dfget/app/root.go
```diff
@@ -237,6 +237,10 @@ func initFlags() {
 		"be verbose")
 
 	// pass to server
+	flagSet.StringVar(&cfg.RV.LocalIP, "ip", "",
```
allencloud: Which server? Could you be more specific?
The command description is user-facing, and maybe some user who are not so familiar with dragonfly yet will use the command as well. We should treat such kind of user very well, to make them all clear to the parameter's usage.

Starnop: Yeah, I will update the comment to make it more clear.

PR #669, opened 2019-07-11
feature: add support for prometheus metrics

### LINKED ISSUE #197: [feature request] support prometheus in supernode to collect metrics
---
name: Feature Request
about: Suggest your excellent idea for this project
---

## Why you need it?

When we deployed a software, how it runs, what it has done and some other aspects are always important for users. We need to get all these data to have more control the software's runtime, like dragonfly's supernode.

So, I request to support [prometheus](https://github.com/prometheus/prometheus) to collect metrics in supernode of Dragonfly.

## How it could be?

Dragonfly's supernode would expose interfaces for others to collect metrics inside itself. Eventually, all metrics of supernode could be displayed in prometheus's dashboard.

yeya24:

Add basic support for prometheus metrics, both dfdaemon and supernode.
I will add more metrics later in this PR.

fix #197 

Currently I add one simple test in supernode router.

It is a WIP PR. I will add some docs for how to configure prometheus.


allencloud:
I do see the importance of this work. While I still wish to see more work to be finished. For example, the docs which guide user to experience Dragonfly and Prometheus, the integration test to see the `/metrics` works. 

I think we can do a thorough plan for this which is composed of fixed parts. And we track how many parts we have finished, while how many have not.

yeya24:
@allencloud  
Thanks for your reply. Currently this PR is in a very early stage because of some other things. I think I can finish the docs and integration test this weekend. When this PR is ready for review, I will ping you :smile:.

test/api_metrics_test.go
```diff
@@ -0,0 +1,72 @@
...
+type APIMetricsSuite struct {
+	starter *command.Starter
+}
+
+func init() {
+	check.Suite(&APIMetricsSuite{})
+}
+
+// SetUpTest does common setup in the beginning of each test.
+func (s *APIMetricsSuite) SetUpTest(c *check.C) {
```
lowzj: `SetUpTest` -> `SetUpSuite`

supernode/server/router.go
```diff
@@ -9,14 +9,16 @@ import (
 	"github.com/dragonflyoss/Dragonfly/apis/types"
 
 	"github.com/gorilla/mux"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
 // versionMatcher defines to parse version url path.
 const versionMatcher = "/v{version:[0-9.]+}"
 
+var m = newMetrics()
```
Starnop: Does `m` need to be a global variable?

yeya24: WDYT?  Become a global variable or add it to struct Server? I need to access this because of unit tests.

docs/user_guide/monitoring.md
```diff
@@ -0,0 +1,114 @@
+# Monitor Dragonfly with Prometheus
+
+Currently metrics become an important part of observability. As for monitoring Dragonfly, we recommend you to use Prometheus.
+
+In Dragonfly project, there are two long-running process: supernode and dfdaemon. Both of these two components expose its metrics via `/metrics` endpoint, so Prometheus can get metrics from these two components. We will also support dfget metrics in the future. As for current metrics, you can check out this docs.
```
allencloud: s/process/processes.
s/Both of these two components expose/Each of the components exposes/g
s/these two components/each component

These is used to describe instances which are more than two.

test/util_api.go
```diff
@@ -15,3 +19,29 @@ func CheckRespStatus(c *check.C, resp *http.Response, status int) {
...
+	)
+
+	resp, err := request.Get("/metrics")
+	c.Assert(err, check.IsNil)
+	defer resp.Body.Close()
+	data, err := ioutil.ReadAll(resp.Body)
+	c.Assert(err, check.IsNil)
+	lines := strings.Split(string(data), "\n")
+	for _, line := range lines {
+		if strings.Contains(line, key) {
```
allencloud: How about:
```
if !strings.Contains(line, key){
    continue
}
data := strings.Split(line, " ")
if len(data) < 2 {
    c.Fatal("blabla")
}
val, err = strconv.ParseFloat(data[1], 64)
c.Assert(err, check.IsNil)
return val, true
```

test/util_api.go
```diff
@@ -15,3 +19,29 @@ func CheckRespStatus(c *check.C, resp *http.Response, status int) {
...
+			val, err = strconv.ParseFloat(strings.Split(line, " ")[1], 64)
+			c.Assert(err, check.IsNil)
+			found = true
+			return val, found
+		}
+	}
+
+	return val, found
```
allencloud: directly use `return val, false` and then we have no need to declare variable `found`. WDYT?

docs/user_guide/monitoring.md
```diff
@@ -0,0 +1,114 @@
+# Monitor Dragonfly with Prometheus
```
allencloud: In these document, I think we should first tell users that which specific metrics supernode and dfdaemon support. Use a table to list the detailed metrics?

yeya24: Yes, I will add a link to another docs which contains all the metrics information.

test/util_api.go
yeya24: Agree. Much better!

test/util_api.go
yeya24: We don't need to check the length of the line. It will always  be 2.

docs/user_guide/monitoring.md
yeya24: Done

test/util_api.go
allencloud: What if someone mocks a `/metrics` endpoint and returns the data with incorrect format. The key point is that we can not guarantee what a unknown backend does and we must take every possibility into consideration and make our own code more robust. WDYT

yeya24:
I want to separate PR into small pieces. So this PR is ready for review. And after this merged, I will add more metrics later.

allencloud:
LGTM

allencloud:
Really appreciate it for your great work. @yeya24 👍

PR #670, opened 2019-07-12
feature: pass the value of dfdaemon verbose flag to dfget

Starnop:

As title described.

None.

None.

Run dfdaemon with `--verbose` and you will get more logs.


Starnop:
> We should consider providing another more flexible way to configure the `dfget` rather than passing the flags of `dfdaemon` to `dfget` in the future.
> 
> * Adding a string property as the `dfget`'s parameters.
> * Different proxies may need to pass different parameters to `dfget` when calling it, such as `urlfilter`.

Yeah. We need a specific plan, and I'll submit a proposal for that.

PR #671, opened 2019-07-12
bugfix: initProperties after trans the param

Starnop:

NOTE: The `rate-related-params` below means the  params `localLimit`, `minRate`,`totalLimit` and so on.

At present, we can configure `rate-related-params`  by cmd or config files. And we should try to trans the param value firstly which from cmd. If not, try to imagine a scene.

If you not specify the `rate-related-params` values , and then they will get the default values when init config. Immediately after that, we will try to parse the `rate-related-params` values specified by cmd. And then you will use the empty value to override the default value.

### DIFF
--- a/dfget/core/core.go
+++ b/dfget/core/core.go
@@ -262,6 +262,10 @@ func checkConnectSupernode(nodes []string) (localIP string) {
 
 func calculateTimeout(fileLength int64, defaultTimeoutSecond int, minRate int) time.Duration {
 	timeout := 5 * 60
+	// avoid trigger panic when minRate equals zero
+	if minRate <= 0 {
+		minRate = config.DefaultMinRate
+	}
 
 	if defaultTimeoutSecond > 0 {
 		timeout = defaultTimeoutSecond
--- a/cmd/dfget/app/root.go
+++ b/cmd/dfget/app/root.go
@@ -76,13 +76,14 @@ func runDfget() error {
 		return err
 	}
 
-	// get config from property files
-	initProperties()
-
 	if err := transParams(); err != nil {
 		util.Printer.Println(err.Error())
 		return err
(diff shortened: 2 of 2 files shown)


dfget/core/core.go
```diff
@@ -262,6 +262,10 @@ func checkConnectSupernode(nodes []string) (localIP string) {
 
 func calculateTimeout(fileLength int64, defaultTimeoutSecond int, minRate int) time.Duration {
 	timeout := 5 * 60
+	// avoid trigger panic when minRate equals zero
+	if minRate == 0 {
```
yeya24: Is it better for `if minRate <= 0 ` ?

jyizheng: Is it possible for a client to provide a negative value for minRate?

allencloud: Here is the definition `flagSet.StringVar(&minRate, "minrate", "",		flagSet.StringVar(&minRate, "minrate", "", "minimal network bandwidth rate for downloading a file, in format of 20M/m/K/k")`

And I think `<= 0` could get more assurance.

allencloud:
CI failure is unrelated:
```
FILE: ./README.md
[✖] https://codecov.io/gh/dragonflyoss/Dragonfly

ERROR: dead links found!
```

allencloud:
Please add more description for this PR. It is really useful for collaboration. And we should treat it as a good tradition to follow. Thanks a lot. @Starnop

dfget/core/core.go
Starnop: Done.

Starnop:
> Please add more description for this PR. It is really useful for collaboration. And we should treat it as a good tradition to follow. Thanks a lot. @Starnop

Done.

PR #674, opened 2019-07-13
feature: add more version information to both cmd and router

yeya24:

- Add more build information to all components, including dfget, dfdaemon and supernode.
- Unify the version cmd, now we can view the version information via version subcmd.
``` bash
➜  bin/linux_amd64/supernode version
supernode, version 0.4.0-50-gb772b6e
  git  branch:       feature/add-version
  git  commit:       b772b6e
  go  version:       go1.12.6
```
- Add a `/version` route to supernode, which will return a json response.
``` bash
➜  curl localhost:8002/version
{"version":"0.4.0-50-gb772b6e","revision":"b772b6e","branch":"feature/add-version","goVersion":"go1.12.6"}
```

NONE

Add a test in supernode router.

For cli:
``` bash
make build
dfget version
dfdaemon version
supernode version
```

For httpServer:
``` bash
curl localhost:65001/debug/version 
curl localhost:8002/version
```


yunfeiyangbuaa:
`  git  commit:       b772b6e`this option is unnecessary,WDYT?

yeya24:
> ` git commit: b772b6e`this option is unnecessary,WDYT?

This cmd is just providing some basic build information so I think it is ok to add this. Also this is important in metrics and you can find these labels in other OSS as well like prometheus、coredns, etc. 

Here it seems duplicated because the git commit is just the same with the last part of `version`.
I am not sure about the `version` parameter. Is it better to only keep the tags?
``` bash
➜  Dragonfly git:(feature/add-version) ✗ git describe --tags
v0.4.0-49-gf7c9d03

➜  Dragonfly git:(feature/add-version) ✗ git describe --tags `git rev-list --tags --max-count=1`
v0.4.0
```

supernode/server/router.go
```diff
@@ -20,6 +21,7 @@ func initRoute(s *Server) *mux.Router {
 	handlers := []*HandlerSpec{
 		// system
 		{Method: http.MethodGet, Path: "/_ping", HandlerFunc: s.ping},
+		{Method: http.MethodGet, Path: "/version", HandlerFunc: version.HandlerWithCtx},
```
allencloud: If you add this HTTP endpoint, you need to add this route into https://github.com/dragonflyoss/Dragonfly/blob/master/apis/swagger.yml first.

cmd/dfdaemon/app/version.go
```diff
@@ -0,0 +1,24 @@
...
+	"fmt"
+
+	"github.com/dragonflyoss/Dragonfly/version"
+
+	"github.com/spf13/cobra"
+)
+
+var versionCmd = &cobra.Command{
+	Use:           "version",
+	Short:         "Show the current version of dfdaemon",
```
allencloud: Could you add a `Long` field for every command, like 
```
Long:              "The dfdaemon is a proxy between container engine and registry used for pulling images.",
```

In addition, add an example for every command, like https://github.com/alibaba/pouch/blob/master/cli/create.go#L36.
It could be helpful when user executed `command -h`.

yeya24: Done

supernode/server/router.go
yeya24: I have already added the definition to swagger.yml. Do I have to generate the docs?

version/version.go
```diff
@@ -17,14 +17,117 @@
...
+	// populate via ldflags
+	branch string
+
+	// buildDate is the build date of project Dragonfly
 	// populate via ldflags
-	Version string
+	buildDate string
+
+	// goVersion is the version of Golang
+	goVersion = runtime.Version()
```
Starnop: Use the complete info with OS/Arch?

version/version.go
```diff
@@ -17,14 +17,117 @@
...
+type DragonflyVersion struct {
+	Version   string `json:"version"`
+	Revision  string `json:"revision"`
+	Branch    string `json:"branch"`
+	BuildDate string `json:"buildDate"`
+	GoVersion string `json:"goVersion"`
+}
+
+// Handler returns build information
+func Handler(w http.ResponseWriter, r *http.Request) {
```
Starnop: Extract the common code of  the functions  `Handler` and `HandlerWithCtx`.

supernode/server/router.go
Starnop: Yes, and you can do that by exec the shell `./hack/generate-swagger-models.sh`.

Starnop: The response should be `VersionInfo` instead of `DragonflyVersion ` according to definition in the `swagger.yml`.

yeya24: Thanks for your reply

version/version.go
yeya24: Done.

version/version.go
yeya24: Done.

supernode/server/router.go
yeya24: I have unified it.

yeya24:
@Starnop  What about change `version` to the specific tag? current version is `0.4.0-50-xxx`, I think we don't need the last part.
``` bash
➜  bin/linux_amd64/supernode version
supernode, version 0.4.0-50-gb772b6e
  git  branch:       feature/add-version
  git  commit:       b772b6e
  go  version:       go1.12.6
```

apis/types/df_get_task.go
```diff
@@ -8,8 +8,9 @@ package types
 import (
 	"encoding/json"
 
-	"github.com/go-openapi/errors"
 	strfmt "github.com/go-openapi/strfmt"
+
```
Starnop: It's unexpected. How did you update the files which generated by go-swagger?

version/version.go
```diff
@@ -17,14 +17,124 @@
...
+	t := template.Must(template.New("version").Parse(versionInfoTmpl))
+
+	var buf bytes.Buffer
+	if err := t.ExecuteTemplate(&buf, "version", m); err != nil {
+		panic(err)
+	}
+	return strings.TrimSpace(buf.String())
+}
+
+// DragonflyVersion contains build information about Dragonfly.
```
Starnop: You don't need to redefine `DragonflyVersion ` because it  will be generated by go-swagger which locates at `apis/types/dragonfly_version.go`.

Starnop:
> @Starnop What about change `version` to the specific tag? current version is `0.4.0-50-xxx`, I think we don't need the last part.
> 
> ```shell
> ➜  bin/linux_amd64/supernode version
> supernode, version 0.4.0-50-gb772b6e
>   git  branch:       feature/add-version
>   git  commit:       b772b6e
>   go  version:       go1.12.6
> ```

I think it's not necessary to change it. And it will give us a more accurate version of the information which will make it easier to locate problems.

apis/types/df_get_task.go
yeya24: I am not sure about this. Maybe it is the reason of swagger version? We update swagger to 0.19 in previous PR. And I know we used swagger 0.17.1 in the past. Is it the cause?

cmd/dfget/app/version.go
```diff
@@ -24,17 +24,33 @@ import (
...
+	Example: versionExample(),
 }
 
 func init() {
 	rootCmd.AddCommand(versionCmd)
 }
+
+// versionExample shows examples in version command, and is used in auto-generated cli docs.
+func versionExample() string {
+	return `$ dfget, version  0.4.1-14-g61aa7a9
```
lowzj: remove the character `$` and string `-14-g61aa7a9`.
Usually a release is built on a tag, and the version value is the tag name without its prefix `v`.

apis/types/df_get_task.go
Starnop: Yeah. That's true.

cmd/dfget/app/version.go
yeya24: Done

yeya24:
PTAL @Starnop

apis/swagger.yml
```diff
@@ -567,6 +581,33 @@ definitions:
...
+      branch:
+        type: "string"
+        description: "Git branch when building supernode"
+      buildDate:
+        type: "string"
+        description: "Build Date of supernode"
+      goVersion:
+        type: "string"
+        description: "Golang runtime version"
+      goOS:
```
allencloud: Why to add a prefix of `go` in `goOS`? It is some kind of weird. I think `OSType` is fine, since OS type is not related to go runtime.

apis/swagger.yml
```diff
@@ -567,6 +581,33 @@ definitions:
...
+        type: "string"
+        description: "Program's operating system"
+      goArch:
```
allencloud: the same issue.

apis/swagger.yml
```diff
@@ -567,6 +581,33 @@ definitions:
       message:
         type: string
 
+  DragonflyVersion:
```
allencloud: I think this is `SupernodeVersion`.

apis/swagger.yml
```diff
@@ -567,6 +581,33 @@ definitions:
...
+    description: |
+      Version and build information from supernode.
+    properties:
+      version:
+        type: "string"
+        description: "Version of supernode"
+      revision:
+        type: "string"
+        description: "Git commit when building supernode"
+      branch:
```
allencloud: I am wondering if we need this branch field. commit ID is uniquely to the commit history. Branch is redundant. Some users will directly use the source code to build supernode binary, while there is no branch. Or branch mean nothing, since I can build commit A from branch and I can also build commitB from branch.

yeya24: Make sense.

apis/swagger.yml
```diff
@@ -44,6 +44,20 @@ paths:
         500:
           $ref: "#/responses/500ErrorResponse"
 
+  /version:
+    get:
+      summary: "Get supernode's version"
+      description: |
+        Get supernode' s version and build information, including GoVersion, Version,
```
Starnop: s/supernode' s/supernode's

apis/swagger.yml
```diff
@@ -44,6 +44,20 @@ paths:
...
+        BuildDate, GitBranch and GitCommit.
```
Starnop: delete GitBranch?

cmd/dfdaemon/app/version.go
```diff
@@ -0,0 +1,40 @@
...
+	},
+	Example: versionExample(),
+}
+
+func init() {
+	rootCmd.AddCommand(versionCmd)
+}
+
+// versionExample shows examples in version command, and is used in auto-generated cli docs.
+func versionExample() string {
```
Starnop: I think you should update the examples. 😄

cmd/dfdaemon/app/version.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"fmt"
+
+	"github.com/dragonflyoss/Dragonfly/version"
+
+	"github.com/spf13/cobra"
+)
+
+// versionDescription is used to describe version command in detail and auto generate command doc.
+var versionDescription = "Display the version and build information of Dragonfly dfdaemon， " +
```
Starnop: Please update the versionDescription contents.

hack/build.sh
```diff
@@ -5,7 +5,9 @@ SUPERNODE_BINARY_NAME=supernode
 PKG=github.com/dragonflyoss/Dragonfly
 BUILD_IMAGE=golang:1.12.6
 VERSION=$(git describe --tags | cut -c 2-)
```
Starnop: >remove the character $ and string -14-g61aa7a9.
Usually a release is built on a tag, and the version value is the tag name without its prefix v.

version/version.go
```diff
@@ -17,14 +17,106 @@
...
 
 	// DFGetVersion is the version of dfget
-	DFGetVersion = Version
+	DFGetVersion = version
+
+	// SupernodeVersion is the version of supernode
+	SupernodeVersion = version
+
+	// DFVersion the global instance of DragonflyVersion
+	DFVersion *types.SupernodeVersion
```
Starnop: I think  you should make a consideration about the relationship between SupernodeVersion and DragonflyVersion.

yeya24: I still think DragonflyVersion is better because dfdaemon and dfget will this struct too.

allencloud: But DfDaemon and Supernode may be different version. Because there are in different binaries.
For that, separated SuperNodeVersion and DfDaemonVersion may be more reasonable. @yeya24

yeya24: Only when building with `make build`, is it a normal building process because we pass some ldflags in. The three components will be built at the same time and they are in the same version.  Did you mean in the procedure of running  `make build`, the dfdaemon and supernode may have different version? If that, I dont think it is reasonable.

Starnop:
LGTM.

Starnop:
@lowzj @allencloud  Do you have any questions about this PR? If not, I will merge it.

allencloud:
LGTM, thanks a lot for your works. @yeya24

PR #686, opened 2019-07-18
add a flag to pass timeout to dfget from dfdaemon

yeya24:

Add a flag to pass global timeout to dfget from dfdaemon.
This is simply a global timeout. Maybe later we can add a mechanism about setting different timeout or other dfget parameters for different tasks.

NONE

Maybe don't need to.

use `docker pull xxx`


Starnop:
In the PR #670, @lowzj  and I discussed the problem that we should provide a more flexible way to configure the dfget rather than pass the flags of dfdaemon to dfget. FYI https://github.com/dragonflyoss/Dragonfly/pull/670#pullrequestreview-262173464

And could you please help to open an issue about this problem？THX.

yeya24:
Since we have discussed further in https://github.com/dragonflyoss/Dragonfly/issues/687, I will close this now.

PR #692, opened 2019-07-18
docs: update readme about version information

allencloud:

docs: update readme about version information

Now we have rewritten Dragonfly in Golang version, and this will be the long term support one. We suggest that end users would adopt Golang version rather than Java version.

none

no need.

none

none


Starnop:
LGTM.

PR #696, opened 2019-07-19
docs: add a doc about how to add metrics

yeya24:

- add a doc about how to add metrics
- move NewMetrics logic to package common

NONE

### DIFF
--- a/apis/swagger.yml
+++ b/apis/swagger.yml
@@ -58,6 +58,17 @@ paths:
         500:
           $ref: "#/responses/500ErrorResponse"
 
+  /metrics:
+    get:
+      summary: "Get Prometheus metrics"
+      description: "Get Prometheus metrics"
+      responses:
+        200:
+          description: "no error"
+          schema:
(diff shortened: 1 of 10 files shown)


apis/swagger.yml
```diff
@@ -58,6 +58,17 @@ paths:
         500:
           $ref: "#/responses/500ErrorResponse"
 
+  /metrics:
+    get:
+      summary: "Get Prometheus metrics"
+      description: "Get Prometheus metrics"
+      responses:
+        200:
```
allencloud: I am wondering if there is any possibilities that the endpoint returns a non-200 status code?

yeya24: I think it is always 200. It is just a Get request without any  parameters. If this /metrics cannot work, the application will throw an error(maybe duplicate of metrics names or bad metric help), rather than returning non-200 status code.

Starnop:
LGTM. 

And @lowzj PTAL again. THX >

PR #700, opened 2019-07-22
Add support for arm32

hainesc:
Add support for arm32. Recently, we can not build binary for arm32 unless we implement function `Atime`, `Ctime`, `AtimeSec` and `CtimeSec` for arm32.

NONE

No test needed because I only add support for a special architecture.

```shell
for i in dfdaemon dfget supernode; do
  cd cmd/$i
  GOOS=linux GOARCH=arm GOARM=7 go build
  cd ../..
done
```


Starnop:
LGTM.

PR #701, opened 2019-07-22
feature: return supernode ip according to the value specified by peer

Starnop:

We must specify the supernode IP which can be connected by the peer node when we use dfget to download a file. However, the different peers have different network environments. And they will use different addresses to connect to the same supernode.  And for now, supernode will always return the IP address that will not change according to the peer node. That's not reasonable. 

So this PR will store the supernode IP which carried by dfget. When a peer download something from supernode, and the supernode will return the IP address that it specifies.

None.

None.

```
# build & install Dragonfly
make build && make install

# start supernode
supernode 

# start nginx server
https://github.com/dragonflyoss/Dragonfly/blob/master/docs/user_guide/install_server_go.md#procedure---when-deploying-with-physical-machines

# use dfget to download a file 
dfget -u https://www.taobao.com --node 127.0.0.1
```

And then checkout the dfclient log (`~/.small-dragonfly/logs/dfclient.log`) to validate the result

```
2019-07-22 07:19:45.639 INFO sign:6251-1563779985.294 : start to download piece from 127.0.0.1:8001 with taskID(9627226867cf6d8496944de44f554b8d0d2e743c991da21ad3e05336f01e5ba9)
```

### DIFF
--- a/apis/swagger.yml
+++ b/apis/swagger.yml
@@ -859,6 +859,10 @@ definitions:
         description: |
           PeerID is used to uniquely identifies a peer which will be used to create a dfgetTask.
           The value must be the value in the response after registering a peer.
+      supernodeIP:
+        type: "string"
+        description: "IP address of supernode which the peer connects to"
+        
 
   TaskCreateResponse:
     type: "object"
@@ -1162,6 +1166,9 @@ definitions:
           PeerID uniquely identifies a peer, and the cID uniquely identifies a 
           download task belonging to a peer. One peer can initiate multiple download tasks, 
           which means that one peer corresponds to multiple cIDs.
+      supernodeIP:
+        type: "string"
+        description: "IP address of supernode which the peer connects to"
  
   ErrorResponse:
     type: "object"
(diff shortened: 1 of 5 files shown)


dfget/core/downloader/p2p_downloader/power_client.go
```diff
@@ -111,6 +111,7 @@ func (pc *PowerClient) downloadPiece() (content *bytes.Buffer, e error) {
 		}
 	}
 
+	logrus.Infof("start to download piece from %s:%d with taskID(%s)", dstIP, peerPort, pc.taskID)
```
lowzj: Why does add this info log? I remember that there is already an info log to output the similar information. Could you confirm again?

Starnop: I have not found the similar log anywhere else. 😄

lowzj: There are already more detailed debug logs to output the information. I think you can just choose one and change its level to info if really necessary.

Starnop: Yeah. I got it. https://github.com/dragonflyoss/Dragonfly/blob/master/dfget/core/downloader/p2p_downloader/power_client.go#L77 

And then I will delete this log.

Starnop: Done. PTAL again. THX>

Starnop:
@lowzj  I have made a rebase. And it's ready to be merged now. 😄

PR #707, opened 2019-07-23
Add options for bash scripts

cd1989:

Add options to bash scripts:

```
set -o nounset
set -o errexit
```
Otherwise the `build.sh` will continue when some steps failed. The final result is build a image with missing artifacts.

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -1,4 +1,9 @@
 #!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
 DFDAEMON_BINARY_NAME=dfdaemon
 DFGET_BINARY_NAME=dfget
 SUPERNODE_BINARY_NAME=supernode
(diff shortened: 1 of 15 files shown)


hack/build.sh
```diff
@@ -1,4 +1,8 @@
 #!/bin/bash
+
+set -o nounset
```
Starnop: Maybe it's better to use `set -euo pipefail` instead.

Starnop:
In addition, could you please help to make a check for all scripts and do the same thing for them. THX >

hack/build.sh
cd1989: I'd prefer the following one, it's more readable.
```
set -o nounset
set -o errexit
set -o pipefail
```

cd1989:
Sure

Starnop:
LGTM.

PR #717, opened 2019-07-24
feature: add callsystem when task create

yeya24:

Add callSystem to taskRegisterRequest struct. Later we can add this information in metrics.

NONE

### DIFF
--- a/apis/swagger.yml
+++ b/apis/swagger.yml
@@ -702,6 +702,13 @@ definitions:
           tells whether it is a call from dfdaemon. dfdaemon is a long running
           process which works for container engines. It translates the image
           pulling request into raw requests into those dfget recognizes.
+      callSystem:
+        type: "string"
+        description: |
+          This attribute represents where the dfget requests come from. Dfget will pass
+          this field to supernode and supernode can do some checking and filtering via
+          black/white list mechanism to guarantee security, or some other purposes like debugging.
+        minLength: 1
 
   PeerCreateRequest:
     type: "object"
@@ -840,9 +847,9 @@ definitions:
       callSystem:
         type: "string"
         description: |
-          This field is for debugging. When caller of dfget is using it to files, he can pass callSystem
-          name to dfget. When this field is passing to supernode, supernode has ability to filter them via 
-          some black/white list to guarantee security, or some other purposes.
+          This attribute represents where the dfget requests come from. Dfget will pass
+          this field to supernode and supernode can do some checking and filtering via
+          black/white list mechanism to guarantee security, or some other purposes like debugging.
         minLength: 1
       filter:
         type: "array"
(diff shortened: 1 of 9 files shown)


allencloud:
I am wondering what is the usage of file https://github.com/dragonflyoss/Dragonfly/blob/master/supernode/server/0.3_bridge.go. Why is there a prefix of `0.3` of file? Do we have some particular meaning and what is the future plan of it? @Starnop @lowzj

apis/swagger.yml
```diff
@@ -963,9 +970,9 @@ definitions:
       callSystem:
         type: "string"
         description: |
-          This field is for debugging. When caller of dfget is using it to files, he can pass callSystem
-          name to dfget. When this field is passing to supernode, supernode has ability to filter them via 
-          some black/white list to guarantee security, or some other purposes.
+          This attribute represents where the dfget requests come from. Dfget will pass
```
Starnop: @lowzj  I don't think the `callSystem` and `dfdaemon` should be held by taskInfo. And maybe more reasonable for `dfgetTask`. WDYT?

Starnop:
> I am wondering what is the usage of file https://github.com/dragonflyoss/Dragonfly/blob/master/supernode/server/0.3_bridge.go. Why is there a prefix of `0.3` of file? Do we have some particular meaning and what is the future plan of it? @Starnop @lowzj

It's the implementation of the old API which represented by 0.3. And then, we will do the following things:
1. Implement the new API for Supernode.
2. Update the dfget to  use the new API.

apis/swagger.yml
lowzj: I think that it's necessary to pass the `callsystem` parameter to supernode when register a task, but there's no need to store it in task info.

Starnop: @yeya24 Could you please make a change for that?
+ delete the `dfdaemon` and `callSystem` fields from taskInfo
+ add them to `dfgetTask`

yeya24: @Starnop  Done. PTAL.

supernode/daemon/mgr/task/manager_util_test.go
```diff
@@ -36,7 +37,7 @@ func init() {
 type TaskUtilTestSuite struct {
 	mockCtl          *gomock.Controller
 	mockCDNMgr       *mock.MockCDNMgr
-	mockDfgetTaskMgr *mock.MockDfgetTaskMgr
+	dfgetTaskMgr     *dfgettask.Manager
```
Starnop: Use the mock mgr instead. And you can update the code about mock by exec the shell `hack/generate-super-mock-mgr`.

lowzj:
please rebase the code from Dragonfly/master to fix the failed CI flow.

supernode/daemon/mgr/task/manager_util_test.go
```diff
@@ -125,3 +122,81 @@ func (s *TaskUtilTestSuite) TestAddOrUpdateTask(c *check.C) {
 		c.Check(taskInfo, check.DeepEquals, v.task)
 	}
 }
+
+func (s *TaskUtilTestSuite) TestAddDfgetTask(c *check.C) {
```
Starnop: I think it's so simple that it's not necessary to add a unit test with `AddDfgetTask`.And if you want to add a test, maybe it's better to do it at `supernode/mgr/dfgettask`.

Starnop:
LGTM.

apis/swagger.yml
```diff
@@ -1169,7 +1163,20 @@ definitions:
       supernodeIP:
         type: "string"
         description: "IP address of supernode which the peer connects to"
- 
+      dfdaemon:
```
allencloud: I suggest that we could remove this parameter, this is so specific for dfdaemon. While I think we can make it more general for dfget, since we can set callSystem with value of `daemon`.

yeya24: I think these are two concepts. `dfget --dfdaemon --callsystem "cluster1"` represents a dfget request from dfdaemon, and the dfdaemon is in cluster1. It's improper to move `dfdaemon` to `callsystem` since they have different meaning. 

But I am wondering do we have any plan about `--dfdaemon` later? Will there be any difference between pure dfget and dfget from dfdaemon? I think black/white list will only apply to `callSystem`, if that, removing this parameter makes sense.

PR #718, opened 2019-07-24
bugfix: use access time to determine whether the file is expired

Starnop:

Now we use file.ModTime() to determine whether the file is expired. However, we just read the file when we upload pieces to other files. And the `modTime` has not changed. So the file will be marked expired and will be deleted after expire time even though it is providing the server for the other peers.

So I added a `accessTime` field which should be updated when doing anything with the file.

None.

None.

### DIFF
--- a/dfget/core/uploader/peer_server.go
+++ b/dfget/core/uploader/peer_server.go
@@ -87,12 +87,13 @@ type peerServer struct {
 
 // taskConfig refers to some name about peer task.
 type taskConfig struct {
-	taskID    string
-	rateLimit int
-	cid       string
-	dataDir   string
-	superNode string
-	finished  bool
+	taskID     string
+	rateLimit  int
+	cid        string
+	dataDir    string
+	superNode  string
+	finished   bool
+	accessTime time.Time
 }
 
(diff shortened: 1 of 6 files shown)


dfget/core/uploader/peer_server.go
```diff
@@ -482,7 +488,9 @@ func (ps *peerServer) deleteExpiredFile(path string, info os.FileInfo,
 		if ok && !task.finished {
 			return false
 		}
-		if time.Now().Sub(info.ModTime()) > expireTime {
+
+		// if the last access time is expireTime ago
+		if time.Since(task.accessTime) > expireTime {
```
lowzj: use the bigger one of `access time` and `modify time` to check whether the task is expired

lowzj:
rebase the code to fix the CI

Starnop:
> rebase the code to fix the CI

Done.

dfget/core/uploader/peer_server.go
Starnop: Done.

Starnop:
@lowzj Can we move on for this PR?  😄

PR #722, opened 2019-07-25
fix: markdown-link-check ignore twitter.com

lowzj:

```bash
FILE: README.md
[✖] https://twitter.com/dragonfly_oss → Status: 0 Error: Exceeded maxRedirects. Probably stuck in a redirect loop https://twitter.com/dragonfly_oss

ERROR: dead links found!
```
The failure is caused by some reason that we don't known(maybe a bug of `markdown-link-check`), but we can confirm that this link is ok and can be accessed successfully. So we ignore checking the urls of `twitter.com`.


Starnop:
LGTM.

Issue #704, opened 2019-07-22
[flaky test] RateLimiterSuite.TestRateLimiter_AcquireBlocking fails sometimes


allencloud:
### Ⅰ. Issue Description

In pr https://github.com/dragonflyoss/Dragonfly/pull/703, which is a doc-related PR, there is a unit test failure: https://circleci.com/gh/dragonflyoss/Dragonfly/8159?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
```
FAIL: rate_limiter_test.go:79: RateLimiterSuite.TestRateLimiter_AcquireBlocking

rate_limiter_test.go:103:
    c.Assert(end-start < cc.e+10, check.Equals, true)
... obtained bool = false
... expected bool = true

OOPS: 43 passed, 1 FAILED
--- FAIL: Test (5.96s)
FAIL
```

### Ⅱ. Describe what happened
There is a flaky test case there.

### Ⅲ. Describe what you expected to happen
No flaky test.

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1.
2.
3.

### Ⅵ. Environment:

- dragonfly version: 0.4.2
- OS (e.g. from /etc/os-release): ubuntu
- Kernel (e.g. `uname -a`): 
- Install tools:
- Others: master branch

allencloud:
Actually it have not occurred for the recent 20 days. We will keep looking for some time( half a month). If it does not happen again, we will close this.

PR #724, opened 2019-07-25
bugfix: fix overflow for rate limit

wybzju:
fix the overflow for rate limit
when use the function `createTokens` it may return negative number because of the overflow
The problem code is here `diff/(rl.window*time.Millisecond.Nanoseconds()) * rl.ratePerWindow`.
When `createTokens` return negative number it may cause sleep and it will make p2p download failed.
use int64 instead of int32 can solve this problem

### DIFF
--- a/common/util/assert.go
+++ b/common/util/assert.go
@@ -26,7 +26,7 @@ import (
 )
 
 // Max returns the larger of x or y.
-func Max(x, y int32) int32 {
+func Max(x, y int64) int64 {
 	if x > y {
 		return x
 	}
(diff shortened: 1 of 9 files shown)


allencloud:
Thanks a lot for your great work. Does this pull request fix issue https://github.com/dragonflyoss/Dragonfly/issues/704? @wybzju

wybzju:
@allencloud  I don't think so. I have encountered this issue #704, and when I retry, it can be fixed. And in other test case, I meet the same problem, for example `TestQueue_infiniteQueue`.
I think this issue is because the time error is only 10ms, and sometimes if the machine is high load, the issue may be reproduced. If we change 10ms to 20ms for example, the probability of reproducing this issue will be much lower.  But, I don't think this is a bug.

Starnop:
LGTM.

PR #727, opened 2019-07-26
feature: add access time for taskID

Starnop:

Add a accessTime for taskID to make it possible for GC. 

None.

None.

### DIFF
--- a/common/util/sys_util.go
+++ b/common/util/sys_util.go
@@ -1,0 +1,26 @@
+/*
+ * Copyright The Dragonfly Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


common/util/sys_util.go
```diff
@@ -0,0 +1,26 @@
...
+
+package util
+
+import (
+	"time"
+)
+
+// GetCurrentTimeMillis returns the time in millis for now.
+func GetCurrentTimeMillis() int64 {
+	return time.Now().UnixNano() / time.Millisecond.Nanoseconds()
```
yeya24: Is it better to make this function return float? float is more precise and then this function can be reused in metrics.
```
return float64(time.Now().UnixNano()) / float64(time.Millisecond.Nanoseconds())
```

Starnop: Maybe it's better to wrap it as another function, otherwise we need to make a lot of changes.

yeya24: Make sense.

PR #735, opened 2019-07-29
feature: supernode support download remote file by setting tls config.

zcc35357949:
We have a private harbor registry which support https with self-signed certification.I config ca path in /etc/dragonfly/dfdaemon.yml，but dfget does not pass it to supernode. So, supernode get X509 error when download from remote.
I add two dfget parameters "--insecure" and "--cacert", set "--insecure" to make supernode skip verify, and set 

- dfget command add two cmd flags,  "--insecure" and "--cacert".

- dfdaemon set "--insecure" and "--cacert" by dfdaemon config "hijack_https.hosts.insecure" and "hijack_https.hosts.certs".

- dfget set "--insecure" to make supernode skip verify when request remote source.

- dfget set "--cacert" to pass local ca info to supernode during registry, then supernode use it as rootcas when request remote source.


Starnop:
The CI has failed. And I make a copy here for you to handle it more conveniently.

ci/circleci: gocode-check-via-gometalinter-swagger
```
apis/types/task_info.go:1::warning: file is not gofmted with -s (gofmt)
common/util/http_util.go:1::warning: file is not gofmted with -s (gofmt)
dfdaemon/config/config.go:1::warning: file is not gofmted with -s (gofmt)
dfdaemon/downloader/dfget/dfget.go:1::warning: file is not gofmted with -s (gofmt)
supernode/daemon/mgr/cdn/downloader.go:1::warning: file is not gofmted with -s (gofmt)
supernode/daemon/mgr/task/manager_util.go:1::warning: file is not gofmted with -s (gofmt)
supernode/server/0.3_bridge.go:1::warning: file is not gofmted with -s (gofmt)
common/util/http_util.go:222:1:warning: exported function HTTPGetWithTls should have comment or be unexported (golint)
common/util/http_util.go:222:6:warning: func HTTPGetWithTls should be HTTPGetWithTLS (golint)
apis/types/task_info.go:1::warning: file is not goimported (goimports)
common/util/http_util.go:1::warning: file is not goimported (goimports)
dfdaemon/config/config.go:1::warning: file is not goimported (goimports)
dfdaemon/downloader/dfget/dfget.go:1::warning: file is not goimported (goimports)
dfget/core/regist/register.go:1::warning: file is not goimported (goimports)
supernode/daemon/mgr/cdn/downloader.go:1::warning: file is not goimported (goimports)
supernode/daemon/mgr/task/manager_util.go:1::warning: file is not goimported (goimports)
supernode/server/0.3_bridge.go:1::warning: file is not goimported (goimports)
```

ci/circleci: unit-test-golang 
```
# github.com/dragonflyoss/Dragonfly/supernode/daemon/mgr/cdn [github.com/dragonflyoss/Dragonfly/supernode/daemon/mgr/cdn.test]
supernode/daemon/mgr/cdn/downloader_test.go:104:27: not enough arguments in call to cm.download
	have (context.Context, string, string, map[string]string, int, int64, int32)
	want (context.Context, string, string, map[string]string, *tls.Config, int, int64, int32)
FAIL	github.com/dragonflyoss/Dragonfly/supernode/daemon/mgr/cdn [build failed]
Makefile:100: recipe for target 'unit-test' failed
```

apis/types/task_create_request.go
```diff
@@ -6,7 +6,7 @@ package types
 // Editing this file might prove futile when you re-run the swagger generate command
 
 import (
-	strfmt "github.com/go-openapi/strfmt"
+	"github.com/go-openapi/strfmt"
```
Starnop: You should not update the files under the dir `apis` . As an instead, you should  update the `swagger.yml` firstly and then run the script `./hack/generate-swagger-models.sh` to generate the code by `go-swagger`.

common/util/http_util.go
```diff
@@ -30,6 +30,7 @@ import (
 
 	errorType "github.com/dragonflyoss/Dragonfly/common/errors"
 
+	"crypto/tls"
```
Starnop: That's the reason why you cannot pass the CI. In fact, we recommend the imported packages should be cut into three parts and separate them with a blank line
+ standard packages
+ the packages from Dragonfly project
+ the packages from the third party projects

cmd/dfget/app/root.go
```diff
@@ -215,6 +215,8 @@ func initFlags() {
 
 	flagSet.StringVar(&cfg.CallSystem, "callsystem", "",
 		"The name of dfget caller which is for debugging. Once set, it will be passed to all components around the request to make debugging easy")
+	flagSet.StringSliceVar(&cfg.Cacert, "cacert", nil,
+		"The cacert file which is used to verify remote server when supernode download from source.")
```
Starnop: Use plural here is more suitable.

dfdaemon/config/config.go
```diff
@@ -137,7 +137,7 @@ func (p *Properties) Validate() error {
 
 // DFGetConfig returns config for dfget downloader
 func (p *Properties) DFGetConfig() DFGetConfig {
-	return DFGetConfig{
+	config := DFGetConfig{
```
Starnop: I think it's much better to use `dfgetConfig` as the variable name. WDYT/

dfget/core/regist/register.go
```diff
@@ -135,12 +136,19 @@ func (s *supernodeRegister) constructRegisterRequest(port int) *types.RegisterRe
...
+		Insecure:   cfg.Insecure,
 	}
 	if cfg.Md5 != "" {
 		req.Md5 = cfg.Md5
 	} else if cfg.Identifier != "" {
 		req.Identifier = cfg.Identifier
 	}
+
+	for _, certPath := range cfg.Cacert {
+		caBytes, _ := ioutil.ReadFile(certPath)
```
Starnop: I think it's not a good idea to ignore the error here.

Starnop:
In fact, the supernode will interact with the source URL for three times:
1. getHTTPFileLength: https://github.com/dragonflyoss/Dragonfly/blob/master/supernode/daemon/mgr/task/manager_util.go#L57
2. detect cache and ensure whether it's expired: https://github.com/dragonflyoss/Dragonfly/blob/master/supernode/daemon/mgr/cdn/cache_detector.go#L52
3. download from source URL:
https://github.com/dragonflyoss/Dragonfly/blob/master/supernode/daemon/mgr/cdn/downloader.go#L21

So it's not enough to only add `tlsConfig` for the step to download.  😄

apis/types/task_create_request.go
```diff
@@ -16,7 +16,13 @@ import (
 // TaskCreateRequest task create request
 // swagger:model TaskCreateRequest
 type TaskCreateRequest struct {
+	TaskCreateRequestBase
```
Starnop: Could you please guarantee the compatibility between the `dfget 0.3.x` and the `latest supernode`.

supernode/daemon/mgr/task/manager_util.go
```diff
@@ -35,6 +37,8 @@ func (tm *Manager) addOrUpdateTask(ctx context.Context, req *types.TaskCreateReq
 		TaskURL:    taskURL,
 		CdnStatus:  types.TaskInfoCdnStatusWAITING,
 		PieceTotal: -1,
+		Insecure:   req.Insecure,
+		RootCAs:    req.RootCAs,
```
Starnop: I am worried that whether it's suitable to put `RootCAs` content in the memory.

allencloud:
Thanks a lot for your work. That is very helpful.  @zcc35357949 
While I am wondering if we could encapsulte `tls.Config` into an HTTP client. And we can use a client to call the backend. Otherwise, in every function which will call backend, it will pass the parameter `tls.Config`, which makes the code much more redudant.

zcc35357949:
> client



> Thanks a lot for your work. That is very helpful. @zcc35357949
> While I am wondering if we could encapsulte `tls.Config` into an HTTP client. And we can use a client to call the backend. Otherwise, in every function which will call backend, it will pass the parameter `tls.Config`, which makes the code much more redudant.



> Thanks a lot for your work. That is very helpful. @zcc35357949
> While I am wondering if we could encapsulte `tls.Config` into an HTTP client. And we can use a client to call the backend. Otherwise, in every function which will call backend, it will pass the parameter `tls.Config`, which makes the code much more redudant.

ok, maybe I can try to save ca info to local instead of in task info in memory. 
In fact, different remote host may use different tls.Config, client should load ca info by host.

zcc35357949:
> Thanks a lot for your work. That is very helpful. @zcc35357949
> While I am wondering if we could encapsulte `tls.Config` into an HTTP client. And we can use a client to call the backend. Otherwise, in every function which will call backend, it will pass the parameter `tls.Config`, which makes the code much more redudant.

Thanks for your suggestions.
To avoid large change in supernode, I use a client that cached tls config, and apply tls config when do request.

dfget/config/config.go
```diff
@@ -165,6 +165,9 @@ type Config struct {
 	// default:`p2p`.
 	Pattern string `json:"pattern,omitempty"`
 
+	// CA certificate to verify when supernode download from source
```
Starnop: Not only download from source.  Maybe `Interact with the source` will be better. WDYT?

And so do with the comment of `Insecure`

supernode/server/0.3_bridge.go
```diff
@@ -88,6 +88,7 @@ func (s *Server) registry(ctx context.Context, rw http.ResponseWriter, req *http
 		TaskURL:     request.TaskURL,
 		SupernodeIP: request.SuperNodeIP,
 	}
+	cutil.DefaultHTTPSClient.RegisterTLSConfig(taskCreateRequest.RawURL, request.Insecure, request.RootCAs)
```
Starnop: I think we should create a http client to interact with the source for supernode instead of use the `cutil.DefaultHTTPSClient` which will bring some potential risks for other callers. WDYT?

common/util/http_util.go
```diff
@@ -235,7 +239,7 @@ func HTTPWithHeaders(method, url string, headers map[string]string, timeout time
 		req.Header.Add(k, v)
 	}
 
-	return http.DefaultClient.Do(req)
+	return DefaultHTTPSClient.Do(req)
```
Starnop: In fact, the `HTTPGet` will call this function finally. And `dfget` and `dfdaemon` will also use the function `HTTPGet`.

Starnop:
> Thanks a lot for your work. That is very helpful. @zcc35357949
> While I am wondering if we could encapsulte `tls.Config` into an HTTP client. And we can use a client to call the backend. Otherwise, in every function which will call backend, it will pass the parameter `tls.Config`, which makes the code much more redudant.

I totally agree with you. And how about add a `http_client` under the dir `supernode/` which wraps the apis that interact with the source.  @allencloud @lowzj @zcc35357949  WDYT?

zcc35357949:
> > Thanks a lot for your work. That is very helpful. @zcc35357949
> > While I am wondering if we could encapsulte `tls.Config` into an HTTP client. And we can use a client to call the backend. Otherwise, in every function which will call backend, it will pass the parameter `tls.Config`, which makes the code much more redudant.
> 
> I totally agree with you. And how about add a `http_client` folder under the dir `supernode/` which wraps the apis that interact with the source. @allencloud @lowzj @zcc35357949 WDYT?

Related interfaces:
- Download
- GetHTTPFileLength
- IsExpired
- IsSupportRange
these apis do not have any direct dependency with supernode, although only supernode is using them now. I'm not sure whether dfclient will use them in future.

Starnop:
LGTM.

allencloud:
Could you help to check if we need to update some docs? @Starnop @lowzj

Starnop:
> Could you help to check if we need to update some docs? @Starnop @lowzj

Yes we need. And @zcc35357949 will do it later.

PR #737, opened 2019-07-29
feature: hijacking https requests with different key pair

wybzju:

For use multi https backend, every host may need its own cert.

### DIFF
--- a/dfdaemon/proxy/proxy.go
+++ b/dfdaemon/proxy/proxy.go
@@ -35,6 +35,18 @@ func WithCert(cert *tls.Certificate) Option {
 func WithHTTPSHosts(hosts ...*config.HijackHost) Option {
 	return func(p *Proxy) error {
 		p.httpsHosts = hosts
+		for _, host := range p.httpsHosts {
+			if host.ServerCertFile != "" && host.ServerKeyFile != "" {
+				logrus.Infof("load cert from local file, %s, %s", host.ServerCertFile, host.ServerKeyFile)
+				cert, err := tls.LoadX509KeyPair(host.ServerCertFile, host.ServerKeyFile)
+				if err != nil {
+					logrus.Errorf("failed to load self-signed certificate (%s, %s) for https hijacking, host: %s, err: %v",
+						host.ServerCertFile, host.ServerKeyFile, host.Regx.String(), err)
+				}
+				logrus.Infof("use self-signed certificate (%s, %s) for https hijacking, host: %s", host.ServerCertFile, host.ServerKeyFile, host.Regx.String())
+				host.ServerCert = &cert
+			}
+		}
 		return nil
 	}
 }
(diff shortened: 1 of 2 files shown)


inoc603:
I'm a little confused by the title. I think your code means "hijacking https requests with different key pair", since the current implementation is not limited to a single host.

I can see how this may be useful, but if the key pair is self-signed, it's just as insecure as the default key pair used for hijacking.  Would you describe the use case you have in mind for this feature?

wybzju:
@inoc603 sorry, it's my mistake. Using different key pair is not for secure. When the different https backend using different key pair(although it is self-signed), the dafault key pair may not work. So, in this situation, we need to set different key pair for different https backend.

inoc603:
> @inoc603 sorry, it's my mistake. Using different key pair is not for secure. When the different https backend using different key pair(although it is self-signed), the dafault key pair may not work. So, in this situation, we need to set different key pair for different https backend.

client <-- hijacked --> dfdaemon <-- normal --> server

The key pair for hijacking is only used for the connection between the client and dfdaemon. The only case I can think of, where the default does not work, is when the client is set to trust the self-signed certificate of a server.In such case, the client can choose to ignore certificate error, or trust the certificate dfdaemon use.

wybzju:
@inoc603 
For example, one client(trust the self-signed certificate of a server), two servers(with different self-signed certificate and different domain). If the dfdaemon want to hijacked the request, we need to set different key pair which is serve used(don't need to generate new key pair) for different https server. In this situation, we need it. If it is not suggested to do, the pr can be closed. 

I think you are right. Maybe, we should generate new key pair for dfdaemon rather than use the server's key pair, and in this situation, the default cert can work.

lowzj:
@wybzju 
Could you give a concrete example containing the `dfdaemon.yml`, dockerd configuration, operation steps? It's much better for us to understand your intension, what problems you want to resolve and give our suggestions about this solution.

wybzju:
```
https-proxy.conf

[Service]
Environment="HTTPS_PROXY=http://127.0.0.1:65001"
```
```
dfdaemon.yml

proxies:
- regx: blobs/sha256
node:
-  127.0.0.1
hijack_https:
  hosts:
  - regx: harbor-a
    insecure: true
    serverCertFile: harbor-a.server.cert
    serverKeyFile: harbor-a.server.key
  - regx: harbor-b
    insecure: true
    serverCertFile: harbor-b.server.cert
    serverKeyFile: harbor-b.server.key
 - regx: harbor-c
    …
```
In this way, I can just use the harbors cert and key and don't need to generate a new key and cert, and the client don't need to trust a new cert. What's more, we don't need set insecure-registry for docker(we don't want to change /etc/docker/daemon.json)

dfdaemon/proxy/proxy.go
```diff
@@ -35,6 +35,18 @@ func WithCert(cert *tls.Certificate) Option {
 func WithHTTPSHosts(hosts ...*config.HijackHost) Option {
 	return func(p *Proxy) error {
 		p.httpsHosts = hosts
+		for _, host := range p.httpsHosts {
+			if host.ServerCertFile != "" && host.ServerKeyFile != "" {
+				logrus.Infof("load cert from local file, %s, %s", host.ServerCertFile, host.ServerKeyFile)
+				cert, err := tls.LoadX509KeyPair(host.ServerCertFile, host.ServerKeyFile)
+				if err != nil {
+					logrus.Errorf("failed to load self-signed certificate (%s, %s) for https hijacking, host: %s, err: %v",
```
allencloud: Do we need to return the err? Just comment to confirm.

wybzju: I think you are right, we should return err here

wybzju:
sorry, it's my mistake for closing the pr. I will create another one. Sorry

PR #739, opened 2019-07-30
bugfix: return error when the url is unreachable

Starnop:

The supernode will success to register the task even though when dfget try to download an 404 URL. It‘s not make sense. 

None.

### DIFF
--- a/supernode/daemon/mgr/task/manager.go
+++ b/supernode/daemon/mgr/task/manager.go
@@ -2,13 +2,14 @@ package task
 
 import (
 	"context"
+	"time"
 
 	"github.com/dragonflyoss/Dragonfly/apis/types"
 	"github.com/dragonflyoss/Dragonfly/pkg/errortypes"
 	"github.com/dragonflyoss/Dragonfly/pkg/httputils"
 	"github.com/dragonflyoss/Dragonfly/pkg/stringutils"
 	"github.com/dragonflyoss/Dragonfly/pkg/syncmap"
-	"github.com/dragonflyoss/Dragonfly/pkg/time"
+	"github.com/dragonflyoss/Dragonfly/pkg/timeutils"
 	"github.com/dragonflyoss/Dragonfly/supernode/config"
 	"github.com/dragonflyoss/Dragonfly/supernode/daemon/mgr"
 	dutil "github.com/dragonflyoss/Dragonfly/supernode/daemon/util"
@@ -32,9 +33,10 @@ var getContentLength = httputils.GetContentLength
 type Manager struct {
 	cfg *config.Config
 
-	taskStore     *dutil.Store
-	taskLocker    *util.LockerPool
-	accessTimeMap *syncmap.SyncMap
+	taskStore               *dutil.Store
+	taskLocker              *util.LockerPool
+	accessTimeMap           *syncmap.SyncMap
+	taskURLUnReachableStore *syncmap.SyncMap
 
 	peerMgr      mgr.PeerMgr
 	dfgetTaskMgr mgr.DfgetTaskMgr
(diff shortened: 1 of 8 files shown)


supernode/daemon/mgr/task/manager_util.go
```diff
@@ -24,6 +24,14 @@ func (tm *Manager) addOrUpdateTask(ctx context.Context, req *types.TaskCreateReq
 	}
 	taskID := generateTaskID(taskURL, req.Md5, req.Identifier)
 
+	isUnReachable := false
+	if value, err := tm.taskURLUnReachableStore.GetAsBool(taskID); err == nil && value == true {
+		isUnReachable = true
```
yeya24: Why do we need this bool variable? Here you can return error directly

supernode/daemon/mgr/task/manager_util.go
```diff
@@ -24,6 +24,14 @@ func (tm *Manager) addOrUpdateTask(ctx context.Context, req *types.TaskCreateReq
 	}
 	taskID := generateTaskID(taskURL, req.Md5, req.Identifier)
 
+	isUnReachable := false
```
allencloud: I suggest we declare a variable which is named by `isReachable` instead of `isUnReachable `. I think that could be more readable.

Starnop: We store the URL into the `taskURLUnReachableStore`. So I think maybe `isUnReachable` is better when we get it. WDYT?

supernode/daemon/mgr/task/manager_util.go
```diff
@@ -24,6 +24,10 @@ func (tm *Manager) addOrUpdateTask(ctx context.Context, req *types.TaskCreateReq
 	}
 	taskID := generateTaskID(taskURL, req.Md5, req.Identifier)
 
+	if isUnReachable, err := tm.taskURLUnReachableStore.GetAsBool(taskID); err == nil && isUnReachable == true {
```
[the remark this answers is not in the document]
Starnop: Done.

supernode/daemon/mgr/task/manager.go
```diff
@@ -16,7 +17,8 @@ import (
 )
 
 const (
-	key = ">I$pg-~AS~sP'rqu_`Oh&lz#9]\"=;nE%"
+	key                  = ">I$pg-~AS~sP'rqu_`Oh&lz#9]\"=;nE%"
+	cacheUnReachableTime = 3 * time.Minute
```
yeya24: I think it is better to make this as a flag. Not hard code it.

supernode/daemon/mgr/task/manager_util.go
```diff
@@ -464,17 +475,21 @@ func isWait(CDNStatus string) bool {
...
 	}
 
 	if code == http.StatusUnauthorized || code == http.StatusProxyAuthRequired {
-		return -1, errors.Wrapf(errorType.ErrAuthenticationRequired, "taskID: %s,code: %d", taskID, code)
+		return 0, errors.Wrapf(errorType.ErrAuthenticationRequired, "taskID: %s,code: %d", taskID, code)
 	}
 	if code != http.StatusOK {
 		logrus.Warnf("failed to get http file length with unexpected code: %d", code)
+		if code == http.StatusNotFound {
+			tm.taskURLUnReachableStore.Add(taskID, time.Now())
```
allencloud: Please do not make `getHTTPFileLength` to be a function of `*Manager`, because of adding this line of code. It works, but unreasonable for readability and unit testing.

Starnop: Done.

supernode/daemon/mgr/task/manager_util_test.go
```diff
@@ -65,19 +62,22 @@ func (s *TaskUtilTestSuite) TearDownSuite(c *check.C) {
 	s.contentLengthStub.Reset()
 }
 
-func (s *TaskUtilTestSuite) TestAddOrUpdateTask(c *check.C) {
+func (s *TaskUtilTestSuite) TestEqualsTask(c *check.C) {
```
yeya24: I am curious that why we change this test? I think you can add `TestEqualsTask` and keep the previous  `TestAddOrUpdateTask`

Starnop: In fact, I have been having a doubt for that when should we add unit tests for functions before until I saw [this article](https://juejin.im/entry/5bc00fb0f265da0af33498c1).

According to this article,I think `AddOrUpdateTask` is a function which is complicates and with many external dependencies. So we should not test this function directly but divide it into two parts: 
1. Handle complex logical parts 
2. Process dependent parts
And then write the unit testing for part one.

PR #740, opened 2019-07-30
feature: install the swagger if it has not been installed

Starnop:

Install the swagger if it has not been installed locally.

None.

```
make swagger-install
make swagger-generate
```


hack/generate-swagger-models.sh
```diff
@@ -4,11 +4,59 @@ set -o nounset
...
+    return
+  fi
+  echo true
+}
+
+# swagger::install installs the swagger binary.
+swagger::install() {
+    has_installed="$(swagger::check_version)"
+    if [[ "${has_installed}" = "true" ]]; then
+        echo "swagger-${SWAGGER_VERSION} has been installed."
```
yeya24: LGTM, just some nits.
If I have installed swagger, it will echo
``` bash
swagger-v0.19.0 has been installed.
```
Looks like a little confused that if I have installed swagger before or just installed through this script.
Is it better to change it to `swagger-${SWAGGER_VERSION} has already been installed. Skip installation.`

Starnop: Updated.

hack/swagger.sh
```diff
@@ -0,0 +1,72 @@
...
+    echo false
+    return
+  fi
+
+  version="$(swagger version | head -n 1 | cut -d " " -f 2)"
+  if [[ "${SWAGGER_VERSION}" != "${version}" ]]; then
+    echo false
+    return
+  fi
+  echo true
```
lowzj: we can just return:
* `0` means success
* `non-0` means fail

hack/swagger.sh
```diff
@@ -0,0 +1,72 @@
...
+    if [[ "${has_installed}" = "true" ]]; then
+        echo ">>>> Detected that swagger-${SWAGGER_VERSION} has already installed. Skip installation. <<<<"
+        return
+    fi
+
+    echo ">>>> Start to install swagger-${SWAGGER_VERSION}. <<<<"
+    
+    local url
+    url="https://github.com/go-swagger/go-swagger/releases/download/${SWAGGER_VERSION}/swagger_${GOOS}_${GOARCH}"
+    wget --quiet -O /usr/local/bin/swagger "${url}"
```
lowzj: I don't think it's a good idea to install an extra-tool in this directory. It may requires `root` permission, and make the user's operating system mess. If user has already installed `swagger` by himself, we just use it directly. If not, we can install them in our project's temporary directories: such as`.cache`, `bin` or others.

hack/swagger.sh
```diff
@@ -0,0 +1,69 @@
...
+SWAGGER_VERSION=v0.19.0
+
+. ./env.sh
+
+BUILD_PATH=../bin/${GOOS}_${GOARCH}
+
+SWAGGER_BIN="swagger"
+
+# swagger::check_version checks the command and the version.
+swagger::check_version() {
```
allencloud: I am wondering if this part of code is necessary. Just use the exit code and stdout to check the existence. Maybe be no need to use `has_installed`. From the meaning of `has_installed`, it should be some type like boolean, while it is a string.

lowzj:
I will close this pull request.
Please reopen it if anything updated.

PR #742, opened 2019-07-30
bugfix: the dfget will wait forever when the final piece failed

wybzju:

the case is:
1 the dfget pull piece task from supernode(for example, the total piece is two, pieceA, pieceB)
2 the dfget get two pieces and start to run download tasks
3 pieceB download successfully and put it into queue. pieceA is running 
4 so the function p2p.getItem can get one Item. Then try to pull task again
5 because the supernode have sent all pieces to dfget， dfget will receive waiting code, then sleep a moment and try to pull task again.
6 the pieceA download failed, and put it into queue, but it will never be dealt. And the step5 will run forever until timeout.

so if dfget receive the code waiting, it should judge the length of the queue. If the length is large than 0, it should break the cycle. Then the failed piece in the queue can be dealt.

### DIFF
--- a/dfget/core/downloader/p2p_downloader/p2p_downloader.go
+++ b/dfget/core/downloader/p2p_downloader/p2p_downloader.go
@@ -178,6 +178,8 @@ func (p2p *P2PDownloader) Run() error {
 			} else if code == constants.CodePeerFinish {
 				p2p.finishTask(response, clientWriter)
 				return nil
+			} else if code == constants.CodePeerWait {
+				continue
 			}
 
 			logrus.Warnf("request piece result:%v", response)
@@ -233,6 +235,9 @@ func (p2p *P2PDownloader) pullPieceTask(item *Piece) (
 		if res.Code != constants.CodePeerWait {
 			break
 		}
+		if p2p.queue.Len() > 0 {
+			break
+		}
 		sleepTime := time.Duration(rand.Intn(1400)+600) * time.Millisecond
 		logrus.Infof("pull piece task(%+v) result:%s and sleep %.3fs", item, res, sleepTime.Seconds())
 		time.Sleep(sleepTime)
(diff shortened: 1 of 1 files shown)


dfget/core/downloader/p2p_downloader/p2p_downloader.go
```diff
@@ -228,6 +230,9 @@ func (p2p *P2PDownloader) pullPieceTask(item *Piece) (
 		if res, err = p2p.API.PullPieceTask(item.SuperNode, req); err != nil {
```
allencloud: Could you please help to refactor the code here a little bit? @wybzju I think we could refactor that in this way:
```
for {
    res, err = p2p.API.PullPieceTask(item.SuperNode, req); 
    if err != nil {
        logrus.Errorf("pull piece task error: %v", err)
        break
    }
    if res.Code != constants.CodePeerWait {
        break
    }
    if p2p.queue.Len() > 0 {
        break
    }
    sleepTime := time.Duration(rand.Intn(1400)+600) * time.Millisecond
    logrus.Infof("pull piece task result:%s and sleep %.3fs", res, sleepTime.Seconds())
    time.Sleep(sleepTime)
}
```

I think that could be more readable. And the logic is the same. Please strictly follow https://github.com/alibaba/pouch/blob/master/docs/contributions/code_styles.md#rule006---return-fast-to-indent-less to improve that. 
/cc @Starnop @lowzj @yeya24

dfget/core/downloader/p2p_downloader/p2p_downloader.go
```diff
@@ -240,7 +245,8 @@ func (p2p *P2PDownloader) pullPieceTask(item *Piece) (
 	if res == nil || (res.Code != constants.CodePeerContinue &&
```
allencloud: We should abstract this judgement, otherwise it is so unreadable for developers. 
Above comment is not related to this PR.

dfget/core/downloader/p2p_downloader/p2p_downloader.go
wybzju: Done

wybzju: Done

Starnop:
LGTM.

Issue #366, opened 2019-02-01
Plan of supernode's golang SDK refactoring


allencloud:
## Golang SDK 

Dragonfly's supernode needs a Golang SDK which is stored in https://github.com/dragonflyoss/Dragonfly/tree/master/client. Then we achieve the following goals:

* every caller of supernode can take advantages of this SDK by integrating it
* golang sdk encapsulates of the details of HTTP communication
* it could provide scalability and portability for other system to integrate supernode.

In project Dragonfly, it will be used in component `dfget` and the `dashboard`. The refactoring plan of golang sdk is described in.

### System Related Endpoints

- GET /_ping: described in https://github.com/dragonflyoss/Dragonfly/blob/master/apis/swagger.yml#L34-L46
    - [ ] Pull Request:
    - [ ] Mock Test:

### Peer Related Endpoints

- POST /peers
    - [ ] Pull Request:
    - [ ] Mock Test:

- GET /peers
    - [ ] Pull Request:
    - [ ] Mock Test:

- GET /peers/{id}
    - [ ] Pull Request:
    - [ ] Mock Test:

- DELETE /peers/{id}
    - [ ] Pull Request:
    - [ ] Mock Test:

### Task Related Endpoints

- POST /tasks: described in 
    - [ ] Pull Request:
    - [ ] Mock Test:

- GET /peers
    - [ ] Pull Request:
    - [ ] Mock Test:

- GET /peers/{id}
    - [ ] Pull Request:
    - [ ] Mock Test:

- PUT /peers/{id}
    - [ ] Pull Request:
    - [ ] Mock Test:

- PUT /peers/{id}
    - [ ] Pull Request:
    - [ ] Mock Test:

zgfh:
@allencloud it mean task?
![image](https://user-images.githubusercontent.com/11729624/62135933-b393be00-b315-11e9-874a-247ad65bbec8.png)

allencloud:
Yes，I think you are right. @zgfh

PR #743, opened 2019-07-30
feature: add tasks create actions in client pkg of golang sdk

zgfh:
feature: add tasks create actions in client pkg of golang sdk

add peers implementation in sdk：

POST /tasks

this fixes one part of #366

none

none

### DIFF
--- a/client/task_create.go
+++ b/client/task_create.go
@@ -1,0 +1,38 @@
+/*
+ * Copyright The Dragonfly Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


client/task_create.go
```diff
@@ -0,0 +1,38 @@
...
+
+package client
+
+import (
+	"context"
+
+	"github.com/dragonflyoss/Dragonfly/apis/types"
+)
+
+// TaskCreate creates a peer node in supernode.
```
yeya24: Update this commend.

zgfh: updated

yeya24: remote `node`

allencloud:
Actually, in the current supernode, we still use the API endpoint of java version. We are still coupling the objects behind the API(not in the restful way), like in the https://github.com/dragonflyoss/Dragonfly/blob/master/supernode/server/router.go#L28-L32

What is our plan that starts to implement the restful API, including API server routes and API client?  @yeya24 @Starnop @lowzj

client/task_create.go
zgfh: @yeya24 very sorry, updated

zgfh:
if we have a plan to implement the restful API of tasks, I can help if need

allencloud:
LGTM

PR #682, opened 2019-07-18
bugfix: modify two supernode api's bug

yunfeiyangbuaa:
Signed-off-by: yunfeiyangbuaa <yunfeiyang@buaa.edu.cn>

### Ⅰ. Describe what this PR did

- First bug:when supernode send  response to dfget's reportpiece and servicedown request,it will return a nil response.So when the dfget do with the response with `json.Unmarshal(body, resp)` ,it will produce an `err unexpected end of JSON input` err.I add a status code info to the supernode's response to resolve these bugs and do with the err and status code
- Second bug:when dfget send pulltask request to supernode.if the supernode is off,the dfget will crash instead of find another supernode to finish its work.because the pulltask will retuen a `{}` instead of `nil`,the coder use `nil` to check the res,so it will happen `a invalid memory address or nil pointer dereference ` err .i make the return be nil when the supernode is off

### Ⅱ. Does this pull request fix one issue?

fixes #680

### Ⅳ. Describe how to verify it
you can add a fmt.println() to the previous code's api return err,it will report an err.and my code won't

PR #752, opened 2019-07-31
bugfix: the dfget will always receive fail result when change supernode

wybzju:

1 there have two supernode A and B
2 dfget registers to A and download some piece from A
3 suddenly, the A down
4 dfget will get error when pull piece task from A, then dfget register to node B
5 it will pull piece task again, but the item has dstCID. However, B doesn't have the task, and can't get dfgettask by dstCID and taskID. Then B will return err. Finally, dfget will receive error again

### DIFF
--- a/supernode/server/0.3_bridge.go
+++ b/supernode/server/0.3_bridge.go
@@ -121,9 +121,11 @@ func (s *Server) pullPieceTask(ctx context.Context, rw http.ResponseWriter, req 
 	if !cutil.IsEmptyStr(dstCID) {
 		dstDfgetTask, err := s.DfgetTaskMgr.Get(ctx, dstCID, taskID)
 		if err != nil {
-			return err
+			logrus.Warnf("failed to get dfget task by dstCID(%s) and taskID(%s), and the srcCID is %s, err: %v",
+				dstCID, taskID, srcCID, err)
+		} else {
+			request.DstPID = dstDfgetTask.PeerID
 		}
-		request.DstPID = dstDfgetTask.PeerID
 	}
 
 	isFinished, data, err := s.TaskMgr.GetPieces(ctx, taskID, srcCID, request)


yunfeiyangbuaa:
> Signed-off-by: wybzju [552044481@qq.com](mailto:552044481@qq.com)
> ### Ⅰ. Describe what this PR did
> 
> 1 there have two supernode A and B
> 2 dfget registers to A and download some piece from A
> 3 suddenly, the A down
> 4 dfget will get error when pull piece task from A, then dfget register to node B
> 5 it will pull piece task again, but the item has dstCID. However, B doesn't have the task, and can't get dfgettask by dstCID and taskID. Then B will return err. Finally, dfget will receive error again
> ### Ⅱ. Does this pull request fix one issue?
> ### Ⅲ. Why don't you add test cases (unit test/integration test)? (你真的觉得不需要加测试吗？)
> ### Ⅳ. Describe how to verify it
> ### Ⅴ. Special notes for reviews

could you put the error in github

yunfeiyangbuaa:
https://github.com/dragonflyoss/Dragonfly/pull/682 does the error is the same as the second bug in this pr

supernode/server/0.3_bridge.go
```diff
@@ -121,9 +121,11 @@ func (s *Server) pullPieceTask(ctx context.Context, rw http.ResponseWriter, req
 	if !cutil.IsEmptyStr(dstCID) {
 		dstDfgetTask, err := s.DfgetTaskMgr.Get(ctx, dstCID, taskID)
 		if err != nil {
-			return err
+			logrus.Warnf("srcCID:%s, failed to get dfget task by dstCID(%s) and taskID(%s), err: %v",
```
yunfeiyangbuaa: please output the log as the following format: logrus.Warnf("failed to ........")

yunfeiyangbuaa: I think it may be better
```
if err!=nil{
     logrus.Warnf("failed to ......")
     return  err
}
request.DstPID=.....
```
WDYT?

yunfeiyangbuaa: in the end,I want you to show me the detail error log.I recently work on this part

wybzju:
@yunfeiyangbuaa It is not the same with #682.
In pr #682, when supernode return err, dfget will panic. because res is not nil, but the BaseResponse is nil. `res == nil || (res.Code != constants.CodePeerContinue` so, here will panic.
And I fix it by using `err != nil || res == nil || (res.Code != constants.CodePeerContinue`  

In this pr, if the pr #682 is merged. when supernode down, dfget will register to another supernode. However, anther supernode don't have the dst dfgettask, it will return date not found err. The dfget will receive error again

yunfeiyangbuaa:
> @yunfeiyangbuaa It is not the same with #682.
> In pr #682, when supernode return err, dfget will panic. because res is not nil, but the BaseResponse is nil. `res == nil || (res.Code != constants.CodePeerContinue` so, here will panic.
> And I fix it by using `err != nil || res == nil || (res.Code != constants.CodePeerContinue`
> 
> In this pr, if the pr #682 is merged. when supernode down, dfget will register to another supernode. However, anther supernode don't have the dst dfgettask, it will return date not found err. The dfget will receive error again

I get what do you mean,so it means the multiple supernodes pattern can't work in fact?

wybzju:
@yunfeiyangbuaa yes, you are right, but it can be fixed. 
here is the log
```
2019-07-31 13:46:07.439 ERRO sign:2028-1564551966.183 : pull piece task error: dial tcp4 supernodeIP:8002: connect: connection refused
2019-07-31 13:46:07.439 ERRO sign:2028-1564551966.183 : pull piece task fail:{} and will migrate
2019-07-31 13:46:07.439 INFO sign:2028-1564551966.183 : do register to one of [supernodeIP:8002 supernodeIP:8002 supernodeIP:8002 supernodeIP:8002 supernodeIP:8002]
2019-07-31 13:46:07.441 INFO sign:2028-1564551966.183 : do register to supernodeIP:8002, res:{"code":200,"msg":"success","data":{"taskId":"taskID","fileLength":length,"pieceSize":4194304}} error:<nil>
2019-07-31 13:46:07.441 INFO sign:2028-1564551966.183 : do register result:{"code":200,"msg":"success","data":{"taskId":"taskID","fileLength":length,"pieceSize":4194304}} and cost:0.001s
2019-07-31 13:46:07.441 ERRO sign:2028-1564551966.183 : pull piece task error: 500:{"message":"{\"Code\":502,\"Msg\":\"key (dstCID@taskID): {\"Code\":0,\"Msg\":\"data not found\"}\"}"}
2019-07-31 13:46:07.441 ERRO sign:2028-1564551966.183 : pull piece task fail:{} and will migrate
```
then the dfget will always receive the data not found error until try all supernode


1 some dfclients start to pull images 
2 stop the corresponding supernode
3 you can see above log

yunfeiyangbuaa:
Thanks!

supernode/server/0.3_bridge.go
wybzju: Done

wybzju:
@yunfeiyangbuaa   what's more, you need to fix the panic problem first. Or you won't see above error, because the dfget will panic and not register to another supernode

yunfeiyangbuaa:
> @yunfeiyangbuaa what's more, you need to fix the panic problem first. Or you won't see above error, because the dfget will panic and not register to another supernode

I have make the res be nil.but I think the team should do more with the multiple supernodes pattern.because dfget can register to another supernode but can't download file in fact.I and you just do with the err.WDYT?

supernode/server/0.3_bridge.go
```diff
@@ -121,9 +121,11 @@ func (s *Server) pullPieceTask(ctx context.Context, rw http.ResponseWriter, req
 	if !cutil.IsEmptyStr(dstCID) {
 		dstDfgetTask, err := s.DfgetTaskMgr.Get(ctx, dstCID, taskID)
 		if err != nil {
-			return err
+			logrus.Warnf("failed to get dfget task by dstCID(%s) and taskID(%s), and the srcCID is %s, err: %v",
```
Starnop: I think it's not a good idea to ignore the error here. Instead, maybe we can delete the `dstCID` after migrate for `dfget`.

wybzju: I do not think so. Here, dstCID is used to obtain the corresponding peerID, and then the state of the corresponding peerID is updated. If there is no corresponding dfgetTask at suopernode, then there is no need to update the state of the peerID. So you can continue to execute. 
Of course, dfget can delete the dstCID and what's more, it need to delete dstCID for all items which are stored in the queue.

Starnop: OK. I agree with you. And I think we should speed ​​up the progress of HA. @yunfeiyangbuaa 😄

Starnop:
LGTM.

PR #757, opened 2019-07-31
feature: support multi https backend

wybzju:

For use multi https backend, every host may need its own cert.
ref #737

### DIFF
--- a/dfdaemon/config/config.go
+++ b/dfdaemon/config/config.go
@@ -59,7 +59,7 @@ var fs = afero.NewOsFs()
 //       direct: true
 //
 //     hijack_https:
-//       # key pair used to hijack https requests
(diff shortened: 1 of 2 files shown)


Starnop:
@wybzju Could you please rebase your code? And we will try to move it on. THX>

inoc603:
I still have some concerns about this feature.

1. Hijacking https requests with different key pairs can be useful, but I'm not quite convinced it's very necessary
2. Distributing private key of your registry to every instance in the cluster doesn't feel very safe.
3. Maybe we can try the *Use self-signed certificates* approach mentioned in [the docker documentation](https://docs.docker.com/registry/insecure/), to see if it works with proxy server, if your  usecase is mainly focused on image distribution.

PR #759, opened 2019-07-31
feature: add some supernode metrics

yeya24:

1. Add some internal metrics of supernode. The metrics I add are put in https://github.com/dragonflyoss/Dragonfly/pull/759/files#diff-da2f8a75b36b591f3f12425d653389ad.
2. Refact common metrics part. Add a register param to make registering metrics more flexible, especially in tests. 

NONE

I have added some unit tests. As for integration test, I will put it later.

### DIFF
--- a/docs/user_guide/metrics.md
+++ b/docs/user_guide/metrics.md
@@ -12,7 +12,6 @@ This doc contains all the metrics that Dragonfly components currently support. N
 - dragonfly_supernode_peers{hostname} - dragonfly peers
 - dragonfly_supernode_tasks{taskid, cdnstatus} - dragonfly tasks
 - dragonfly_supernode_dfgettasks{taskid, callsystem} - dragonfly dfget tasks
-- dragonfly_supernode_daemon_dfgettasks{taskid, callsystem} - dragonfly current dfget tasks, which are called from dfdaemon
 - dragonfly_supernode_schedule_duration_milliseconds{taskid} - duration for task scheduling in milliseconds
 - dragonfly_supernode_trigger_cdn_total{} - total times of triggering cdn.
 - dragonfly_supernode_trigger_cdn_failed_total{} - total failed times of triggering cdn.
--- a/docs/user_guide/metrics.md
+++ b/docs/user_guide/metrics.md
@@ -9,10 +9,10 @@ This doc contains all the metrics that Dragonfly components currently support. N
 - dragonfly_supernode_http_request_duration_seconds{code, handler, method} - http request latency in seconds
 - dragonfly_supernode_http_request_size_bytes{code, handler, method} - http request size in bytes
 - dragonfly_supernode_http_response_size_bytes{code, handler, method} - http response size in bytes
-- dragonfly_supernode_peers{hostname} - dragonfly peers
-- dragonfly_supernode_tasks{taskid, cdnstatus} - dragonfly tasks
-- dragonfly_supernode_dfgettasks{taskid, callsystem} - dragonfly dfget tasks
-- dragonfly_supernode_schedule_duration_milliseconds{taskid} - duration for task scheduling in milliseconds
+- dragonfly_supernode_peers{peer} - dragonfly peers, the label peer consists of the hostname and ip address of one peer.
+- dragonfly_supernode_tasks{cdnstatus} - dragonfly tasks
+- dragonfly_supernode_dfgettasks{callsystem} - dragonfly dfget tasks
+- dragonfly_supernode_schedule_duration_milliseconds{peer} - duration for task scheduling in milliseconds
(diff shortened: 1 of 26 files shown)


yeya24:
The previous PR uses `promauto` to register metrics, it will auto register metrics to default registry. Because of the duplication of metrics problem, it is kind of inflexible in tests. Maybe it is better to specify registry when registering metrics.

supernode/daemon/mgr/task/manager_util.go
```diff
@@ -116,36 +118,33 @@ func (tm *Manager) updateTask(taskID string, updateTaskInfo *types.TaskInfo) err
...
-	if !isSuccessCDN(updateTaskInfo.CdnStatus) {
-		// when the origin CDNStatus equals success, do not update it to unsuccessful
-		if isSuccessCDN(task.CdnStatus) {
-			return nil
-		}
-
-		// only update the task CdnStatus when the new CDNStatus and
-		// the origin CDNStatus both not equals success
+	// Update the origin task CDNStatus when the origin CDNStatus not equals success.
+	if !isSuccessCDN(task.CdnStatus) {
```
Starnop: The logic here has been changed. When the `update task` succeeds, the `origin task` should be updated regardless of whether it was successful before.

Starnop:
Could you please make a rebase? THX.

docs/user_guide/metrics.md
```diff
@@ -9,6 +9,13 @@ This doc contains all the metrics that Dragonfly components currently support. N
 - dragonfly_supernode_http_request_duration_seconds{code, handler, method} - http request latency in seconds
 - dragonfly_supernode_http_request_size_bytes{code, handler, method} - http request size in bytes
 - dragonfly_supernode_http_response_size_bytes{code, handler, method} - http response size in bytes
+- dragonfly_supernode_peers{hostname} - dragonfly peers
+- dragonfly_supernode_tasks{taskid, cdnstatus} - dragonfly tasks
+- dragonfly_supernode_dfgettasks{taskid, callsystem} - dragonfly dfget tasks
+- dragonfly_supernode_daemon_dfgettasks{taskid, callsystem} - dragonfly current dfget tasks, which are called from dfdaemon
+- dragonfly_supernode_schedule_duration_milliseconds{taskid} - duration for task scheduling in milliseconds
```
Starnop: IMOP, we should add a clientID label for it? WDYT/

supernode/daemon/mgr/task/manager_util.go
yeya24: Yes, thanks for pointing out this! I will change it.

docs/user_guide/metrics.md
yeya24: I have added peer label.

yeya24:
深度截图_选择区域_20190804194356

Here is what we should pay attention to. Currently I use the pair of hostname and ip address to construct the peer label of metrics. 
Seems like in a relative static network environment, which the ip address and hostname is fixed, it is OK. But the ip address maybe dynamic and the hostname is also changeable. I am wondering if we really should keep two. 
Any ideas about it? @lowzj @Starnop

yeya24:
Ready for review now. @Starnop  @lowzj @allencloud

Starnop:
In addition, I am wondering  that whether we can add some buried points calculate the inbound and outbound network traffic of the supernode. So that we can know the network pressure of supernode. In fact, the inbound network  traffic should including the traffic from the source URL. And the outbound network traffic can be collected by file server(such as nginx server)or calculated by adding all the total length of the piece that is dispatched to the supernode. 

WDYT? @yeya24 @lowzj

yeya24:
> In addition, I am wondering that whether we can add some buried points calculate the inbound and outbound network traffic of the supernode. So that we can know the network pressure of supernode. In fact, the inbound network traffic should including the traffic from the source URL. And the outbound network traffic can be collected by file server(such as nginx server)or calculated by adding all the total length of the piece that is dispatched to the supernode.
> 
> WDYT? @yeya24 @lowzj

Network traffic is not application-level, it is more like system or host level metrics. For example, you can use node exporter to get the network traffic for each of your NIC. If you run containers, you also have other exporters to calculate network traffic. So it should not be calculated in applcation metrics.

I know traffic is important and I will include inbound metrics in later PR, we can add the bytes size when triggering cdn. As for outbound traffic, I decide to put it in next part. When dfget upload metrics to supernode, we can add the download piece size in supernode.

Does this solve your question? @Starnop

yeya24:
Ready for review.

yeya24:
Can we move on this PR?

Starnop:
LGTM.

lowzj:
LGTM

PR #760, opened 2019-07-31
refactor: make common/util to be specific pkgs

allencloud:

This pull request tries to make the util packages of current project to be more clear.

Without this PR, there are several util packages:
* https://github.com/dragonflyoss/Dragonfly/tree/master/common/util
* https://github.com/dragonflyoss/Dragonfly/tree/master/supernode/util
* https://github.com/dragonflyoss/Dragonfly/tree/master/dfget/util
* https://github.com/dragonflyoss/Dragonfly/tree/master/supernode/daemon/util

The location is a little bit messy. And each of them actually provides general usages and has nothing to do the specific directory. This is not reasonable. And we can combine them all together into a pkg directory as a common usage.

In addition, all go files in the pkg folders is not a good way. Since every time to call a function of a struct's function in pkg folder, we should use `pkg.Object1FuncA` and `pkg.Object2FuncB`. Actually the best way is to call `Object1.FuncA` and `Object2.FuncB`, the `pkg` could only be used when importing packages.

none

no need

none

none


Starnop:
```
# github.com/dragonflyoss/Dragonfly/supernode/store [github.com/dragonflyoss/Dragonfly/supernode/store.test]
supernode/store/local_storage_test.go:31:2: imported and not used: "github.com/dragonflyoss/Dragonfly/pkg/stat" as util
supernode/store/local_storage_test.go:356:15: undefined: stat
FAIL	github.com/dragonflyoss/Dragonfly/supernode/store [build failed]
Makefile:100: recipe for target 'unit-test' failed
```

cmd/dfget/app/root.go
```diff
@@ -72,38 +71,38 @@ func init() {
...
 	// initialize logger
 	if err := initClientLog(); err != nil {
-		util.Printer.Println(fmt.Sprintf("init log error: %v", err))
+		fmt.Printf("init log error: %v\n", err)
 		return err
 	}
 
 	if err := transParams(); err != nil {
-		util.Printer.Println(err.Error())
+		fmt.Println(err.Error())
```
lowzj: I think that it's better to encapsulate the stdout, it's convenient for us to add some extra information in the future.

allencloud: Finished that. Thanks for your explanation. @lowzj

Starnop:
LGTM.

Issue #687, opened 2019-07-18
[feature request] provide a more flexible way to pass parameter from dfdaemon to dfget


yeya24:
## Why you need it?

Currently we just directly pass the parameters from dfdaemon to dfget. However these flags are global and sometimes different proxies may need different parameters such as timeout or urlfilters.

We should provide a more flexible way to construct dfget parameters in dfdaemon.

## Other related information

Related: 
https://github.com/dragonflyoss/Dragonfly/pull/670
https://github.com/dragonflyoss/Dragonfly/pull/686

Starnop:
Maybe we can add a string field for dfdaemon such as `dfgetParam` which will be passed to dfget just as @lowzj mentioned [here](https://github.com/dragonflyoss/Dragonfly/pull/670#pullrequestreview-262173464).

lowzj:
The `dfgetParam` of a proxy inherits and overrides the `flags` of `dfdaemon` passed to `dfget`.

yeya24:
Do we need that key-value map? What about just using k-v pair like `minrate: 20k`?

IMO if we want to use config files to pass variables, is it  confusing to still pass variables from dfdaemon flags? Maybe we can add a field of `globalParams` in configuration and remove most of flags from dfdaemon.

Starnop:
> Do we need that key-value map? What about just using k-v pair like `minrate: 20k`?

I agree with that.

Starnop:
> IMO if we want to use config files to pass variables, is it confusing to still pass variables from dfdaemon flags? Maybe we can add a field of `globalParams` in configuration and remove most of flags from dfdaemon.

Good suggestion. A globalParams is necessary and I think it's better to temporarily reserve the way to pass parameters via flag. WDYT?

So let's do these things next:
1. use config file to pass the values to dfget firstly and merge the configs from config file and flags.
2. provide a config template and with a script that automatically generates it

@lowzj @yeya24 Any other supplements？

lowzj:
agree

Starnop:
@yeya24 So would you like to do that? 😄

yeya24:
Sure
/assign

yeya24:
I have a question about how to deal with the config files between dfdaemon and dfget. If we set some dfgetParams, how we deal with dfget.yml?

lowzj:
The properties of `dfget` from CLI flags have a higher priority than config files.
The priority of properties that `dfdfaemon` passed to `dfget`:
```bash
[...]

PR #771, opened 2019-08-01
feature: add golbal dfget params to the dfdaemon config

Starnop:

fixes part of #687

TODO.

```
cat /etc/dragonfly/dfdaemon.yml
# node is the supernode address
# ip should be the ip address that the other nodes in the p2p network can access.
# port is the port that peer server will listen on
global_dfget_params: ["--node","192.168.33.21","--verbose","--ip","192.168.33.22","--port","15001","--alivetime","0s","--expiretime","5m0s"]
......
```

In my opition,  I think it's not a good way to use CLI args to pass values to the dfget compared to the config file. So I have not added a flag for dfget params. @lowzj @yeya24 WDYT?

In this PR, I only add a `globalDfgetParams` through the dfdaemon config file firstly. And we can add a `dfgetParam` for every proxy later.

 In addition, there is one more thing we need to pay attention to: how should we verify the correctness of the dfget params? Do you have any good ideas?

### DIFF
--- a/dfdaemon/config/config.go
+++ b/dfdaemon/config/config.go
@@ -40,6 +40,10 @@ var fs = afero.NewOsFs()
 // Properties holds all configurable properties of dfdaemon.
 // The default path is '/etc/dragonfly/dfdaemon.yml'
 // For examples:
+//     dfget_flags: ["--node","192.168.33.21","--verbose","--ip","192.168.33.23",
+//                   "--port","15001","--expiretime","3m0s","--alivetime","5m0s",
+//                   "-f","filterParam1&filterParam2"]
+//
 //     registry_mirror:
 //       # url for the registry mirror
 //       remote: https://index.docker.io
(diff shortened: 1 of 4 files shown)


dfdaemon/config/config.go
```diff
@@ -88,18 +88,14 @@ type Properties struct {
...
-	URLFilter  string   `yaml:"urlfilter" json:"urlfilter"`
-	CallSystem string   `yaml:"callsystem" json:"callsystem"`
-	Notbs      bool     `yaml:"notbs" json:"notbs"`
-
 	Verbose bool `yaml:"verbose" json:"verbose"`
 
 	MaxProcs int `yaml:"maxprocs" json:"maxprocs"`
+
+	// dfget config
+	GlobalDfgetParams []string `yaml:"global_dfget_params" json:"global_dfget_params"`
```
yeya24: I don't think this should be a slice. Can we define a struct for this such as `DfgetParams`? Then we can reuse it in proxy. And we can do some validation using this struct.

Starnop: If that, we have to redefine a struct just like `dfget config`.

yeya24: make sense

Starnop:
> Can you provide an example configuration? Since we remove the flags, I think it is better to add a configuration with some default global dfget params.

I have added the corresponding comment in the code. In addition, we need to add a docs dir which including all docs about configuration files. And we should make it ASAP.

Starnop:
@lowzj @yeya24 PTAL. THX

dfdaemon/config/config.go
lowzj: just use `dfgetflags`? And it should be assigned a default value.

dfdaemon/downloader/dfget/dfget.go
```diff
@@ -69,22 +69,15 @@ func (dfGetter *DFGetter) getCommand(
 		"-o", output,
 	}
 
-	if dfGetter.config.Notbs {
-		args = append(args, "--notbs")
-	}
-
-	if dfGetter.config.Verbose {
-		args = append(args, "--verbose")
+	for _, param := range dfGetter.config.GlobalDfgetParams {
```
lowzj: use `args = append(args, dfGetter.config.GlobalDfgetParams...)`

dfdaemon/downloader/dfget/dfget.go
```diff
@@ -69,22 +69,15 @@ func (dfGetter *DFGetter) getCommand(
 		"-o", output,
 	}
 
-	if dfGetter.config.Notbs {
-		args = append(args, "--notbs")
-	}
-
-	if dfGetter.config.Verbose {
```
lowzj: This should not be delete if the `--verbose` flag of `dfdaemon` has been set.

Starnop: Done. PTAL again. THX

PR #783, opened 2019-08-05
refactor: new api for dflog and output log to console by default

### LINKED ISSUE #651: [feature request] output logs to stdout
## Why you need it?

Dragonfly is used in cloud native situations and its components like dfdaemon and supernode maybe run as containers.

But currently the detailed logs are stored in logs file inside the container instead of outputing to stdout. If I want to see the logs file, I have to exec into the container first. It is better to have an option to configure logs output to stdout so I can use `docker logs` or `kubectl logs` to see the logs.

## How it could be?

Add an option or make it default to output logs to stdout

inoc603:

1. Refactored dflog, make log initialization consistent across dfdaemon/dfclient/supernode
2. Use [natefinch/lumberjack.v2](https://github.com/natefinch/lumberjack) for log rotation. It's a little different than the current approach we use, rather than rotating the same log file, it creates a new file and backup the current log file. If we still want to keep the current log rotating behavior, I can change it back.
3. Output dfdaemon's log to stdout by default

fixes #651 

Start dfdaemon and check the console output


cmd/supernode/app/root.go
```diff
@@ -145,12 +145,17 @@ func runSuperNode() error {
...
-	logPath := path.Join(options.HomeDir, "logs", "app.log")
-	err := dflog.InitLog(options.Debug, logPath, fmt.Sprintf("%d", os.Getpid()))
-	if err != nil {
-		logrus.Errorf("failed to initialize logs: %v", err)
+	logFilePath := path.Join(options.HomeDir, "logs", "app.log")
+
+	opts := []dflog.Option{
+		dflog.WithLogFile(logFilePath),
+		dflog.WithSign(fmt.Sprintf("%d", os.Getpid())),
+		dflog.WithDebug(options.Debug),
```
yeya24: Do we also need to make supernode output logs to console as an default option?

inoc603: I think so. But currently there's no configuration to disable console log. Should we provide options to turn it off?

pkg/dflog/log.go
```diff
@@ -20,113 +20,114 @@ import (
...
-// AddConsoleLog will add a ConsoleLog into Logger's hooks.
-// It will output logs to console when Logger's outputting logs.
-func AddConsoleLog(Logger *logrus.Logger) {
-	consoleLog := &logrus.Logger{
-		Out:       os.Stdout,
-		Formatter: Logger.Formatter,
-		Hooks:     make(logrus.LevelHooks),
-		Level:     Logger.Level,
+// Init initializes the logger with given options. If no option is provided,
+// the logger's fomatter will be set with an empty sign.
```
lowzj: `fomatter` -> `formatter`

pkg/dflog/log.go
```diff
@@ -20,113 +20,114 @@ import (
...
+			logger: consoleLog,
+			levels: logrus.AllLevels,
+		}
+		l.AddHook(hook)
+		return nil
 	}
+}
 
-	// open log file
-	logFile, err := os.OpenFile(logFilePath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0644)
```
lowzj: It's better to specify the `FileMode` as `0644`, all users can read the log file. If not, the created file's mode is relied on the os's `umask`. If the os's `umask` is `027`, then the new file's mode is `0640`, it means only the owner and users in same group can read the file.

inoc603: The version of lumberjack I'm using, [v2.0.0](https://github.com/natefinch/lumberjack/blob/v2.0.0/lumberjack.go#L215) did set the file mode to `0644`. However in [this commit](https://github.com/natefinch/lumberjack/commit/2e8fbeea3f9e456ee5062dfbcbe3ee902a44bace) it changes the default mode to `0600`.

lowzj: OK. I find that the previous code has same problem with this pr. If we want to create a file with permission `0644` exactly, we should explicitly call `syscall.Umask` to set `umask`.

lowzj:
LGTM

PR #789, opened 2019-08-05
fix valid ip not support ipv6

### LINKED ISSUE #334: 本地调试dfget 和 supernode，在客户端向supernode注册时客户端ip非法
### Ⅰ. Issue Description
      本地调试dfget 和 supernode，在客户端向supernode注册时客户端ip非法

### Ⅱ. Describe what happened
      查看客户端$HOME/.small-dragonfly下面的日志
image

    原因：
  当启动客户端dfget依次执行的runDfget方法中的core.Start(cfg)方法中prepare方法在赋值LocalIp时，CheckConnect方法获取的localip是[::1]（ipv6的localhost），然而服务器端isValidIp判断peer ip的方式是ipv4的方式,suo所以服务器返回ip of peer is illegal code 501
[image](https://user-images.githubusercontent.com/16266102/51292339-573a1280-1a45-11e9-9f0f-c30479dbb98b.png)
image
image
image
 
个人解决办法
1.dfget客户端修改
image
2.supernode客户端修改
3.

### Ⅵ. Environment:

- dragonfly version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

zgfh:
fix valid ip not support ipv6

fixes #334 

unit test

### DIFF
--- a/pkg/netutils/netutils.go
+++ b/pkg/netutils/netutils.go
@@ -211,16 +211,11 @@ func IsValidIP(ip string) bool {
 	if strings.TrimSpace(ip) == "" {
 		return false
 	}
-
-	// str is a regex which matches a digital
-	// greater than or equal to 0 and less than or equal to 255
-	str := "(?:25[0-5]|2[0-4]\\d|[01]?\\d?\\d)"
-	result, err := regexp.MatchString("^(?:"+str+"\\.){3}"+str+"$", ip)
-	if err != nil {
+	ipNet := net.ParseIP(ip)
+	if ipNet == nil {
 		return false
 	}
-
-	return result
+	return true
 }
 
 // GetAllIPs returns all non-loopback addresses.
--- a/pkg/netutils/netutils_test.go
+++ b/pkg/netutils/netutils_test.go
@@ -165,6 +165,18 @@ func (suite *UtilSuite) TestIsValidIP(c *check.C) {
 			expected: true,
 		},
 		{
+			ip:       "FE80::0202:B3FF:FE1E:8329",
+			expected: true,
+		},
+		{
+			ip:       "fe80::1",
+			expected: true,
+		},
+		{
+			ip:       "::1",
+			expected: true,
+		},
+		{
 			ip:       "255.255.255.255",
 			expected: true,
 		},


pkg/netutils/netutils.go
```diff
@@ -211,16 +211,11 @@ func IsValidIP(ip string) bool {
...
 		return false
 	}
-
-	// str is a regex which matches a digital
-	// greater than or equal to 0 and less than or equal to 255
-	str := "(?:25[0-5]|2[0-4]\\d|[01]?\\d?\\d)"
-	result, err := regexp.MatchString("^(?:"+str+"\\.){3}"+str+"$", ip)
-	if err != nil {
+	testInput := net.ParseIP(ip)
+	if testInput.To4() == nil && testInput.To16() == nil {
```
lowzj: > If s is not a valid textual representation of an IP address, ParseIP returns nil.

So it should test the result of `ParseIP` whether is nil here.

dzzg: thanks ,updated

pkg/netutils/netutils.go
```diff
@@ -211,16 +211,11 @@ func IsValidIP(ip string) bool {
 	if strings.TrimSpace(ip) == "" {
 		return false
 	}
-
-	// str is a regex which matches a digital
-	// greater than or equal to 0 and less than or equal to 255
-	str := "(?:25[0-5]|2[0-4]\\d|[01]?\\d?\\d)"
-	result, err := regexp.MatchString("^(?:"+str+"\\.){3}"+str+"$", ip)
-	if err != nil {
+	testInput := net.ParseIP(ip)
```
Starnop: Maybe `testInput` is not a good variable name. Could you please change that?

Starnop:
In fact, supernode will validate the IP format when dfget register to it. FYI https://github.com/dragonflyoss/Dragonfly/blob/master/apis/swagger.yml#L642

zgfh:
@Starnop sorry, I'm not understand you, 
1. I try to fix #334 (please have a look). you mean we should just print the err 'not support ipv6' when ip is ipv6?
2. test case is updated , what's you mean add a test ?

Starnop:
> 1. I try to fix #334 (please have a look). you mean we should just print the err 'not support ipv6' when ip is ipv6?

I mean it's not enough to update the validation of dfget. And you should update the supernode.

Starnop:
> 2\. test case is updated , what's you mean add a test ?

An integration test. We should ensure that there is a test to cover this scenario. WDYT? And you can add it under the dir [test](https://github.com/dragonflyoss/Dragonfly/tree/master/test).

zgfh:
@Starnop thanks a lot

pkg/netutils/netutils.go
zgfh: fixed

Starnop:
@zgfh Any updates? 😄

zgfh:
@Starnop I need more time to learn code or some help .
I find just update this not work https://github.com/dragonflyoss/Dragonfly/blob/76893b08820e9934657b90abeb1de9a94ef9a684/pkg/netutils/netutils.go#L202

Starnop:
> @Starnop I need more time to learn code or some help .
> I find just update this not work
> 
> https://github.com/dragonflyoss/Dragonfly/blob/76893b08820e9934657b90abeb1de9a94ef9a684/pkg/netutils/netutils.go#L202

FYI. The peer IP is defined in type IPv4 [here]( https://github.com/dragonflyoss/Dragonfly/blob/master/apis/swagger.yml).

And you the swagger will validate the param format. So when you call the API with a IPv6 value, you will get failed. 

AS for how to support both IPv4 and IPv6, you can define your own pattern regex.

zgfh:
@Starnop  
I find support ipv6 is a big feature.
but I find something else: the reason get a ipv6  is there return error: https://github.com/dragonflyoss/Dragonfly/blob/master/pkg/netutils/netutils.go#L227
image
if we just return ipv4, it will fix this.
should I create  a new PR to fix it?

Starnop:
> if we just return ipv4, it will fix this.
> should I create a new PR to fix it?

I don't think so. The current demand is that we need to support IPv6 instead of just not reporting an error. So I think we should move on this PR and make it complete.

And as for how to do that. Maybe it's not as troublesome as you think. The reason why we get failed when use IPv6 address as the IP value is that the validation rules do not allow incoming IPv6. And what we should do is to cancel the restriction. 

> FYI. The peer IP is defined in type IPv4 [here](https://github.com/dragonflyoss/Dragonfly/blob/master/apis/swagger.yml).
> 
> And you the swagger will validate the param format. So when you call the API with a IPv6 value, you will get failed.
> 
> As for how to support both IPv4 and IPv6, you can define your own [pattern](https://swagger.io/docs/specification/data-models/) regex.

yeya24:
Any progress now?

Starnop:
Unfortunately, there is no reply. And I think we need to reopen a PR to accomplish this. @fengzixu  Are you willing to do that? 😄

PR #791, opened 2019-08-06
feature: add part of tasks HTTP API

Starnop:

We have defined some new restful APIs for golang version supernode. According to the [swagger.yml](https://github.com/dragonflyoss/Dragonfly/blob/master/apis/swagger.yml), I added the implementation about the part of task CRUD.

fixes part of #348

None.

Now we could not delete the cdn files which on the disk. Later, we should implement it.


supernode/server/task_bridge.go
```diff
@@ -0,0 +1,82 @@
...
+
+	if err := request.Validate(strfmt.NewFormats()); err != nil {
+		return errors.Wrap(errortypes.ErrInvalidValue, err.Error())
+	}
+
+	resp, err := s.TaskMgr.Register(ctx, request)
+	if err != nil {
+		return err
+	}
+	return EncodeResponse(rw, http.StatusCreated, resp)
```
lowzj: the current `dfget` cannot handle the response code `http.StatusCreated`

Starnop: I know that. And we should use the new API for dfget later. 😄

lowzj:
closed.
please reopen it if anything updated.

PR #795, opened 2019-08-08
feature: not return error directly when failed to get http file Length

Starnop:

When supernode failed to get file length from source which is slow to be accessed due to `context deadline exceeded`, we should not return error directly and tell the client failed to register the task.  Instead, we should make it success unless something goes wrong with 404, 401,407, etc.
 
None.

None.


lowzj:
LGTM

PR #807, opened 2019-08-12
feature: add boilerplate check script

SataQiu:
1. add boilerplate check script (for go files)
2. complete all detected missing boilerplate
3. ignore generated files

./hack/check.sh

Next Steps:
1. add `update-boilerplate` script
2. add this check into CircleCI

### DIFF
--- a/hack/check.sh
+++ b/hack/check.sh
@@ -32,6 +32,14 @@ check() {
     echo "CHECK: go vet, check code syntax"
     packages=$(go list ./... | sed 's/^_//')
     echo "${packages}" | xargs go vet 2>&1
+
+    # boilerplate check
+    echo "CHECK: boilerpalte, check code boilerplate"
+    result=$(git ls-files | xargs go run ./hack/boilerplate/check-boilerplate.go)
+    if [[ ${#result} -gt 0 ]]; then
+        echo "${result}"
+        return 1
+    fi
 }
 
 check
--- a/cmd/dfdaemon/app/init.go
+++ b/cmd/dfdaemon/app/init.go
@@ -1,3 +1,19 @@
+/*
+ * Copyright The Dragonfly Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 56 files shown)


allencloud:
That is so good. I like it very much. Thanks so much. I will review this asap. @SataQiu 👏

hack/check.sh
```diff
@@ -32,6 +32,14 @@ check() {
     echo "CHECK: go vet, check code syntax"
     packages=$(go list ./... | sed 's/^_//')
     echo "${packages}" | xargs go vet 2>&1
+
+    # boilerplate check
+    echo "CHECK: boilerpalte, check code boilerplate"
+    result=$(git ls-files | xargs go run ./hack/boilerplate/check-boilerplate.go)
```
allencloud:
```
Dragonfly (new-807) $ git ls-files | xargs go run ./hack/boilerplate/check-boilerplate.go
Dragonfly (new-807) $ echo $?
0
version (new-807) $ vi version/version.go
Dragonfly (new-807) $ git ls-files | xargs go run ./hack/boilerplate/check-boilerplate.go
error validating "version/version.go": the file is missing a boilerplate
exit status 1
```
It works with the check file.

While I am afraid that `make check` fails since there is no docker daemon installed on my machine. I think `make file` should be friendly with developer's local machine, like windows, mac and so on. And there is no need to depend on the docker engine.

allencloud:
LGTM

hack/check.sh
SataQiu: Yes. I am also curious whether the `./hack/check-docker.sh` can be stripped out from `make check`.
But `hack/check.sh` is OK, right?
```Makefile
check:
	@echo "Begin to check code formats."
	./hack/check.sh
	@echo "Begin to check dockerd whether is startup"
	./hack/check-docker.sh
.PHONY: check
```

SataQiu:
> 
> 
> LGTM, but I think we should make it work in the integration test of travis.

@allencloud 
IMO, `travis` only runs `docker-build`, do you mean to add some checks, like `boilerplate-check` in `circleci`? 
How about do that in another PR? That will change circleci's behavior.

hack/check.sh
allencloud: > whether the ./hack/check-docker.sh can be stripped out from `make check`.

I think we should strip that from `make check`. 

`hack/hack.sh` is OK.

allencloud:
LGTM, I would like to merge this first. And then we can keep moving on to polish the check scripts to fit travisCI. Thanks for your work again. @SataQiu

PR #813, opened 2019-08-14
Feature: add code boilerplate check into circleci

SataQiu:
Feature: add code format check into circleci
1. add code boilerplate check as a circleci job
2. make `check-boilerplate.go` always exit with 0, so we can catch the output
3. fix gofmt check bug (ignore vendor)

make boilerplate-check

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,6 +49,10 @@ jobs:
           name: use golangci-lint to check gocode of this project.
           command: |
             make golangci-lint
+      - run:
+          name: boilerplate check
+          command: |
+            make boilerplate-check
 
   unit-test-golang:
     docker:
--- a/hack/boilerplate-check.sh
+++ b/hack/boilerplate-check.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+curDir=$(cd "$(dirname "$0")" && pwd)
+cd "${curDir}/../" || return
+
+check() {
+    result=$(git ls-files | xargs go run ./hack/boilerplate/check-boilerplate.go 2>&1)
+    if [[ ${#result} -gt 0 ]]; then
+        echo "${result}"
+        return 1
+    fi
+}
+
+check
--- a/hack/boilerplate/check-boilerplate.go
+++ b/hack/boilerplate/check-boilerplate.go
@@ -17,7 +17,6 @@
 package main
 
 import (
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -78,7 +77,7 @@ func checkBoilerplate(content string) error {
 		return fmt.Errorf("the file is missing a boilerplate")
 	}
 	if index < len(boilerplate) {
-		return errors.New("boilerplate has missing lines")
+		return fmt.Errorf("boilerplate has missing lines")
 	}
 	return nil
 }
@@ -86,7 +85,7 @@ func checkBoilerplate(content string) error {
 // verifyFile verifies if a file contains the boilerplate
 func verifyFile(filePath string) error {
 	if len(filePath) == 0 {
-		return errors.New("empty file name")
+		return fmt.Errorf("empty file name")
 	}
 
 	// check file extension is go
@@ -112,18 +111,13 @@ func verifyFile(filePath string) error {
 
 func main() {
 	if len(os.Args) < 2 {
-		fmt.Println("usage: go run check-boilerplate.go <path-to-file> <path-to-file> ...")
+		fmt.Fprintln(os.Stderr, "usage: go run check-boilerplate.go <path-to-file> <path-to-file> ...")
 		os.Exit(1)
 	}
 
-	hasErr := false
 	for _, filePath := range os.Args[1:] {
 		if err := verifyFile(filePath); err != nil {
-			fmt.Printf("error validating %q: %v\n", filePath, err)
-			hasErr = true
+			fmt.Fprintf(os.Stderr, "error validating %q: %v\n", filePath, err)
 		}
-	}
-	if hasErr {
-		os.Exit(1)
 	}
 }
--- a/hack/check.sh
+++ b/hack/check.sh
@@ -32,14 +32,6 @@ check() {
     echo "CHECK: go vet, check code syntax"
     packages=$(go list ./... | sed 's/^_//')
     echo "${packages}" | xargs go vet 2>&1
-
-    # boilerplate check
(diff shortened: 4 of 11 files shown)


.circleci/config.yml
```diff
@@ -102,6 +112,7 @@ workflows:
   ci:
     jobs:
       - markdownlint-misspell-shellcheck
+      - gocode-format-check
```
allencloud: You can add this in `gocode-check-via-gometalinter-swagger`.

hack/boilerplate/check-boilerplate.go
```diff
@@ -116,14 +116,9 @@ func main() {
 		os.Exit(1)
 	}
 
-	hasErr := false
 	for _, filePath := range os.Args[1:] {
 		if err := verifyFile(filePath); err != nil {
 			fmt.Printf("error validating %q: %v\n", filePath, err)
```
allencloud: This message should be printed in stderr, while now it is streamed to stdout, which seems to be incorrect. WDYT? @SataQiu

SataQiu: `pouchcontainer/pouchlinter:v0.2.1` has no GCC installed. 
Emm, I don't know where GCC is used, some cgo code?
Update `pouchcontainer/pouchlinter:v0.2.1` or create a new job using `circleci/golang:1.12.6` ?
WDYT? @allencloud

yeya24: We plan to replace gometalinter with golangci-lint and golangci-lint needs gcc dependency. Latest pouchcontainer/pouchlinter:v0.2.2 has installed gcc if you need it.

SataQiu: OK! Let's bump to `pouchcontainer/pouchlinter:v0.2.2`. :laughing:

yeya24: But in `pouchcontainer/pouchlinter:v0.2.2` we removed gometaliner. We need to update circle-ci part to migrate to golangci-lint first. So I will open a PR tonight and after that you can work on this PR. Is it ok? I will finish it ASAP.

SataQiu: OMG. 
`pouchcontainer/pouchlinter:v0.2.2` has no `gometalinter` installed. :(

hack/boilerplate/check-boilerplate.go
```diff
@@ -78,15 +77,15 @@ func checkBoilerplate(content string) error {
 		return fmt.Errorf("the file is missing a boilerplate")
 	}
 	if index < len(boilerplate) {
-		return errors.New("boilerplate has missing lines")
+		return fmt.Errorf("boilerplate has missing lines")
```
allencloud: I am wondering if the change is unnecessary, since I think both are the same. WDYT?

.circleci/config.yml
```diff
@@ -62,6 +62,9 @@ jobs:
           command: |
             make go-mod-vendor
             gometalinter --disable-all --skip vendor --skip test -E deadcode ./...
+      - run:
```
allencloud: I think we need rebase this part onto the latest master branch.

hack/boilerplate/check-boilerplate.go
SataQiu: Just want to clean something as I already imported `fmt`. ：）

SataQiu:
@allencloud @yeya24 
Considering that golanglint has checked gofmt, golint and govet, I recommend checking boilerplate separately.

yeya24:
> @allencloud @yeya24
> Considering that golanglint has checked gofmt, golint and govet, I recommend checking boilerplate separately.

Agree with that. But looks like the ci image get some problems here.

SataQiu:
Need review. @yeya24 @allencloud

hack/boilerplate-check.sh
```diff
@@ -0,0 +1,20 @@
...
+set -o nounset
+set -o errexit
+set -o pipefail
+
+curDir=$(cd "$(dirname "$0")" && pwd)
+cd "${curDir}/../" || return
+
+check() {
+    # boilerplate check
+    echo "CHECK: boilerpalte, check code boilerplate"
```
yeya24: This sentence looks weird. Why not just use `Begin to check code boilerplate` here?

hack/check.sh
```diff
@@ -10,7 +10,7 @@ cd "${curDir}/../" || return
 check() {
     # gofmt
     echo "CHECK: gofmt, check code formats"
-    result=$(find . -name '*.go' -print0 | xargs -0 gofmt -s -l -d 2>/dev/null)
+    result=$(find . ! -path './vendor*' -name '*.go' -print0 | xargs -0 gofmt -s -l -d 2>/dev/null)
```
yeya24: Do you need to add `vendor` here? @SataQiu

hack/boilerplate-check.sh
SataQiu: Because I write `Begin to check code boilerplate` in Makefile. Maybe just remove this? WDYT?

hack/check.sh
SataQiu: This is an implicit bug. It does not exclude possible `vendor` directory. If we run `make go-mod-vendor` and then run `make check`, it will failed.

SataQiu: We should not check vendor, Right?

yeya24: Yes, but after migrating to golangci-lint we don't run `make go-mod-vendor` anymore. So I think you can remove this.

hack/boilerplate-check.sh
yeya24: Yes, just keep one comment is OK.

allencloud:
Generally LGTM. While I still have a question, if I am new guy to contribute to dragonfly, when I forgot to add the boilerplate, then the CI would fail, but what exactly error message the CI output would throw to me? Is the error message friendly to a new guy?

SataQiu:
> 
> 
> Generally LGTM. While I still have a question, if I am new guy to contribute to dragonfly, when I forgot to add the boilerplate, then the CI would fail, but what exactly error message the CI output would throw to me? Is the error message friendly to a new guy?

Someting like this:
```
error validating "dfdaemon/config/config.go": the file is missing a boilerplate
```

allencloud:
> If we have a command like `make update-boilerplate`, that will be great.

No, I do not think so. I think the boilerplate should be made from the developer manually. Every should be aware of the copyright part and some others. If the script helps to do this, then lots of developer can escape from deep understanding the duties and obligations.

allencloud:
LGTM

PR #814, opened 2019-08-14
feature: replace gometalinter with golangci-lint

yeya24:

Since gometaliner is deprecated, we need to migrate golangci-lint.
Ref: https://github.com/alecthomas/gometalinter

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,11 +3,11 @@
 # Check https://circleci.com/docs/2.0/language-go/ for more details
 version: 2
 jobs:
-  markdownlint-misspell-shellcheck:
+  linters-check:
     docker:
       # this image is build from Dockerfile
       # https://github.com/pouchcontainer/pouchlinter/blob/master/Dockerfile
-      - image: pouchcontainer/pouchlinter:v0.2.2
+      - image: pouchcontainer/pouchlinter:v0.2.3
     working_directory: /go/src/github.com/dragonflyoss/Dragonfly
     steps:
       - checkout
@@ -32,7 +32,6 @@ jobs:
       - run:
           name: use opensource tool client9/misspell to correct commonly misspelled English words
           command: |
-            go get -u github.com/client9/misspell/cmd/misspell
             find  ./* -name  "*" | xargs misspell -error
       - run:
           name: use ShellCheck (https://github.com/koalaman/shellcheck) to check the validateness of shell scripts in pouch repo
@@ -42,16 +41,6 @@ jobs:
           name: validate go mod files
           command: |
             make check-go-mod
-
-  gocode-check-via-golangci-lint-swagger:
-    docker:
-      - image: pouchcontainer/pouchlinter:v0.2.2
-        environment:
-          GO111MODULE: "on"
-          GOPROXY: "https://goproxy.io"
-    working_directory: /go/src/github.com/dragonflyoss/Dragonfly
-    steps:
-      - checkout
       - run:
           name: validate swagger.yml
           command: |
@@ -59,7 +48,7 @@ jobs:
       - run:
           name: use golangci-lint to check gocode of this project.
           command: |
-            golangci-lint run
+            make golangci-lint
 
   unit-test-golang:
     docker:
@@ -99,7 +88,6 @@ workflows:
   version: 2
   ci:
     jobs:
-      - markdownlint-misspell-shellcheck
+      - linters-check
       - unit-test-golang
-      - gocode-check-via-golangci-lint-swagger
       - api-integration-test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ jobs:
     docker:
       # this image is build from Dockerfile
       # https://github.com/pouchcontainer/pouchlinter/blob/master/Dockerfile
-      - image: pouchcontainer/pouchlinter:v0.2.1
+      - image: pouchcontainer/pouchlinter:v0.2.2
     working_directory: /go/src/github.com/dragonflyoss/Dragonfly
     steps:
       - checkout
@@ -32,6 +32,7 @@ jobs:
       - run:
           name: use opensource tool client9/misspell to correct commonly misspelled English words
           command: |
+            go get -u github.com/client9/misspell/cmd/misspell
             find  ./* -name  "*" | xargs misspell -error
       - run:
           name: use ShellCheck (https://github.com/koalaman/shellcheck) to check the validateness of shell scripts in pouch repo
@@ -42,9 +43,12 @@ jobs:
           command: |
             make check-go-mod
 
-  gocode-check-via-gometalinter-swagger:
+  gocode-check-via-golangci-lint-swagger:
     docker:
-      - image: pouchcontainer/pouchlinter:v0.2.1
+      - image: pouchcontainer/pouchlinter:v0.2.2
+        environment:
+          GO111MODULE: "on"
+          GOPROXY: "https://goproxy.io"
     working_directory: /go/src/github.com/dragonflyoss/Dragonfly
     steps:
       - checkout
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -32,6 +32,7 @@ jobs:
       - run:
           name: use opensource tool client9/misspell to correct commonly misspelled English words
           command: |
+            go get -u github.com/client9/misspell/cmd/misspell
```
yeya24: This is a bug in the latest pouchlinter image because I forget to retrieve this binary after removing gometalinter. Should we update  pouchlinter image first? @allencloud

dfget/core/downloader/p2p_downloader/p2p_downloader.go
```diff
@@ -251,8 +252,9 @@ func (p2p *P2PDownloader) pullPieceTask(item *Piece) (
 
 	logrus.Errorf("pull piece task fail:%v and will migrate", res)
 	var registerRes *regist.RegisterResult
-	if registerRes, err = p2p.Register.Register(p2p.cfg.RV.PeerPort); err != nil {
-		return nil, err
+	var registerErr *errortypes.DfError
```
Starnop: No need to define `registerErr` here if we will not use it out of the  block 
```
if registerRes, registerErr = p2p.Register.Register(p2p.cfg.RV.PeerPort)
```

yeya24: registerErr is *errortypes.DfError type, if we re-use variable err here, it is always no-nil. See golangci-lint output below.
```
dfget/core/downloader/p2p_downloader/p2p_downloader.go:255:72: nilness: tautological condition: non-nil != nil (govet)
        if registerRes, err = p2p.Register.Register(p2p.cfg.RV.PeerPort); err != nil {
```

.circleci/config.yml
```diff
@@ -7,7 +7,7 @@ jobs:
     docker:
       # this image is build from Dockerfile
       # https://github.com/pouchcontainer/pouchlinter/blob/master/Dockerfile
-      - image: pouchcontainer/pouchlinter:v0.2.1
+      - image: pouchcontainer/pouchlinter:v0.2.2
```
allencloud: you could update this line to be version of `v0.2.3`.
see https://github.com/pouchcontainer/pouchlinter/pull/19

.circleci/config.yml
allencloud: Yes, I have updated that. so we can remove this line of code.

.circleci/config.yml
```diff
@@ -42,9 +43,12 @@ jobs:
           command: |
             make check-go-mod
 
-  gocode-check-via-gometalinter-swagger:
+  gocode-check-via-golangci-lint-swagger:
     docker:
-      - image: pouchcontainer/pouchlinter:v0.2.1
+      - image: pouchcontainer/pouchlinter:v0.2.2
```
allencloud: v0.2.3

.circleci/config.yml
```diff
@@ -53,15 +57,9 @@ jobs:
...
+          name: use golangci-lint to check gocode of this project.
           command: |
-            make go-mod-vendor
-            gometalinter --disable-all --cyclo-over=20 --skip vendor -E gofmt -E goimports -E golint -E misspell -E vet -E goconst -E gocyclo ./...
-      - run:
-          name: detect deadcode without test folder
-          command: |
-            make go-mod-vendor
-            gometalinter --disable-all --skip vendor --skip test -E deadcode ./...
+            golangci-lint run
```
allencloud: I am wondering if it is a good way to use `make golangci-lint` to do this part. Then this command in Makefile could be used locally on developers' environment. WDYT?

yeya24: Agree. I will update it soon.

yeya24:
Now we support run golangci-lint locally.
In docker:
```
➜ make golangci-lint USE_DOCKER=1
./hack/golangci-lint.sh
Begin to check gocode with golangci-lint in docker.
Check gocode with golangci-lint successfully 
```
If you run it local, you have to install golangci-lint first by yourself. I don't think it is a good idea to install this binary through this script.
```
./hack/golangci-lint.sh
Begin to check gocode with golangci-lint locally
Detected that golangci-lint has not been installed. You have to install golangci-lint first.
make: *** [Makefile:158：golangci-lint] 错误 1
```

Makefile
```diff
@@ -154,3 +153,7 @@ df.crt: df.key
 	openssl req -new -key df.key -out df.csr
 	openssl x509 -req -sha256 -days 365 -in df.csr -signkey df.key -out df.crt
 	rm df.csr
+
+golangci-lint:
+	./hack/golangci-lint.sh
+.PHONY: golangci-lint
```
Starnop: Please Add a blank break here.

hack/golangci-lint.sh
```diff
@@ -0,0 +1,72 @@
...
+        -e GO111MODULE=on                                                 \
+        -e GOPROXY="${GOPROXY}"                                           \
+        -w /go/src/${PKG}                                                 \
+        ${LINT_IMAGE}                                                     \
+        golangci-lint run
+}
+
+check-golangci-lint() {
+  has_installed="$(command -v golangci-lint || echo false)"
+  if [[ "${has_installed}" = "false" ]]; then
```
Starnop: Just return is OK here.

hack/build.sh
```diff
@@ -63,7 +63,7 @@ build-docker() {
         -e GOARCH="${GOARCH}"                                             \
         -e CGO_ENABLED=0                                                  \
         -e GO111MODULE=on                                                 \
-        -e GOPROXY=https://goproxy.io                                     \
+        -e GOPROXY="${GOPROXY}"                                           \
```
Starnop: Could you please update the `Makefile` and `env.sh` for environmental variable `GOPROXY `? THX

dfget/core/downloader/p2p_downloader/p2p_downloader.go
Starnop:
```
registerRes, registerErr := p2p.Register.Register(p2p.cfg.RV.PeerPort)
if registerErr != nil {
	return nil, registerErr
}
```

hack/golangci-lint.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+PKG=github.com/dragonflyoss/Dragonfly
+LINT_IMAGE=pouchcontainer/pouchlinter:v0.2.3
```
Starnop: In fact, there is no need to use `pouchcontainer/pouchlinter:v0.2.3` which contains a lot of other features and the size of this image is very huge. And you can write a `Dockerfile` and I will build and push it to the repo of [dragonflyoss](https://hub.docker.com/u/dragonflyoss).

hack/golangci-lint.sh
```diff
@@ -0,0 +1,72 @@
...
+    then
+        echo "Begin to check gocode with golangci-lint in docker."
+        golangci-lint-docker
+    else
+        echo "Begin to check gocode with golangci-lint locally"
+        golangci-lint-local
+    fi
+    echo "Check gocode with golangci-lint successfully "
+}
+
```
Starnop: Add a blank break int the end of the file.

.circleci/config.yml
```diff
@@ -42,9 +42,9 @@ jobs:
           command: |
             make check-go-mod
 
-  gocode-check-via-gometalinter-swagger:
+  gocode-check-via-golangci-lint-swagger:
```
Starnop: In fact, there is no need to start a new job for `golangci-lint`. If so, the CI will pull the image `pouchcontainer/pouchlinter:v0.2.3` twice. I think add it to the job `markdownlint-misspell-shellcheck` and as a step is OK. WDYT?

yeya24: Agree

.golangci.yml
```diff
@@ -0,0 +1,28 @@
+run:
+  deadline: 3m
+  modules-download-mode: readonly
+
+linters-settings:
+  gocyclo:
+    min-complexity: 20
+  goconst:
+    min-len: 3
+    min-occurrences: 5
```
Starnop: Remain the default value `3` is enough. 😢

yeya24: I will remove this goconst configuration.

hack/build.sh
yeya24: This env has already exists before. We set it to `https://goproxy.io`

hack/golangci-lint.sh
allencloud: No, I do not think the `make golangci-lint` command should depend on docker runtime.
If user installed the golangci-lint, it can directly use `make golangci-lint`.

SataQiu: Agree. We should try to keep the image light.

yeya24: Yes and I have already opened a PR for this linter image. https://github.com/dragonflyoss/dragonfly-scripts/pull/4

allencloud:
Thanks for your review. @SataQiu @Starnop 
LGTM as well. Merging.

PR #815, opened 2019-08-14
bugfix: add timeout for download piece

### LINKED ISSUE #864: What is the purpose of time.Sleep in function pullPieceTask of P2PDownloader?
## Question

I wonder that why there is time.Sleep in function pullPieceTask. It may slows down the whole download process.
Is it meant to reduce the pressure on supernode?

Thanks for answering!

https://github.com/dragonflyoss/Dragonfly/blob/3cbac7639dec3cb25a7c3c28ff0ac12558be5092/dfget/core/downloader/p2p_downloader/p2p_downloader.go#L232-L247

Starnop:

When dfget download files from other peers, we should set a timeout for that. And the timeout should be calculated by fileLength and the network rate.

fixes #864

Added.


pkg/netutils/netutils.go
```diff
@@ -277,3 +277,23 @@ func isExist(mmap map[string]bool, key string) bool {
...
+
+// CalculateTimeout calculate the timeout according to the fileLength and the min rate of network.
+//
+// The 0 will be returned when both minRate and defaultMinRate both are <=0.
+func CalculateTimeout(fileLength int64, minRate int, defaultMinRate int) time.Duration {
+	var timeout int
+
+	// ensure the minRate to avoid trigger panic when minRate equals zero
+	if minRate <= 0 && defaultMinRate <= 0 {
+		return time.Duration(0) * time.Second
```
allencloud: you can directly `return 0`.

pkg/netutils/netutils.go
```diff
@@ -277,3 +277,23 @@ func isExist(mmap map[string]bool, key string) bool {
 	}
 	return false
 }
+
+// CalculateTimeout calculate the timeout according to the fileLength and the min rate of network.
+//
+// The 0 will be returned when both minRate and defaultMinRate both are <=0.
+func CalculateTimeout(fileLength int64, minRate int, defaultMinRate int) time.Duration {
```
allencloud: And please add a unit test for this function.

Starnop: Done.

dfget/core/downloader/p2p_downloader/p2p_downloader.go
```diff
@@ -241,7 +241,7 @@ func (p2p *P2PDownloader) pullPieceTask(item *Piece) (
 		if p2p.queue.Len() > 0 {
 			break
 		}
-		sleepTime := time.Duration(rand.Intn(1400)+600) * time.Millisecond
+		sleepTime := netutils.CalculateTimeout(int64(item.PieceSize), p2p.cfg.MinRate, config.DefaultMinRate)
```
lowzj: The sleeping here is to wait for supernode to prepare an available piece. But the `MinRate` means the minimum speed of `dfget` downloading a piece. We cannot use the `MinRate` to calculate the `sleepTime`.

dfget/core/downloader/p2p_downloader/p2p_downloader.go
```diff
@@ -229,6 +229,10 @@ func (p2p *P2PDownloader) pullPieceTask(item *Piece) (
 		TaskID: item.TaskID,
 	}
 
+	// dfget will sleep some time which between minTimeout and maxTimeout,
+	// the initial value is [50-100]
+	minTimeout := 50
```
lowzj: The sleepTime would be re-initialized again when get the next available pieces from supernode. It's unnecessary and increases the pressure of supernode. Actually, the initial sleepTime in the next pulling can be the value of previous sleepTime.

Starnop: Has been updated. 😄

lowzj:
LGTM

PR #819, opened 2019-08-15
feature: limit the number which download from supernode for one task

Starnop:

When dragonfly begin to distribute a new file, a large number of download requests will be dispatched to the supernode node because there are no nodes in the current P2P network owns the file. Then the load of supernode will become heavier and heavier. According to the test, sometimes one peer will cost almost 2 minutes to download a piece which is only 4M from supernode(NOTE: the network bandwidth is enough).  In order to avoid this, we should add a limit which only allows a limited number of downloads from supernode for one task at the same time. And then the others will download the piece form the peers which have downloaded from the supernode. 

This will greatly reduce the load of the supernode.

None.

None.

### DIFF
--- a/supernode/daemon/mgr/progress/superload_manager.go
+++ b/supernode/daemon/mgr/progress/superload_manager.go
@@ -1,0 +1,78 @@
+/*
+ * Copyright The Dragonfly Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package progress
+
+import (
+	"context"
+	"time"
+
+	"github.com/dragonflyoss/Dragonfly/pkg/errortypes"
+)
+
+const (
+	// renewInterval is the interval time to check the superload.
+	renewInterval = 2 * time.Second
+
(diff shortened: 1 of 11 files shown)


supernode/daemon/mgr/progress/superload_manager.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		if !errortypes.IsDataNotFound(err) {
+			return false, err
+		}
+
+		// init the value for taskID when not found.
+		pm.superLoad.add(taskID, newSuperLoadState())
+	}
+
+	if loadState.loadValue.Add(delta) > limit && limit > 0 {
```
lowzj: The `loadState` maybe `nil`. It's not the object we created above when it's not exist before. We should get it again from map after adding. Otherwise we may use different objects with same `taskId` in multi-goroutines. It's better to make the `get` and `add` as one atomic operation.

Starnop: Done.

supernode/daemon/mgr/task/manager.go
```diff
@@ -272,6 +272,15 @@ func (tm *Manager) UpdatePieceStatus(ctx context.Context, taskID, pieceRange str
 			pieceRange, taskID, pieceUpdateRequest.ClientID)
 	}
 
+	// when a peer success to download a piece from supernode,
+	// and the load of supernode for the taskID should be decremented by one.
+	if tm.cfg.IsSuperPID(pieceUpdateRequest.DstPID) {
+		_, err := tm.progressMgr.UpdateSuperLoad(ctx, taskID, -1, -1)
```
lowzj: The load of supernode cannot be decreased here because that the `limit` is less than zero and the `if` statement below would not execute the code `loadState.loadValue.Add(delta)`.

```go
	if limit > 0 && loadState.loadValue.Add(delta) > limit {
		loadState.loadValue.Add(-delta)
		return false, nil
	}
```

Starnop: Thanks for pointing that. And I have updated it. PTAL again. THX>

lowzj:
LGTM

Issue #824, opened 2019-08-15
[feature request] add dfget flags check


yeya24:
## Why you need it?

We cannot ensure the dfget flags are correct now. For robustness of the program,  we must add some check mechanism to ensure configuration is correct.

inoc603:
I think we can refactor dfget/config.Properties a little bit, put all configurable options of dfget in there, like what we did with dfdaemon. Then in yaml files, we can use a struct rather than a list of strings, which is easier to to parse and validate.

yeya24:
> I think we can refactor dfget/config.Properties a little bit, put all configurable options of dfget in there, like what we did with dfdaemon. Then in yaml files, we can use a struct rather than a list of strings, which is easier to to parse and validate.

Agree. A struct for dfget params is what I want.

lowzj:
> I think we can refactor dfget/config.Properties a little bit, put all configurable options of dfget in there, like what we did with dfdaemon. 

Agree.

> Then in yaml files, we can use a struct rather than a list of strings, which is easier to to parse and validate.

Agree if the `yaml` you said here is `dfget.yml`. 

Not agree if it's `dfdaemon.yml`.

Although we often say that dragonfly is used to distribute images, in fact, dragonfly can support a variety of http file distribution. We can use `dfget` independently without `dfdaemon`. We can use the same `dfget` to download common files and download image layers on the same host.
It means that `dfget` and `dfdaemon` can be maintained independently. We can just upgrade `dfget` or `dfdaemon` without considering whether they're compatible.
If `dfdaemon` use the structure of `dfget` properties, when we want to add a flag to `dfget` and upgrade, we have to upgrade the `dfdaemon` at the same time, but it's unnecessary.

PR #826, opened 2019-08-15
feature: support dfget params for each proxy

yeya24:
Add support for each proxy to use different dfget params.

NONE

### DIFF
--- a/dfdaemon/config/config.go
+++ b/dfdaemon/config/config.go
@@ -79,6 +79,9 @@ type Properties struct {
 	// Proxies is the list of rules for the transparent proxy. If no rules
 	// are provided, all requests will be proxied directly. Request will be
 	// proxied with the first matching rule.
+	// You can also define custom dfget flags for each proxy. These extra flags
+	// will append to global dfget flags and extra flags can overwrite global flags
+	// if they have same parameters.
 	Proxies []*Proxy `yaml:"proxies" json:"proxies"`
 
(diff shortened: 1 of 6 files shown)


inoc603:
The idea is good and the code looks fine. 

But I'm a little concerned about the overriding dfget params part. Before we actually use a struct for dfget config as proposed in #824 , appending custom parameters to global parameters does not feel very robust. I'm sure it will work, but I think we should write tests to verify the override behavior. Also we should check if we print the final args which might have duplicated parameters in them.

Starnop:
Yeah, I agree with you. @yeya24 @inoc603 

And I think we can do  as @inoc603 said [here](https://github.com/dragonflyoss/Dragonfly/issues/824#issuecomment-521875334).

yeya24:
I am wondering some fields in dfdaemon config. Do you want to keep supernodes, ratelimit in dfdaemon's configuration? Since we already have GlobalDfgetConfig now, we can set dfget flags in it. 
As for dfget flags in each proxy, proxy can use its own nodes and rate configuration.

yeya24:
So what's the plan of this pr? Do I need to work on this again? Do we plan to rework the dfget configuration as my previous pr did?

yeya24:
@Starnop  Hello, do you think this feature still needed? If yes, I can do a rebase.

yeya24:
Close this pr since it seems not needed anymore.

PR #843, opened 2019-08-20
[WIP] feature: add dfget metrics

yeya24:

Add dfget metrics upload mechanism.

NONE


apis/swagger.yml
```diff
@@ -584,6 +584,24 @@ paths:
         500:
           $ref: "#/responses/500ErrorResponse"   
 
+  /task/metrics:
+    post:
+      summary: "registry a task"
+      description: |
+        Create a peer-to-peer downloading task in supernode.
```
allencloud: To see the endpoint， actually I cannot get the connection of the summary part and description part.

cmd/dfget/app/root.go
```diff
@@ -230,6 +230,8 @@ func initFlags() {
 		"identify whether supernode should skip secure verify when interact with the source.")
 	flagSet.IntVar(&cfg.ClientQueueSize, "clientqueue", config.DefaultClientQueueSize,
 		"specify the size of client queue which controls the number of pieces that can be processed simultaneously")
+	flagSet.BoolVar(&cfg.UploadMetrics, "uploadmetrics", false,
```
allencloud: How about making it true by default? 
I think we should encourage users could use the monitoring feature by default? WDYT？

dfget/core/core.go
```diff
@@ -189,8 +190,36 @@ func downloadFile(cfg *config.Config, supernodeAPI api.SupernodeAPI,
...
+			Duration:         downloadTime,
+			FileLength:       cfg.RV.FileLength,
+			Port:             int32(cfg.RV.PeerPort),
+			Success:          success,
+			TaskID:           result.TaskID,
+		}
+		for _, node := range cfg.Node {
+			resp, err := supernodeAPI.ReportMetrics(node, req)
+			if err != nil {
+				logrus.Error(err)
```
allencloud: Please see the error structure of function `ReportMetrics`, and you will find that this error log is unreadable when it is streamed in the logfile. please use:
```
logrus.Errorf("failed to report metrics from node %s to supernode node %s: %v", a, b, c)
```
WDYT?
Error and log construction is some kind of art. 😄

dfget/core/core.go
```diff
@@ -175,11 +177,10 @@ func downloadFile(cfg *config.Config, supernodeAPI api.SupernodeAPI,
 	}
 
 	timeout := calculateTimeout(cfg.RV.FileLength, cfg.Timeout, cfg.MinRate)
-	err := downloader.DoDownloadTimeout(getter, timeout)
-	success := "SUCCESS"
-	if err != nil {
+	success := true
+	if err = downloader.DoDownloadTimeout(getter, timeout); err != nil {
 		logrus.Error(err)
```
allencloud: same unreadable log here. Please help to update this.Thanks

dfget/core/core.go
```diff
@@ -189,8 +190,36 @@ func downloadFile(cfg *config.Config, supernodeAPI api.SupernodeAPI,
...
+	}
+
+	if success {
+		logrus.Infof("download SUCCESS cost:%.3fs length:%d",
+			time.Since(cfg.StartTime).Seconds(), cfg.RV.FileLength)
+	} else {
+		logrus.Infof("download FAIL cost:%.3fs length:%d reason:%d",
+			time.Since(cfg.StartTime).Seconds(), cfg.RV.FileLength, cfg.BackSourceReason)
+	}
 	return err
```
allencloud: `return err` is unreasonable, I think. 
Even if the report metric action fails, the function`downloadFile ` should return nil, right?

allencloud:
It is a good idea to do this. I like it.
However, I still have several question to make the situation clear. When a file from other peer(supernode) has finished downloading to the host peer, then multiple metrics data would be sent to supernodes. Does it make sense? Shall we evaluate whether it brings some kind of network traffic storm. It could be burden for the network. 
Just some question for the implementation.

apis/swagger.yml
yeya24: Sorry, I will recheck and update it.

yeya24:
> It is a good idea to do this. I like it.
> However, I still have several question to make the situation clear. When a file from other peer(supernode) has finished downloading to the host peer, then multiple metrics data would be sent to supernodes. Does it make sense? Shall we evaluate whether it brings some kind of network traffic storm. It could be burden for the network.
> Just some question for the implementation.

This is kind of a question about architecture. Yes, it will add some network traffic but I don't think it will really burden network traffic. For each dfget download task, we will finally upload metrics only once. Compared to other api calls such as `pull pieces`, which called several times in a loop during the dfget lifetime, this api will not cause something like network storm.

If we want to get some dfget related metrics, it is sure that this is a push-based mechanism because dfget is a short live job. So at lease we need one http call to implement this.

dfget/core/core.go
```diff
@@ -189,8 +190,36 @@ func downloadFile(cfg *config.Config, supernodeAPI api.SupernodeAPI,
...
+			CID:              cfg.RV.Cid,
+			CallSystem:       cfg.CallSystem,
+			Duration:         downloadTime,
+			FileLength:       cfg.RV.FileLength,
+			Port:             int32(cfg.RV.PeerPort),
+			Success:          success,
+			TaskID:           result.TaskID,
+		}
+		for _, node := range cfg.Node {
+			resp, err := supernodeAPI.ReportMetrics(node, req)
```
yeya24: I am not sure about this part. In the offline discuss with @lowzj and @Starnop , we agreed that it is OK that dfget upload to one supernode. I am wondering here if we need some shuffle or hash mechanism to distribute api requests? Otherwise dfget will always upload to one supernode

dfget/core/core.go
yeya24: No. This error stands for the error from function `DoDownloadTimeout`, we must return this error if that function failed because it relates to the main downloading logic. 
As for the error of reporting metrics, it will never be returned. We just log it.

dfget/core/core.go
yeya24: I prefer removing this log here. Because we will log the error and download status in the end of the function. Maybe we can add more fields in that log.
``` Go
	if success {
		logrus.Infof("download SUCCESS cost:%.3fs length:%d",
			time.Since(cfg.StartTime).Seconds(), cfg.RV.FileLength)
	} else {
		logrus.Infof("download FAIL cost:%.3fs length:%d reason:%d",
			time.Since(cfg.StartTime).Seconds(), cfg.RV.FileLength, cfg.BackSourceReason)
	}
```

cmd/dfget/app/root.go
yeya24: OK. I' ll change default to true

dfget/core/core.go
yeya24: Agree, I will update it

apis/swagger.yml
```diff
@@ -584,6 +584,25 @@ paths:
         500:
           $ref: "#/responses/500ErrorResponse"   
 
+  /task/metrics:
```
yeya24: I am not sure about this name, is there a better url for this?

dfget/core/core.go
allencloud: Then I suggest that encapsulate the code block you added into a single function. That brings much more certainty.

yeya24:
Fix the ci problem. PTAL again.

cmd/dfget/app/root.go
```diff
@@ -235,6 +235,8 @@ func initFlags() {
 		"identify whether supernode should skip secure verify when interact with the source.")
 	flagSet.IntVar(&cfg.ClientQueueSize, "clientqueue", config.DefaultClientQueueSize,
 		"specify the size of client queue which controls the number of pieces that can be processed simultaneously")
+	flagSet.BoolVar(&cfg.UploadMetrics, "uploadmetrics", true,
```
Starnop: Why do you want to add a flag of `uploadmetrics`? Is there any usage scenario? 
IMOP, monitoring is  something that is not perceived by the users. In addition, the metrics collected by the supernnode will not be complete if we let the user decide whether to report it or not.

dfget/core/core.go
```diff
@@ -269,3 +280,26 @@ func checkConnectSupernode(nodes []string) (localIP string) {
...
+		FileLength:       cfg.RV.FileLength,
+		Port:             int32(cfg.RV.PeerPort),
+		Success:          success,
+		TaskID:           taskID,
+	}
+	for _, node := range cfg.Node {
+		resp, err := supernodeAPI.ReportMetrics(node, req)
+		if err != nil {
+			logrus.Errorf("failed to report metrics to supernode %s: %v", node, err)
+		} else if resp.IsSuccess() {
```
Starnop: no need to use `else`.

apis/swagger.yml
Starnop: I think that's OK. FYI https://apiguide.readthedocs.io/en/latest/build_and_publish/use_RESTful_urls.html#

cmd/supernode/app/root.go
```diff
@@ -146,18 +147,26 @@ func runSuperNode() error {
...
 	logFilePath := path.Join(options.HomeDir, "logs", "app.log")
 
 	opts := []dflog.Option{
-		dflog.WithLogFile(logFilePath),
 		dflog.WithSign(fmt.Sprintf("%d", os.Getpid())),
 		dflog.WithDebug(options.Debug),
 	}
 
-	logrus.Debugf("use log file %s", logFilePath)
+	logrus.Debugf("use app log file %s", logFilePath)
```
Starnop: So weird here. Could you please encapsulate  the code about `init a log`?

yeya24: Done

supernode/server/metrics.go
```diff
@@ -70,3 +95,33 @@ func (m *metrics) instrumentHandler(handlerName string, handler http.HandlerFunc
...
+
+	if err := request.Validate(strfmt.NewFormats()); err != nil {
+		return errors.Wrap(errortypes.ErrInvalidValue, err.Error())
+	}
+	status := "success"
+	if !request.Success {
+		status = "failed"
+	}
+
+	dfgetLogger.Debugf("dfget download %s from peer %s, taskid %s, callsystem %s, filelength %d, "+
```
Starnop: s/`dfget download %s from peer %s  balabala`/`dfget peer %s download %s balabala`? WDYT?

yeya24: Done

cmd/supernode/app/root.go
```diff
@@ -122,8 +122,14 @@ func setupFlags(cmd *cobra.Command, opt *Options) {
...
 func runSuperNode() error {
-	// initialize log.
-	if err := initLog(); err != nil {
+	// initialize supernode logger.
+	if _, err := initLog(logrus.StandardLogger(), "app.log"); err != nil {
+		return err
+	}
+
+	// initialize dfget logger.
+	dfgetLogger, err := initLog(logrus.New(), "dfget.log")
```
Starnop: You can define the dfgetLogger beforehand，just like this:
```
// initialize dfget logger.
dfgetLogger=logrus.New()
err := initLog(dfgetLogger, "dfget.log")
```

PR #851, opened 2019-08-22
bugfix: modify gc logic err

### LINKED ISSUE #844: [bug]supernode server down  failed
### Ⅰ. Issue Description
```
func (s *Server) reportServiceDown(ctx context.Context, rw http.ResponseWriter, req *http.Request) (err error) {
	params := req.URL.Query()
	taskID := params.Get("taskId")
	cID := params.Get("cid")

	// get peerID according to the CID and taskID
	dfgetTask, err := s.DfgetTaskMgr.Get(ctx, cID, taskID)
	if err != nil {
		return err
	}
	s.ProgressMgr.UpdatePeerServiceDown(ctx, dfgetTask.PeerID)

	return EncodeResponse(rw, http.StatusOK, &types.ResultInfo{
		Code: constants.CodeGetPeerDown,
	})
}
```
 s.DfgetTaskMgr.Get(ctx, cID, taskID) dfgetTask will be delete before because of gc,so server down failed.

### Ⅱ. Describe what happened
server down failed

1.log the err
2.
3.

### Ⅵ. Environment:

- dragonfly version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

yunfeiyangbuaa:
Signed-off-by: yunfeiyangbuaa <yunfeiyang@buaa.edu.cn>

modify  gc logic err,if in previous logic,when download time is long,the peer will be delete during download 

fixes #844

### DIFF
--- a/supernode/daemon/mgr/gc/gc_peer.go
+++ b/supernode/daemon/mgr/gc/gc_peer.go
@@ -40,7 +40,7 @@ func (gcm *Manager) gcPeers(ctx context.Context) {
 			continue
 		}
 
-		if peerState.ServiceDownTime != 0 &&
+		if peerState.ServiceDownTime == 0 ||
 			timeutils.GetCurrentTimeMillis()-peerState.ServiceDownTime < int64(gcm.cfg.PeerGCDelay) {
 			continue
 		}


supernode/daemon/mgr/gc/gc_peer.go
```diff
@@ -40,7 +40,7 @@ func (gcm *Manager) gcPeers(ctx context.Context) {
 			continue
 		}
 
-		if peerState.ServiceDownTime != 0 &&
+		if peerState.ServiceDownTime == 0 ||
 			timeutils.GetCurrentTimeMillis()-peerState.ServiceDownTime < int64(gcm.cfg.PeerGCDelay) {
```
allencloud: Actually I like your former way to add a boolean to describe `timeutils.GetCurrentTimeMillis()-peerState.ServiceDownTime < int64(gcm.cfg.PeerGCDelay)`. However, I do not think `interval` is readable. We should use a more readable variable name, like `xxxxxxInterval`.

yunfeiyangbuaa: I remove the former way because if` peerState.ServiceDownTime==0`,supernode can go on instead of calculating `timeutils.GetCurrentTimeMillis()-peerState.ServiceDownTime` again

Starnop:
LGTM.

PR #854, opened 2019-08-23
bugfix: prevent dfgettask metrics < 0

### LINKED ISSUE #853: [bug] current dfget task metrics < 0
### Ⅰ. Issue Description
I run my supernode in master branch and find out that dfget tasks metrics value < 0 in the Grafana Dashboard. This is not what we expect.
image

### Ⅱ. Describe what happened
After digging into this code, I found that the problem is from gcing supernode dfget tasks.
When supernode trigger cdn, it will create a dfget task with prefix 'cdnnode', and its status will always be 'WAITING'. 

This dfget task will be deleted when gc, thus caused the metrics value < 0.
We should add a check when deleting dfget tasks.

This minimal value of this metrics is 0, we cannot decrease it.

1.
2.
3.

### Ⅵ. Environment:

- dragonfly version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

yeya24:

Add a check before deleting supernode dfget tasks

fixes #853 

No

### DIFF
--- a/supernode/daemon/mgr/dfgettask/manager.go
+++ b/supernode/daemon/mgr/dfgettask/manager.go
@@ -198,7 +198,9 @@ func (dtm *Manager) Delete(ctx context.Context, clientID, taskID string) error {
 		return err
 	}
 	dtm.ptoc.Delete(generatePeerKey(dfgetTask.PeerID, dfgetTask.TaskID))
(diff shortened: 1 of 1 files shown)


lowzj:
LGTM

PR #880, opened 2019-09-04
bugfix: create supernode homedir with perm 755

zcc35357949:

supernode's default homedir `/home/admin/supernode` is created with file mode 644.
That makes nginx can not return repo data in this dir for 403 forbidden.

None


yeya24:
LGTM. I met the same problem and it can also be fixed for specifying user in nginx config file.

zcc35357949:
> > supernode's default homedir /home/admin/supernode is created with file mode 644.
> > That makes nginx can not return repo data in this dir for 403 forbidden.
> 
> It's caused by the nginx and supernode are not stared by the same user.
> It's better to create `HomeDir` directly when supernode startup in the function `runSuperNode` rather than creating it indirectly by creating log file.

Ok, I have moved it to `runSuperNode`.

cmd/supernode/app/root.go
```diff
@@ -106,6 +106,11 @@ func setupFlags(cmd *cobra.Command, opt *Options) {
 
 // runSuperNode prepares configs, setups essential details and runs supernode daemon.
 func runSuperNode() error {
+	// create home dir
```
Starnop: Use `fiileutils.CreateDirectory` instead.

zcc35357949: Or running nginx as root can also solve this problem.

yeya24:
Maybe we can change the nginx config file?

Issue #836, opened 2019-08-18
error Config.LocalLimit description


zgfh:
### Ⅰ. Issue Description
the Properties.LocalLimit is `int`, but description is `format: 20M/m/K/k`
https://github.com/dragonflyoss/Dragonfly/blob/master/dfget/config/config.go#L62

should I change Properties.LocalLimit to `string` and transform to int use `transParams`?

yeya24:
I hope we can make the configuration more readable.  
For some configuration like `timeout`, we still use int type like 3600. Change to string is much readable such as `5d`, `10m`.

Starnop:
Yeah, please help to update the comments. THX.

yeya24:
Actually I don't think it is a problem about the comment.
Just using int type to represent these types like `duration` or `rate` is kind of unreadable.
I hope we can use some string to represent these. Like `5d`, `10m`. For rate, we can use `10K` or `1M`. Using this is much better than just use integer like `10407568`

WDYT @Starnop @allencloud  @lowzj

inoc603:
For duration, `time.Duration` can be unmarshalled from strings like `5d`, `10m` in yaml and json by default. For rates, we can change int to a named type `type Rate int`, and add `UnmarshalJSON(YAML)` and `String` method to it, to make the refactor as smooth as possible.

yeya24:
> For duration, `time.Duration` can be unmarshalled from strings like `5d`, `10m` in yaml and json by default. For rates, we can change int to a named type `type Rate int`, and add `UnmarshalJSON(YAML)` and `String` method to it, to make the refactor as smooth as possible.

Yes, I will open a PR soon.
/assign

PR #881, opened 2019-09-05
feature: support rate type

### LINKED ISSUE #879: [Help wanted] feature: support dfget configuration items support formats such as 20M/m/K/k
## Why you need it?

The users have to convert config values to bytes when users want to config the dfget item such as `localLimit`, `totalLimit` and `minRate`. That's not user-friendly. So we should make these items support formats such as `20M/m/K/k`.

## Other related information

Please pay attention to [this PR](https://github.com/dragonflyoss/Dragonfly/pull/802#discussion_r320548118).

yeya24:

1. Add model.Rate type to represent rate type, includes GB, MB, KB and B.

fixes #879 #836 

Added some tests already.

### DIFF
--- a/cmd/dfget/app/root.go
+++ b/cmd/dfget/app/root.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"os"
 	"path"
-	"strconv"
 	"strings"
 	"time"
 
@@ -36,12 +35,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var (
-	localLimit string
-	totalLimit string
-	minRate    string
-	filter     string
-)
+var filter string
 
 var cfg = config.NewConfig()
 
@@ -75,9 +69,7 @@ func runDfget() error {
 		return err
 	}
 
-	if err := transParams(); err != nil {
-		return err
-	}
+	cfg.Filter = transFilter(filter)
 
 	// get config from property files
 	initProperties()
@@ -145,27 +137,6 @@ func initProperties() {
 	}
 }
 
-// transParams trans the user-friendly parameter formats
-// to the format corresponding to the `Config` struct.
-func transParams() error {
-	cfg.Filter = transFilter(filter)
-
-	var err error
-	if cfg.LocalLimit, err = transLimit(localLimit); err != nil {
-		return errors.Wrapf(errortypes.ErrConvertFailed, "locallimit: %v", err)
-	}
-
-	if cfg.MinRate, err = transLimit(minRate); err != nil {
-		return errors.Wrapf(errortypes.ErrConvertFailed, "minrate: %v", err)
-	}
-
-	if cfg.TotalLimit, err = transLimit(totalLimit); err != nil {
-		return errors.Wrapf(errortypes.ErrConvertFailed, "totallimit: %v", err)
-	}
-
-	return nil
-}
-
 // initClientLog initializes dfget client's logger.
 // There are two kinds of logger dfget client uses: logfile and console.
 // logfile is used to stored generated log in local filesystem,
@@ -198,13 +169,13 @@ func initFlags() {
 		"Destination path which is used to store the requested downloading file. It must contain detailed directory and specific filename, for example, '/tmp/file.mp4'")
 
 	// localLimit & minRate & totalLimit & timeout
-	flagSet.StringVarP(&localLimit, "locallimit", "s", "",
-		"network bandwidth rate limit for single download task, in format of 20M/m/K/k")
-	flagSet.StringVar(&minRate, "minrate", "",
-		"minimal network bandwidth rate for downloading a file, in format of 20M/m/K/k")
-	flagSet.StringVar(&totalLimit, "totallimit", "",
-		"network bandwidth rate limit for the whole host, in format of 20M/m/K/k")
-	flagSet.IntVarP(&cfg.Timeout, "timeout", "e", 0,
+	flagSet.VarP(&cfg.LocalLimit, "locallimit", "s",
+		"network bandwidth rate limit for single download task, in format of G(B)/g/M(B)/m/K(B)/k/B, pure number will also be parsed as Byte")
+	flagSet.Var(&cfg.MinRate, "minrate",
+		"minimal network bandwidth rate for downloading a file, in format of G(B)/g/M(B)/m/K(B)/k/B, pure number will also be parsed as Byte")
+	flagSet.Var(&cfg.TotalLimit, "totallimit",
+		"network bandwidth rate limit for the whole host, in format of G(B)/g/M(B)/m/K(B)/k/B, pure number will also be parsed as Byte")
+	flagSet.DurationVarP(&cfg.Timeout, "timeout", "e", 0,
 		"Timeout set for file downloading task. If dfget has not finished downloading all pieces of file before --timeout, the dfget will throw an error and exit")
 
 	// md5 & identifier
@@ -255,29 +226,6 @@ func initFlags() {
 		"Alive duration for which uploader keeps no accessing by any uploading requests, after this period uploader will automatically exit")
 
 	flagSet.MarkDeprecated("exceed", "please use '--timeout' or '-e' instead")
-}
-
-// Helper functions.
-func transLimit(limit string) (int, error) {
-	if stringutils.IsEmptyStr(limit) {
-		return 0, nil
-	}
-	l := len(limit)
-	i, err := strconv.Atoi(limit[:l-1])
-
-	if err != nil {
-		return 0, err
-	}
-
-	unit := limit[l-1]
-	if unit == 'k' || unit == 'K' {
-		return i * 1024, nil
-	}
-	if unit == 'm' || unit == 'M' {
-		return i * 1024 * 1024, nil
-	}
-	return 0, fmt.Errorf("invalid unit '%c' of '%s', 'KkMm' are supported",
-		unit, limit)
 }
 
 func transFilter(filter string) []string {
--- a/dfdaemon/config/config.go
+++ b/dfdaemon/config/config.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/dragonflyoss/Dragonfly/dfdaemon/constant"
 	dferr "github.com/dragonflyoss/Dragonfly/pkg/errortypes"
+	"github.com/dragonflyoss/Dragonfly/pkg/rate"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/afero"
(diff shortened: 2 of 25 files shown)


supernode/config/config.go
```diff
@@ -22,6 +22,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/dragonflyoss/Dragonfly/pkg/model"
```
allencloud: Could you reorg the import sequence here? Thanks.

yeya24: Thanks for pointing out this. I think I have addressed all these similar nits now.

dfdaemon/config/config.go
```diff
@@ -127,30 +134,47 @@ func (p *Properties) Validate() error {
...
+	// if set supernodes in dfdaemon, dfget will use them,
+	// otherwise dfget will use the supernodes in config.
+	if len(p.SuperNodes) > 0 {
+		dfgetFlags = append(dfgetFlags, "--node", strings.Join(p.SuperNodes, ","))
+	} else if conf != nil && len(conf.Nodes) > 0 {
+		dfgetFlags = append(dfgetFlags, "--node", strings.Join(conf.Nodes, ","))
+	}
+
+	// If don't set limit in dfget config, use dfdaemon rate limit.
+	if p.RateLimit != 0 {
```
Starnop: Consider encapsulating  it as a function just like this

```
func (p *Properties) parseLimit(conf *config.Properties) []string {
	dfgetFlags := make([]string, 0)
	if conf == nil {
		if p.RateLimit != 0 {
			dfgetFlags = append(dfgetFlags, "--locallimit", p.RateLimit.String())
			dfgetFlags = append(dfgetFlags, "--totallimit", p.RateLimit.String())
			return dfgetFlags
		}
		rate := netutils.NetLimit()
		dfgetFlags = append(dfgetFlags, "--locallimit", rate.String())
		dfgetFlags = append(dfgetFlags, "--totallimit", rate.String())
		return dfgetFlags
	}

	if p.RateLimit == 0 {
		return nil
	}

	if conf.LocalLimit == 0 {
		dfgetFlags = append(dfgetFlags, "--locallimit", p.RateLimit.String())
	}
	if conf.TotalLimit == 0 {
		dfgetFlags = append(dfgetFlags, "--totallimit", p.RateLimit.String())
	}
	return dfgetFlags
}
```

dfget/core/core.go
```diff
@@ -176,6 +176,7 @@ func downloadFile(cfg *config.Config, supernodeAPI api.SupernodeAPI,
 	}
 
 	timeout := netutils.CalculateTimeout(cfg.RV.FileLength, cfg.MinRate, config.DefaultMinRate, 10*time.Second)
+	logrus.Error(timeout)
```
Starnop: Delete it or perfect it/

yeya24: This is used for debugging 🤣  I will delete it, thank you.

dfdaemon/config/config.go
yeya24: Done

cmd/dfget/app/root.go
```diff
@@ -237,8 +208,6 @@ func initFlags() {
 		"specify the size of client queue which controls the number of pieces that can be processed simultaneously")
 
 	// others
-	flagSet.BoolVarP(&cfg.ShowBar, "showbar", "b", false,
```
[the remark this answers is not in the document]
yeya24: Done

pkg/netutils/netutils.go
```diff
@@ -107,10 +109,11 @@ func NetLimit() string {
 	}
 
 	if maxInterfaceLimit > 0 {
-		return strconv.FormatUint(maxInterfaceLimit/8, 10) + "M"
+		r := model.Rate(maxInterfaceLimit)
```
Starnop: If the maxInterfaceLimit speed is `1000Mb/s`, and then you will get `800B` here.

pkg/model/rate.go
```diff
@@ -0,0 +1,137 @@
...
+
+// Type implements pflag.Value
+func (d *Rate) Type() string {
+	return "rate"
+}
+
+var rateRE = regexp.MustCompile("^([0-9]+)([mMKkBGg])$")
+
+// ParseRate parses a string into a int64.
+func ParseRate(rateStr string) (Rate, error) {
```
Starnop: I am wondering that shall we support pure digital and take `rate.B` as the unit? If not, there may be some compatibility issues.

pkg/model/rate.go
```diff
@@ -0,0 +1,137 @@
...
+	case n % int64(GB):
+		symbol = "G"
+		unit = GB
+	case n % int64(MB):
+		symbol = "M"
+		unit = MB
+	case n % int64(KB):
+		symbol = "K"
+		unit = KB
+	case n % int64(B):
```
Starnop: No need to do `mod operation` here.

pkg/model/rate.go
yeya24: Why adding `B` will cause compatibility issue? I knew that we just support K and M, so IMO adding B is OK?

pkg/model/rate.go
yeya24: Nice catch

pkg/netutils/netutils.go
yeya24: I will fix this tomorrow

inoc603:
Sorry for the delay. I actually started the review last week but forgot to click "submit review", and my reviews have been pending for a week...

pkg/model/rate.go
```diff
@@ -0,0 +1,137 @@
...
+	*d, err = ParseRate(s)
+	return err
+}
+
+// Type implements pflag.Value
+func (d *Rate) Type() string {
+	return "rate"
+}
+
+var rateRE = regexp.MustCompile("^([0-9]+)([mMKkBGg])$")
```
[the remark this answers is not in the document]
yeya24: > > By definition, rate should be something like `10GB/s`. I think we should continue to support parsing short representation like `10G`, and more accurate ones like `10MBps` or `10MB/s`. When talking about network bandwidth, the unit is very important, 'b(it)ps' and 'B(yte)ps' are both widely used. Current code use `Bps` as default unit, so when we print the rate as string, we should always be clear about it, like `10GB/s`.
> 
> @inoc603 I totally agree with you. However, in order to make it easier to be handled, I think we can make a deal with users that the rate is in seconds. And then we only support formats like KB, Kb, etc. WDYT?
> 
> @yeya24 And there is one thing you need to pay attention to that the conversion between bit and Byte.

I am sorry but I don't catch the point about the conversion between bit and byte. I think we should only support Byte here, not bit. What does this conversion mean?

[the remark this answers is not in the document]
Starnop: > I am sorry but I don't catch the point about the conversion between bit and byte. I think we should only support Byte here, not bit. What does this conversion mean?

Oops, I think what @inoc603  says is to support these forms
+ 10M
+ 10MB
+ 10MB/s
+ 10MBps

And now, let's make an agreement and we should only support the following forms
+ 10M
+ 10MB

WDYT? @inoc603 @yeya24 @lowzj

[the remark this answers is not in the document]
yeya24: What about 10m? I think in the previous version, we just support M/m

dfdaemon/config/config.go
```diff
@@ -27,7 +27,7 @@ import (
 
 	"github.com/dragonflyoss/Dragonfly/dfdaemon/constant"
 	dferr "github.com/dragonflyoss/Dragonfly/pkg/errortypes"
-
+	"github.com/dragonflyoss/Dragonfly/pkg/rate"
```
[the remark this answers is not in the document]
yeya24: Done

pkg/model/rate.go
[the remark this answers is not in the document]
Starnop: > What about 10m? I think in the previous version, we just support M/m

Oops, updated. 😄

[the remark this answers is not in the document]
yeya24: OK, I think this PR just still needs to support the unit with B, like MB, KB. Because we already support M\m now.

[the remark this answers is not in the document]
inoc603: Considering `bit`, I would suggest:
1. In parsing, reject lowercase `b`
2. In printing, always print `B`. For example, user can input `10M`, and `String()` method will always return `10MB` with `B`

cmd/dfget/app/root.go
```diff
@@ -198,13 +169,13 @@ func initFlags() {
...
 	// localLimit & minRate & totalLimit & timeout
-	flagSet.StringVarP(&localLimit, "locallimit", "s", "",
-		"network bandwidth rate limit for single download task, in format of 20M/m/K/k")
-	flagSet.StringVar(&minRate, "minrate", "",
-		"minimal network bandwidth rate for downloading a file, in format of 20M/m/K/k")
-	flagSet.StringVar(&totalLimit, "totallimit", "",
-		"network bandwidth rate limit for the whole host, in format of 20M/m/K/k")
-	flagSet.IntVarP(&cfg.Timeout, "timeout", "e", 0,
+	flagSet.VarP(&cfg.LocalLimit, "locallimit", "s",
+		"network bandwidth rate limit for single download task, in format of G/g/M/m/K/k/B")
```
lowzj: The default value of flag `locallimit` showed in help information is `0B`. Actually, `20M` is expected.

pkg/model/rate.go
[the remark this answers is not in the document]
yeya24: @inoc603 @Starnop  Maybe we don't need to support the unit with B as user input like `10MB`. We just need to support 10M\m to keep compatibility and we always output `10MB` for these units.

cmd/dfget/app/root.go
```diff
@@ -198,13 +170,13 @@ func initFlags() {
...
 
 	// localLimit & minRate & totalLimit & timeout
-	flagSet.StringVarP(&localLimit, "locallimit", "s", "",
-		"network bandwidth rate limit for single download task, in format of 20M/m/K/k")
-	flagSet.StringVar(&minRate, "minrate", "",
-		"minimal network bandwidth rate for downloading a file, in format of 20M/m/K/k")
-	flagSet.StringVar(&totalLimit, "totallimit", "",
-		"network bandwidth rate limit for the whole host, in format of 20M/m/K/k")
-	flagSet.IntVarP(&cfg.Timeout, "timeout", "e", 0,
+	flagSet.VarP(rate.NewRateValue(20*rate.MB, &cfg.LocalLimit), "locallimit", "s",
```
Starnop: It's weird. And I think you can init the config at the function `config.NewConfig()`. WDYT?

supernode/config/config.go
```diff
@@ -163,19 +164,17 @@ type BaseProperties struct {
 	// default: 5
 	FailureCountLimit int `yaml:"failureCountLimit"`
 
-	// LinkLimit is set for supernode to limit every piece download network speed (unit: MB/s).
+	// LinkLimit is set for supernode to limit every piece download network speed.
 	// default: 20
```
Starnop: Could you please update the comments by the way? THX.

Starnop:
LGTM>

yeya24:
Done. @lowzj  @Starnop

Issue #905, opened 2019-09-14
[bug]dfget will panic if we config a unexpected supernode


yunfeiyanggzq:
### Ⅱ. Describe what happened
supernode
`yunfeiyang@buaa:~/go/src/github.com/dragonflyoss/Dragonfly/cmd/supernode$ sudo go run main.go --advertise-ip 127.0.0.3  --download-port 7001 --home-dir ~/suplog1  -D
`
dfget:
```

yunfeiyang@buaa:~/go/src/github.com/dragonflyoss/Dragonfly/cmd/dfget$ for (( i=1; i<=20; i++ )); do        sudo go run main.go -u https://dl.google.com/go/go1.12.7.src.tar.gz  -o /tmp/test/b.test --notbs --expiretime 20s  --node 127.0.0.3:8002,127.0.3.0:8003; done
--2019-09-14 13:03:39--  https://dl.google.com/go/go1.12.7.src.tar.gz
dfget version:
workspace:/root/.small-dragonfly
sign:30645-1568437419.219
client:127.0.0.1 connected to node:127.0.0.3:8002
start download by dragonfly...
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x988b92]

goroutine 1 [running]:
github.com/dragonflyoss/Dragonfly/dfget/types.(*BaseResponse).IsSuccess(...)
	/home/yunfeiyang/go/src/github.com/dragonflyoss/Dragonfly/dfget/types/base_response.go:42
github.com/dragonflyoss/Dragonfly/dfget/core.reportMetrics(0xc0002e5440, 0xbc43a0, 0xc0003e6bd0, 0x3ff5e265b3c19696, 0xc000326f00, 0x40, 0x1)
	/home/yunfeiyang/go/src/github.com/dragonflyoss/Dragonfly/dfget/core/core.go:302 +0x1f2
github.com/dragonflyoss/Dragonfly/dfget/core.downloadFile(0xc0002e5440, 0xbc43a0, 0xc0003e6bd0, 0xbb31e0, 0xc0002c43c0, 0xc000079500, 0x0, 0x0)
	/home/yunfeiyang/go/src/github.com/dragonflyoss/Dragonfly/dfget/core/core.go:200 +0x6b5
github.com/dragonflyoss/Dragonfly/dfget/core.Start(0xc0002e5440, 0x4)
[...]

PR #907, opened 2019-09-14
bugfix:fix dfget panic

yunfeiyanggzq:
Signed-off-by: yunfeiyangbuaa <yunfeiyang@buaa.edu.cn>

fixes:#905

### DIFF
--- a/dfget/core/core.go
+++ b/dfget/core/core.go
@@ -299,7 +299,7 @@ func reportMetrics(cfg *config.Config, supernodeAPI api.SupernodeAPI, downloadTi
 		if err != nil {
 			logrus.Errorf("failed to report metrics to supernode %s: %v", node, err)
 		}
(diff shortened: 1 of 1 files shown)


dfget/core/core.go
```diff
@@ -299,7 +299,7 @@ func reportMetrics(cfg *config.Config, supernodeAPI api.SupernodeAPI, downloadTi
 		if err != nil {
 			logrus.Errorf("failed to report metrics to supernode %s: %v", node, err)
 		}
-		if resp.IsSuccess() {
+		if resp != nil && resp.IsSuccess() {
```
yeya24: What about moving this nil check to method resp.IsSuccess?  If resp is nil, then return false.  IMO it is more clear.

Starnop:
LGTM.

PR #914, opened 2019-09-17
fix(client): add error handling in client

hwdef:

add error handling in client

NONE

may be need test

### DIFF
--- a/client/request.go
+++ b/client/request.go
@@ -104,8 +104,12 @@ func (client *APIClient) hijack(ctx context.Context, path string, query url.Valu
 	}
 
 	if tcpConn, ok := conn.(*net.TCPConn); ok {
-		tcpConn.SetKeepAlive(true)
-		tcpConn.SetKeepAlivePeriod(30 * time.Second)
+		if err := tcpConn.SetKeepAlive(true); err != nil {
+			return nil, nil, err
+		}
+		if err := tcpConn.SetKeepAlivePeriod(30 * time.Second); err != nil {
+			return nil, nil, err
+		}
 	}
 
 	clientconn := httputil.NewClientConn(conn, nil)
--- a/client/utils.go
+++ b/client/utils.go
@@ -31,10 +31,13 @@ func decodeBody(obj interface{}, body io.Reader) error {
 	return nil
 }
 
-func ensureCloseReader(resp *Response) {
+func ensureCloseReader(resp *Response) error {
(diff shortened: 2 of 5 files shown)


client/request.go
```diff
@@ -104,8 +104,14 @@ func (client *APIClient) hijack(ctx context.Context, path string, query url.Valu
 	}
 
 	if tcpConn, ok := conn.(*net.TCPConn); ok {
-		tcpConn.SetKeepAlive(true)
-		tcpConn.SetKeepAlivePeriod(30 * time.Second)
+		err = tcpConn.SetKeepAlive(true)
```
yeya24: I hope we can use the format of 
``` Go
if err := tcpConn.SetKeepAlive(); err != nil {
    return ...
}
```
Please address these all

hwdef: @yeya24 
thanks for your suggestion.I changed it.

client/utils.go
```diff
@@ -31,10 +31,17 @@ func decodeBody(obj interface{}, body io.Reader) error {
...
 
-func ensureCloseReader(resp *Response) {
+func ensureCloseReader(resp *Response) error {
 	if resp != nil && resp.Body != nil {
 		// close body ReadCloser to make Transport reuse the connection
-		io.CopyN(ioutil.Discard, resp.Body, 512)
-		resp.Body.Close()
+
+		if _, err := io.CopyN(ioutil.Discard, resp.Body, 512); err != nil {
+			return fmt.Errorf("failed to ensure close reader: %v", err)
```
Starnop: IMOP, `return err` is enough. In addition, the caller should also handle the error that this function return.

client/utils.go
```diff
@@ -31,10 +31,15 @@ func decodeBody(obj interface{}, body io.Reader) error {
...
-func ensureCloseReader(resp *Response) {
+func ensureCloseReader(resp *Response) error {
 	if resp != nil && resp.Body != nil {
 		// close body ReadCloser to make Transport reuse the connection
-		io.CopyN(ioutil.Discard, resp.Body, 512)
-		resp.Body.Close()
+		if _, err := io.CopyN(ioutil.Discard, resp.Body, 512); err != nil {
+			return err
+		}
+		if err := resp.Body.Close(); err != nil {
```
Starnop: `resp.Body.Close()` should be executed regardless of the result of `io.CopyN`

Starnop:
LGTM.

PR #915, opened 2019-09-17
test: add err handling in test

hwdef:

add err handling in test

NONE

maybe need test

### DIFF
--- a/test/cli_dfget_p2p_test.go
+++ b/test/cli_dfget_p2p_test.go
@@ -88,7 +88,8 @@ func (s *DFGetP2PTestSuite) TestDownloadFile(c *check.C) {
 
 	for _, ca := range cases {
 		if ca.createFile {
-			s.starter.WriteSupernodeFileServer(ca.filePath, ca.fileContent, os.ModePerm)
+			err := s.starter.WriteSupernodeFileServer(ca.filePath, ca.fileContent, os.ModePerm)
+			c.Assert(err, check.IsNil)
 		}
 		cmd, err := s.starter.DFGet(ca.timeout*time.Second,
 			"-u", fmt.Sprintf("http://127.0.0.1:%d/%s", environment.SupernodeDownloadPort, ca.filePath),


test/command/cmd.go
```diff
@@ -150,12 +150,18 @@ func (s *Starter) Supernode(running time.Duration, args ...string) (
 	}
 
 	if err = check("localhost", environment.SupernodeListenPort, 10*time.Second); err != nil {
-		s.Kill(cmd)
+		err = s.Kill(cmd)
+		if err != nil {
```
yunfeiyanggzq: does this place necessary? the return is same

test/command/cmd.go
```diff
@@ -150,12 +150,18 @@ func (s *Starter) Supernode(running time.Duration, args ...string) (
...
+			return nil, err
+		}
 		return nil, err
 	}
 	s.supernodeFileServerHome = fp.Join(dir, "repo")
 	if _, err = s.fileServer(cmd, s.supernodeFileServerHome, environment.SupernodeDownloadPort); err != nil {
-		s.Kill(cmd)
+		err = s.Kill(cmd)
```
yunfeiyanggzq: The same problem as the previous one

yunfeiyanggzq:
you add some err return in function but I haven't find other functions that use this function do with the err

test/command/cmd.go
hwdef: @yunfeiyanggzq 
I think I can rename the second error,avoid overwriting value.
what do you think?
or I delete this change,keep it the same as before.

test/command/cmd.go
```diff
@@ -204,35 +214,53 @@ func (s *Starter) KillAll() {
...
 // kill can not be called directly, use Kill instead.
-func (s *Starter) kill(cmd *exec.Cmd) {
+func (s *Starter) kill(cmd *exec.Cmd) error {
 	if cmd == nil || cmd.Process == nil {
-		return
+		return nil
 	}
 	if cmd.ProcessState == nil {
-		cmd.Process.Signal(os.Interrupt)
+		err := cmd.Process.Signal(os.Interrupt)
```
Starnop:
```
if err := cmd.Process.Signal(os.Interrupt); err != nil {
	return err
}
```

Could you please use the code style as above? THX.

test/command/cmd.go
```diff
@@ -204,35 +214,53 @@ func (s *Starter) KillAll() {
 			continue
 		}
 		if v, ok := cmd.Value.(*exec.Cmd); ok {
-			s.kill(v)
+			err := s.kill(v)
```
Starnop: I think we should continue even if failed to kill some cmd. And we can return the errors with a slice. WDYT?

test/command/cmd.go
Starnop: I think it's better to update the first err. WDYT?

In addition, could you please update the scope of the err which means update the code as follows:

```
// the origin code
       if err = check("localhost", environment.SupernodeListenPort, 10*time.Second); err != nil {
		s.Kill(cmd)
		return nil, err
	}

// updated code
	if err := check("localhost", environment.SupernodeListenPort, 10*time.Second); err != nil {
		s.Kill(cmd)
		return nil, err
	}
```

hwdef:
@Starnop 
I read all your suggestions.
thanks, and I will fix it.

test/command/cmd.go
hwdef: @Starnop 
I don't understand this.
It is different with above.

test/command/cmd.go
hwdef: use a slice?
It will change a lot of code.
or keep it as same as before, do not handle this error.WDYT

Starnop: IMOP, that's OK. 😄

test/command/cmd.go
Starnop: I mean you can update your code style

from
```
err := cmd.Process.Signal(os.Interrupt)
if err != nil {
    return err
}
```

to
```
if err := cmd.Process.Signal(os.Interrupt); err != nil {
	return err
}
```

😄

hwdef:
@Starnop 
I deleted all controversial changes.

test/command/cmd.go
```diff
@@ -210,11 +210,12 @@ func (s *Starter) KillAll() {
 }
 
 // Clean cleans all temporary directories and processes.
-func (s *Starter) Clean() {
+func (s *Starter) Clean() error {
```
yeya24: Do you still need the return value `error`? I see here only return nil

hwdef: @yeya24 
You are right,I changed it.

Starnop:
LGTM.

PR #917, opened 2019-09-17
chore: remove beego submodules

Starnop:


Starnop:
> IMOP, just delete the beego, it's unnecessary to add it into `.gitignore`.

I agree with you. I do that because we have do the same thing before. FYI https://github.com/dragonflyoss/Dragonfly/pulls?q=is%3Apr+beego+is%3Aclosed

And we should pay attention to the similar updates later.

Starnop:
Has been updated. PTAL.THX

Issue #921, opened 2019-09-17
[Bug] The behavior of starting Dragonfly project is different with the document


fengzixu:
### Ⅰ. Issue Description

The actual download image behavior is inconsistent with the document description

### Ⅱ. Describe what happened

I try to use Dragonfly according to document [QuickStart](https://github.com/dragonflyoss/Dragonfly/blob/master/docs/quick_start/README.md )

When I performed the Step 5, I didn't see any output. But I found the file(/root/.small-dragonfly/logs/dfclient.log) appears in my host machine, not within the container of dfclient.

```
root@ubuntu-s-1vcpu-1gb-sgp1-01:~/.small-dragonfly# ll
total 24
drwxr-xr-x 6 root root 4096 Sep 17 14:17 ./
drwx------ 6 root root 4096 Sep 17 14:17 ../
drwxr-xr-x 2 root root 4096 Sep 17 14:17 data/
drwxr-xr-x 3 root root 4096 Sep 17 14:16 dfdaemon/
drwxr-xr-x 2 root root 4096 Sep 17 14:32 logs/
drwxr-xr-x 2 root root 4096 Sep 17 14:17 meta/
root@ubuntu-s-1vcpu-1gb-sgp1-01:~/.small-dragonfly#
```

I am so wired if the document describes wrongly or I perform wrongly.

### Ⅲ. Describe what you expected to happen

The actual behavior should be the same with the document.

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

Just perform every step in that doc.

### Ⅵ. Environment:

- dragonfly version: 0.4.3
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

yeya24:
Yes, since https://github.com/dragonflyoss/Dragonfly/pull/783, the log file will be rotated regularly. So we cannot find the output in dfclient.log. I think we should add some docs about this behavior. @inoc603

fengzixu:
@yeya24 I think we need to find an approach to show "The file has been downloaded successfully by Dragonfly", not only to describe the behavior of log roration in doc. WDYT?

yeya24:
> @yeya24 I think we need to find an approach to show "The file has been downloaded successfully by Dragonfly", not only to describe the behavior of log roration in doc. WDYT?

[...]

PR #926, opened 2019-09-18
log: not rotate log file when startup

lowzj:

fixes part of : #921

### DIFF
--- a/dfget/core/downloader/p2p_downloader/p2p_downloader.go
+++ b/dfget/core/downloader/p2p_downloader/p2p_downloader.go
@@ -188,13 +188,14 @@ func (p2p *P2PDownloader) Run() error {
 			} else if code == constants.CodePeerFinish {
 				p2p.finishTask(response, clientWriter)
 				return nil
-			} else if code == constants.CodePeerWait {
-				continue
-			}
-
-			logrus.Warnf("request piece result:%v", response)
-			if code == constants.CodeSourceError {
-				p2p.cfg.BackSourceReason = config.BackSourceReasonSourceError
+			} else {
+				logrus.Warnf("request piece result:%v", response)
+				if code == constants.CodePeerWait {
+					continue
+				}
+				if code == constants.CodeSourceError {
+					p2p.cfg.BackSourceReason = config.BackSourceReasonSourceError
+				}
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


dfget/core/downloader/p2p_downloader/p2p_downloader.go
```diff
@@ -188,13 +188,14 @@ func (p2p *P2PDownloader) Run() error {
...
 				p2p.finishTask(response, clientWriter)
 				return nil
-			} else if code == constants.CodePeerWait {
-				continue
-			}
-
-			logrus.Warnf("request piece result:%v", response)
-			if code == constants.CodeSourceError {
-				p2p.cfg.BackSourceReason = config.BackSourceReasonSourceError
+			} else {
```
Starnop: Redundant `else`.

pkg/dflog/log.go
```diff
@@ -61,7 +61,7 @@ func WithLogFile(f string) Option {
 		if logger := getLumberjack(l); logger == nil {
 			l.SetOutput(&lumberjack.Logger{
 				Filename:   f,
-				MaxSize:    20, // mb
+				MaxSize:    40, // mb
```
Starnop: make it configurable? WDYT? And so do with the `MaxBackups`.

yeya24:
What will happen if the log file exceed the max file size 40MB?

dfget/core/downloader/p2p_downloader/p2p_downloader.go
lowzj: It's not redundant. It makes the `warn log` only outputted when getting an unexpected response code.

pkg/dflog/log.go
lowzj: agree. There's already a function named `WithMaxSizeMB` to configure `MaxSize`. It's better to open a new pr to implement this especially make the supernode's log configurable.

dfget/core/downloader/p2p_downloader/p2p_downloader.go
Starnop: Yes, you are right. Sorry for not have a look at the code carefully.

Starnop:
LGTM.

Starnop:
> What will happen if the log file exceed the max file size 40MB?

FYI. https://github.com/natefinch/lumberjack

image

PR #932, opened 2019-09-18
feature: use constant value to replace literals in supernode config.

zcc35357949:

Replace literal values when initial supernode's base properties.

`FailureCountLimit` config item does not work now, fix it in this pr.

### DIFF
--- a/supernode/config/constants.go
+++ b/supernode/config/constants.go
@@ -18,6 +18,17 @@ package config
 
 import (
 	"time"
+
+	"github.com/dragonflyoss/Dragonfly/pkg/rate"
+)
+
+const (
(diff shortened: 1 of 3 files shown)


supernode/config/constants.go
```diff
@@ -100,6 +109,9 @@ const (
 )
 
 const (
+	// DefaultLinkLimit is the default network speed limit for each piece.
+	// unit: MB/s
+	DefaultLinkLimit = 20
```
Starnop: use `20*rate.MB` instead of `20`. WDYT?

zcc35357949: You are right,I will add unit.

Starnop:
LGTM.

PR #929, opened 2019-09-18
bugfix: paas registry mirror tls config into dfget config.

zcc35357949:
Signed-off-by: zhouchencheng <zhouchencheng@bilibili.com>

### Ⅰ. Describe what this PR did
The tls config in registry_mirror should save to dfget config.Otherwise, it doesn't work.

### Ⅱ. Does this pull request fix one issue?
#928

PR #942, opened 2019-09-19
feature:registry mirror support direct param.

zcc35357949:

I find that there is no config item `direct=true` in `registry_mirror` just like in `proxy`.
It is necessary for us to support this config, because in some cases we want to make some nodes pull image from remote registry directly.

### DIFF
--- a/dfdaemon/proxy/proxy.go
+++ b/dfdaemon/proxy/proxy.go
@@ -179,6 +179,7 @@ func (proxy *Proxy) mirrorRegistry(w http.ResponseWriter, r *http.Request) {
 	t, err := transport.New(
 		transport.WithDownloader(proxy.downloadFactory()),
 		transport.WithTLS(proxy.registry.TLSConfig()),
+		transport.WithCondition(proxy.shouldUseDfgetForMirror),
 	)
 	if err != nil {
 		http.Error(w, fmt.Sprintf("failed to get transport: %v", err), http.StatusInternalServerError)
@@ -262,6 +263,12 @@ func (proxy *Proxy) shouldUseDfget(req *http.Request) bool {
 		}
 	}
 	return false
+}
+
+// shouldUseDfgetForMirror returns whether we should use dfget to proxy a request
+// when we use registry mirror.
+func (proxy *Proxy) shouldUseDfgetForMirror(req *http.Request) bool {
+	return proxy.registry != nil && !proxy.registry.Direct && transport.NeedUseGetter(req)
 }
 
 // tunnelHTTPS handles a CONNECT request and proxy an https request through an
--- a/dfdaemon/config/config.go
+++ b/dfdaemon/config/config.go
@@ -52,6 +52,8 @@ var fs = afero.NewOsFs()
 //       insecure: false
 //       # optional certificates if the remote server uses self-signed certificates
 //       certs: []
+//       # whether to request the remote registry directly
+//       direct: false
 //
 //     proxies:
 //     # proxy all http image layer download requests with dfget
@@ -174,6 +176,9 @@ type RegistryMirror struct {
 
 	// Whether to ignore certificates errors for the registry
 	Insecure bool `yaml:"insecure" json:"insecure"`
+
+	// Request the remote registry directly.
+	Direct bool `yaml:"direct" json:"direct"`
 }
 
 // TLSConfig returns the tls.Config used to communicate with the mirror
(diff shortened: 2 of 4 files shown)


inoc603:
You can simply ingore `registry_mirror` and the default value, which is the official registry, will be used. see [cmd/dfdaemon/app/root.go#L92](https://github.com/dragonflyoss/Dragonfly/blob/master/cmd/dfdaemon/app/root.go#L92)

zcc35357949:
> You can simply ingore `registry_mirror` and the default value, which is the official registry, will be used. see [cmd/dfdaemon/app/root.go#L92](https://github.com/dragonflyoss/Dragonfly/blob/master/cmd/dfdaemon/app/root.go#L92)

Sorry，I‘m not sure I understand your meaning. I want to config `registry_mirror` as harbor private registry,.

inoc603:
Oh, sorry I missed your point, my bad. A `direct` field is fine, We can just use a different transport for the mirror reverse proxy if `direct = true`.

zcc35357949:
> Oh, sorry I missed your point, my bad. A `direct` field is fine, we can simply add `transport.WithCondition` here: [dfdaemon/proxy/proxy.go#L179](https://github.com/dragonflyoss/Dragonfly/blob/master/dfdaemon/proxy/proxy.go#L179)

Yes, I do add `transport.WithCondition` in `mirrorRegistry()`.
Do you mean to simplify `shouldUseDfgetForRegistryMirror`?

dfdaemon/proxy/proxy.go
```diff
@@ -179,6 +179,7 @@ func (proxy *Proxy) mirrorRegistry(w http.ResponseWriter, r *http.Request) {
 	t, err := transport.New(
 		transport.WithDownloader(proxy.downloadFactory()),
 		transport.WithTLS(proxy.registry.TLSConfig()),
+		transport.WithCondition(proxy.shouldUseDfgetForRegistryMirror),
```
inoc603: Since you've added the registry mirror to proxy rules in #929 , you only have to use a very simple condition here.
```go
transport.WithCondition(func(req *http.Request) bool {
    // TODO: handle nil
    return !proxy.registry.Direct
})
```

inoc603:
> Yes, I do add `transport.WithCondition` in `mirrorRegistry()`.
> Do you mean to simplify `shouldUseDfgetForRegistryMirror`?

@zcc35357949 I made the last quick comment from my phone and I'm sorry for not being clear enough. Please see the review.

dfdaemon/proxy/proxy.go
zcc35357949: I consider that multi-mirrors may be supported in the future, so I match the host.
We do not need that indeed currently.

allencloud:
Could we make this move on? @zcc35357949 @inoc603 @lowzj @Starnop

dfdaemon/proxy/proxy.go
inoc603: > I consider that multi-mirrors may be supported in the future, so I match the host.

How about we only support a single registry for now as a start, and add support in for multiple registries in another PR when we've decided how to handle multiple registries in #945 ?

@zcc35357949 @lowzj @Starnop

zcc35357949: > > I consider that multi-mirrors may be supported in the future, so I match the host.
> 
> How about we only support a single registry for now as a start, and add support for multiple registries in another PR when we've decided how to handle multiple registries in #945 ?
> 
> @zcc35357949 @lowzj @Starnop

ok, I have simplified the condition func.

inoc603:
@zcc35357949 Would you please add some tests for this? Thanks.

zcc35357949:
> @zcc35357949 Would you please add some tests for this? Thanks.

I have added some unit tests now.

inoc603:
@zcc35357949 The code looks fine. Would you please add the field to the comment for config.Properties [here](https://github.com/dragonflyoss/Dragonfly/blob/85e7e4b58a45de6926a9843e0b5317820f372f23/dfdaemon/config/config.go#L48), then we can merge this PR.

Starnop:
LGTM.

dfdaemon/proxy/proxy.go
```diff
@@ -264,6 +265,12 @@ func (proxy *Proxy) shouldUseDfget(req *http.Request) bool {
 	return false
 }
 
+// shouldUseDfgetForMirror returns whether we should use dfget to proxy a request
+// when we use registry mirror.
+func (proxy *Proxy) shouldUseDfgetForMirror(req *http.Request) bool {
+	return transport.NeedUseGetter(req) && proxy.registry != nil && !proxy.registry.Direct
```
lowzj: please move the `proxy.registry != nil && !proxy.registry.Direct` to the front.

PR #943, opened 2019-09-20
Using GOPROXY environment variables if set instead of default value

SataQiu:
Using GOPROXY environment variables if set instead of default value

### DIFF
--- a/hack/env.sh
+++ b/hack/env.sh
@@ -28,6 +28,3 @@ GOOS=$(go env GOOS)
 GOARCH=$(go env GOARCH)
 export GOOS
 export GOARCH
-
-export GO111MODULE=on
-export GOPROXY=https://goproxy.io
--- a/Makefile
+++ b/Makefile
@@ -16,16 +16,20 @@
 
 # You can assign 1 to USE_DOCKER so that you can build components of Dragonfly
 # with docker.
-USE_DOCKER?=0 # Default: build components in the local environment.
+USE_DOCKER ?= 0 # Default: build components in the local environment.
 
 # Assign the Dragonfly version to DF_VERSION as the image tag.
-DF_VERSION?=latest # Default: use latest as the image tag which built by docker.
+DF_VERSION ?= latest # Default: use latest as the image tag which built by docker.
 
 # Default: in order to use go mod we have to export GO111MODULE=on.
-export GO111MODULE=on
+export GO111MODULE := on
 
-# Default: use GOPROXY to speed up downloading go mod dependency.
-export GOPROXY=https://goproxy.io
+# Use GOPROXY environment variable if set
+GOPROXY := $(shell go env GOPROXY)
+ifeq ($(GOPROXY),)
+    GOPROXY := https://goproxy.io
+endif
+export GOPROXY
 
 help:  ## Display this help information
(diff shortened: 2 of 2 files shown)


Starnop:
LGTM.  cc/ @lowzj

Starnop:
Could you please rebase your code and push it again.

yeya24:
Just now I found we don' t have any docs about GOPROXY. Do you think we should add some about it? @SataQiu  @Starnop

SataQiu:
> 
> 
> Just now I found we don' t have any docs about GOPROXY. Do you think we should add some about it? @SataQiu @Starnop

That's a good idea, but what to add and where to add it?

SataQiu:
Rebased. Emmm... I wonder why rebase is needed even if there is no conflict. @Starnop

yeya24:
> > Just now I found we don' t have any docs about GOPROXY. Do you think we should add some about it? @SataQiu @Starnop
> 
> That's a good idea, but what to add and where to add it?

Maybe to CONTRIBUTING.md? The default GOPROXY value via Makefile, and you can overwrite it by setting GOPROXY

Starnop:
> Rebased. Emmm... I wonder why rebase is needed even if there is no conflict. @Starnop

FYI https://www.liaoxuefeng.com/wiki/896043488029600/1216289527823648

Starnop:
> > > Just now I found we don' t have any docs about GOPROXY. Do you think we should add some about it? @SataQiu @Starnop
> > 
> > 
> > That's a good idea, but what to add and where to add it?
> 
> Maybe to CONTRIBUTING.md? The default GOPROXY value via Makefile, and you can overwrite it by setting GOPROXY

Agree. BTW. The script `hack/env.sh` will also set values for `GO111MODULE` and `GOPROXY`. Could you please make a unified？

SataQiu:
> 
> 
> > Rebased. Emmm... I wonder why rebase is needed even if there is no conflict. @Starnop
> 
> FYI https://www.liaoxuefeng.com/wiki/896043488029600/1216289527823648

I think it would be better to use the robot to rebase automatically.

hack/env.sh
```diff
@@ -30,4 +30,8 @@ export GOOS
 export GOARCH
 
 export GO111MODULE=on
-export GOPROXY=https://goproxy.io
+
```
Starnop: I'm wondering that whether we need `GOPROXY` both are defined in the Makefile and env.sh.

Starnop:
> > > Rebased. Emmm... I wonder why rebase is needed even if there is no conflict. @Starnop
> > 
> > 
> > FYI https://www.liaoxuefeng.com/wiki/896043488029600/1216289527823648
> 
> I think it would be better to use the robot to rebase automatically.

In fact, it's difficult because it is a branch belongs to the user‘s own.

hack/env.sh
SataQiu: /cc @allencloud

SataQiu:
Maybe we can merge this now.
I don't think we need to make a special statement about this change because it's a development detail.
@allencloud @Starnop @yeya24

yeya24:
LGTM @Starnop

hack/env.sh
Starnop: Any updates? @SataQiu

Starnop:
LGTM.

PR #951, opened 2019-09-23
fix(supernode/server): add error handling

hwdef:

add error handling in test and version folder

Ref: #952

### DIFF
--- a/supernode/server/0.3_bridge.go
+++ b/supernode/server/0.3_bridge.go
@@ -256,7 +256,9 @@ func (s *Server) reportServiceDown(ctx context.Context, rw http.ResponseWriter, 
 	if err != nil {
 		return err
 	}
-	s.ProgressMgr.UpdatePeerServiceDown(ctx, dfgetTask.PeerID)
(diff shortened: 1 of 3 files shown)


Starnop:
Could you please open an issue to make a plan and track the progress for `err handling`?

WDYT?

hwdef:
Thanks for your suggestion.
I will do it .

version/version.go
```diff
@@ -125,7 +125,9 @@ func Handler(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, fmt.Sprintf("error encoding JSON: %s", err), http.StatusInternalServerError)
 	} else {
 		w.WriteHeader(http.StatusOK)
-		w.Write(data)
+		if _, err := w.Write(data); err != nil {
+			http.Error(w, fmt.Sprintf("error writing data: %s", err), http.StatusInternalServerError)
```
Starnop: In fact, http.Error will call the function of `w.Write`. And that's meaningless. So no need to handle the error here.

Starnop:
LGTM.

Issue #952, opened 2019-09-23
some function does not have error handling.


hwdef:
### What happened:
some function does not have error handling.
### What should we do:
1.add error handling for these functions in right way.
2.Don't use Fixes: # which will shut down the issue, just use Ref:#

functions list:

- [ ] client/peer_create.go:35:19:	ensureCloseReader(resp)
- [ ] client/peer_info.go:35:19:	ensureCloseReader(resp)
- [ ] client/peer_list.go:35:19:	ensureCloseReader(resp)
- [ ] client/ping.go:31:23:	defer resp.Body.Close()
- [ ] client/preheat_create.go:35:19:	ensureCloseReader(resp)
- [ ] client/preheat_info.go:35:19:	ensureCloseReader(resp)
- [ ] client/preheat_list.go:35:19:	ensureCloseReader(resp)
- [ ] client/request.go:117:24:	defer clientconn.Close()
- [ ] client/request.go:158:24:	defer resp.Body.Close()
- [ ] client/task_create.go:35:19:	ensureCloseReader(resp)
- [ ] client/task_info.go:35:19:	ensureCloseReader(resp)
- [ ] client/utils.go:36:24:	defer resp.Body.Close()
- [ ] cmd/dfdaemon/app/init.go:106:16:	filepath.Walk(dfpath, func(path string, info os.FileInfo, err error) error {
- [ ] cmd/dfdaemon/app/root_test.go:49:21:	rootCmd.Flags().Set("config", fake)
- [ ] cmd/dfdaemon/app/root_test.go:107:20:	defer os.RemoveAll(f.Name())
- [ ] cmd/dfget/app/root.go:228:24:	flagSet.MarkDeprecated("exceed", "please use '--timeout' or '-e' instead")
- [ ] cmd/dfget/app/root_test.go:53:20:	defer os.RemoveAll(dirName)
- [ ] cmd/dfget/app/root_test.go:59:18:	ioutil.WriteFile(iniFile, iniContent, os.ModePerm)
- [ ] cmd/dfget/app/root_test.go:60:18:	ioutil.WriteFile(yamlFile, yamlContent, os.ModePerm)
- [ ] cmd/dfget/app/root_test.go:83:24:	rootCmd.Flags().Parse([]string{
- [ ] cmd/supernode/app/root_test.go:75:18:	ioutil.WriteFile(s.confPath, []byte(content), os.ModePerm)
[...]

PR #964, opened 2019-09-25
 fix(dfdaemon): add error handling 

hwdef:

Add error handling in supernode/httpclient

Ref: #952

### DIFF
--- a/dfdaemon/handler/env_handler.go
+++ b/dfdaemon/handler/env_handler.go
@@ -29,7 +29,9 @@ import (
 // getEnv returns the environments of dfdaemon.
 func getEnv(w http.ResponseWriter, r *http.Request) {
 	logrus.Debugf("access:%s", r.URL.String())
-	json.NewEncoder(w).Encode(ensureStringKey(viper.AllSettings()))
+	if err := json.NewEncoder(w).Encode(ensureStringKey(viper.AllSettings())); err != nil {
+		logrus.Errorf("failed to encode env json: %v", err)
+	}
 }
 
 // ensureStringKey recursively ensures all maps in the given interface are string,
--- a/dfdaemon/handler/param_handler.go
+++ b/dfdaemon/handler/param_handler.go
@@ -31,7 +31,9 @@ func getArgs(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "text/plain;charset=utf-8")
 	for index, value := range os.Args {
 		if index > 0 {
-			w.Write([]byte(value + " "))
+			if _, err := w.Write([]byte(value + " ")); err != nil {
+				logrus.Errorf("failed to respond information: %v", err)
+			}
 		}
 
 	}
--- a/dfdaemon/proxy/proxy.go
+++ b/dfdaemon/proxy/proxy.go
@@ -330,14 +330,19 @@ func (proxy *Proxy) handleHTTPS(w http.ResponseWriter, r *http.Request) {
 	// We have to wait until the connection is closed
 	wg := sync.WaitGroup{}
 	wg.Add(1)
-	http.Serve(&singleUseListener{&customCloseConn{sConn, wg.Done}}, rp)
+	if err := http.Serve(&singleUseListener{&customCloseConn{sConn, wg.Done}}, rp); err != nil {
+		logrus.Errorf("failed to accept incoming HTTP connections: %v", err)
+	}
 	wg.Wait()
 }
 
-func copyAndClose(dst io.WriteCloser, src io.ReadCloser) {
-	io.Copy(dst, src)
-	dst.Close()
-	src.Close()
+func copyAndClose(dst io.WriteCloser, src io.ReadCloser) error {
+	defer src.Close()
+	defer dst.Close()
+	if _, err := io.Copy(dst, src); err != nil {
+		return err
+	}
+	return nil
 }
 
 func copyHeader(dst, src http.Header) {
--- a/dfdaemon/server.go
+++ b/dfdaemon/server.go
@@ -117,7 +117,7 @@ func NewFromConfig(cfg config.Properties) (*Server, error) {
 
 // Start runs dfdaemon's http server.
 func (s *Server) Start() error {
-	proxy.WithDirectHandler(handler.New())(s.proxy)
+	_ = proxy.WithDirectHandler(handler.New())(s.proxy)
 	s.server.Handler = s.proxy
 	if s.server.TLSConfig != nil {
 		logrus.Infof("start dfdaemon https server on %s", s.server.Addr)


dfdaemon/handler/env_handler.go
```diff
@@ -29,7 +29,9 @@ import (
 // getEnv returns the environments of dfdaemon.
 func getEnv(w http.ResponseWriter, r *http.Request) {
 	logrus.Debugf("access:%s", r.URL.String())
-	json.NewEncoder(w).Encode(ensureStringKey(viper.AllSettings()))
+	if err := json.NewEncoder(w).Encode(ensureStringKey(viper.AllSettings())); err != nil {
+		logrus.Errorf("failed to encode json: %v", err)
```
Starnop: Update the log to `"get Env: failed to encode json: %v", err`? WDYT?

dfdaemon/handler/param_handler.go
```diff
@@ -31,7 +31,9 @@ func getArgs(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "text/plain;charset=utf-8")
 	for index, value := range os.Args {
 		if index > 0 {
-			w.Write([]byte(value + " "))
+			if _, err := w.Write([]byte(value + " ")); err != nil {
+				logrus.Error(err)
```
Starnop: please record more info details

dfdaemon/proxy/proxy.go
```diff
@@ -330,12 +330,16 @@ func (proxy *Proxy) handleHTTPS(w http.ResponseWriter, r *http.Request) {
 	// We have to wait until the connection is closed
 	wg := sync.WaitGroup{}
 	wg.Add(1)
-	http.Serve(&singleUseListener{&customCloseConn{sConn, wg.Done}}, rp)
+	if err := http.Serve(&singleUseListener{&customCloseConn{sConn, wg.Done}}, rp); err != nil {
+		logrus.Error(err)
```
Starnop: ditto

dfdaemon/proxy/proxy.go
```diff
@@ -330,12 +330,16 @@ func (proxy *Proxy) handleHTTPS(w http.ResponseWriter, r *http.Request) {
...
+	}
 	wg.Wait()
 }
 
 func copyAndClose(dst io.WriteCloser, src io.ReadCloser) {
```
Starnop: I think `copyAndClose` is a util function which should not include output logs and it should return errors to make the caller determines how to deal with the errors.

dfdaemon/server.go
```diff
@@ -117,7 +117,9 @@ func NewFromConfig(cfg config.Properties) (*Server, error) {
 
 // Start runs dfdaemon's http server.
 func (s *Server) Start() error {
-	proxy.WithDirectHandler(handler.New())(s.proxy)
+	if err := proxy.WithDirectHandler(handler.New())(s.proxy); err != nil {
+		return err
```
Starnop: use `errors.Wrapf` to record more details

hwdef:
Thanks for you suggestions,I tried to modify it. 
It may not be very good. Do you have any suggestions?

dfdaemon/server.go
```diff
@@ -117,7 +117,9 @@ func NewFromConfig(cfg config.Properties) (*Server, error) {
 
 // Start runs dfdaemon's http server.
 func (s *Server) Start() error {
-	proxy.WithDirectHandler(handler.New())(s.proxy)
+	if err := proxy.WithDirectHandler(handler.New())(s.proxy); err != nil {
```
inoc603: [proxy.WithDirectHandler](https://github.com/dragonflyoss/Dragonfly/blob/master/dfdaemon/proxy/proxy.go#L81) always returns `nil`. Maybe we can just leave it be? Or explicitly use 
```_ = proxy.WithDirectHandler(handler.New())(s.proxy)```
 to show that we've ignored the error.

Comment (inoc603) [inline dfdaemon/server.go]: [proxy.WithDirectHandler](https://github.com/dragonflyoss/Dragonfly/blob/master/dfdaemon/proxy/proxy.go#L81) always returns nil. Maybe we can just leave it be? Or explicitly use

```
_ = proxy.WithDirectHandler(handler.New())(s.proxy)
```

to show that we've ignored the error.

File: dfdaemon/proxy/proxy.go [hunk 7]
```diff
@@ -330,14 +330,19 @@ func (proxy *Proxy) handleHTTPS(w http.ResponseWriter, r *http.Request) {
...
+		logrus.Errorf("failed to accepts incoming HTTP connections: %v", err)
+	}
 	wg.Wait()
 }
 
-func copyAndClose(dst io.WriteCloser, src io.ReadCloser) {
-	io.Copy(dst, src)
+func copyAndClose(dst io.WriteCloser, src io.ReadCloser) error {
+	if _, err := io.Copy(dst, src); err != nil {
+		return err
```
inoc603: If you return here, `Close` on both `dst` and `src` will not be called

hwdef: The change here is because of @Starnop  suggestion, maybe you can discuss it with him.

dfdaemon/server.go
hwdef: Thank you!
you are right,I will fix it.
I think proxy.WithDirectHandler always return nil is not good.should we fix it?

dfdaemon/handler/env_handler.go
```diff
@@ -29,7 +29,9 @@ import (
 // getEnv returns the environments of dfdaemon.
 func getEnv(w http.ResponseWriter, r *http.Request) {
 	logrus.Debugf("access:%s", r.URL.String())
-	json.NewEncoder(w).Encode(ensureStringKey(viper.AllSettings()))
+	if err := json.NewEncoder(w).Encode(ensureStringKey(viper.AllSettings())); err != nil {
+		logrus.Errorf("get Env: failed to encode json: %v", err)
```
allencloud: failed to encode env json: %v ?

dfdaemon/proxy/proxy.go
allencloud: @hwdef just introduce `defer to make it`. I agree with @inoc603

dfdaemon/server.go
allencloud: I am on `_ = proxy.WithDirectHandler(handler.New())(s.proxy)`. 
I agree with that we usually do not ignore returned instance when we can, but `_` could make it more clear.

dfdaemon/server.go
```diff
@@ -117,7 +117,9 @@ func NewFromConfig(cfg config.Properties) (*Server, error) {
...
+		return errors.Wrapf(err, "failed set handler: %s", s.server.Addr)
```
allencloud: failed to do something

dfdaemon/proxy/proxy.go
```diff
@@ -330,14 +330,19 @@ func (proxy *Proxy) handleHTTPS(w http.ResponseWriter, r *http.Request) {
 	// We have to wait until the connection is closed
 	wg := sync.WaitGroup{}
 	wg.Add(1)
-	http.Serve(&singleUseListener{&customCloseConn{sConn, wg.Done}}, rp)
+	if err := http.Serve(&singleUseListener{&customCloseConn{sConn, wg.Done}}, rp); err != nil {
+		logrus.Errorf("failed to accepts incoming HTTP connections: %v", err)
```
allencloud: accept

dfdaemon/handler/param_handler.go
```diff
@@ -31,7 +31,9 @@ func getArgs(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "text/plain;charset=utf-8")
 	for index, value := range os.Args {
 		if index > 0 {
-			w.Write([]byte(value + " "))
+			if _, err := w.Write([]byte(value + " ")); err != nil {
+				logrus.Errorf("failed to response information: %v", err)
```
allencloud: respond

hwdef:
@allencloud 
I read all your suggestions.
Thank you !
I fixed them.

Starnop:
LGTM.

dfdaemon/server.go
inoc603: > I think proxy.WithDirectHandler always return nil is not good.should we fix it?

The reason it returns an error is that `proxy.WithDirectHandler` is meant to be used as `proxy.Option`, which is `func (p *proxy.Proxy) error`. The nil error is only used to satisfy the function signature.

hwdef:
@Starnop 
Cloud you please merge this PR?
THX

Starnop:
LGTM.

Issue #959, opened 2019-09-24
[Help Wanted] Use viper as the configuration solution and support hot reload for config files


Starnop:
## Why you need it?

Now, we use different ways to handle the config files for `dfdaemon`, `dfget` and `supernode`. 

Let's take supernode as an example. In order to achieve that  command lines have a higher priority than the configuration file. We should do the three things as follows:
1. load config from config file
2. get cli config which specified by user explicitly
3. integrate the above two configurations which means use cli configurations overrides the configurations from config file

And everything will be simple with [viper](https://github.com/spf13/viper).

In addition, support for reloading a configuration with zero downtime is very useful especially after the project goes online. And [viper](https://github.com/spf13/viper#watching-and-re-reading-config-files) also supports the ability to have your application live read a config file while running. 

So I think we need to do following things:
- [ ] use viper to manage the configurations for `supernode`
- [ ] support hot reload config files for both `dfdaemon` and `supernode`

Starnop:
cc/ @lowzj  @inoc603 @yeya24 

Please take a review. THX. 😄

inoc603:
Agreed. `dfdaemon` is already using viper, and it should be rather easy to support config hot reloading.

lowzj:
agree

Starnop:
@SataQiu Would you like to fix this issue? 😄

SataQiu:
Yes, I‘d like to pick it up.  @Starnop 
But these two days are a bit busy, I will try to fix this issue in the next two days.

Starnop:
@SataQiu I have updated this issue. And would you like to continue to fix it? 😄

SataQiu:
Recently, I was trying to implement hot reload for `dfdaemon`.
But I'm confused about how to deal with `Config` object.
For now, we create the final `Config` object by integrating configurations from different sources(flag and file), then pass this `Config` object to the programs which need it.
It's easy to refresh `Config` object with viper, but how do we tell other components to re-read the configuration. 
I think we might need a global `Config` singleton. Instead of passing the `Config` object everywhere, we should let the program get the global `Config` object through a static method(maybe `config.GetConfig()`). We can then implement hot reload by updating the global `Config` singleton.
Also, I don't think all configuration items support hot reloading. For example, listening ports are difficult to change without restarting the program. Perhaps we should plan which configuration items should support hot reload.
Looking forward to your Suggestions! @yeya24 @inoc603 @Starnop

[...]

PR #878, opened 2019-09-03
feature: add supernode disk gc feature

Starnop:
Signed-off-by: Starnop <starnop@163.com>

### Ⅰ. Describe what this PR did
With supernode downloading more and more tasks, and it will take up more and more disk space. So we need  a gc mechanism to clean up the disk space occupied by the supenrode when necessary. And that's what this PR does.

### Ⅱ. Does this pull request fix one issue?

#867

PR #981, opened 2019-10-02
feature: enable viper for supernode

SataQiu:
feature: enable viper for supernode

Ref #959


cmd/supernode/app/root.go
```diff
@@ -21,24 +21,28 @@ import (
...
+	"github.com/spf13/viper"
+	"gopkg.in/yaml.v2"
 )
 
 var (
-	configFilePath = config.DefaultSupernodeConfigFilePath
-	cfg            = config.NewConfig()
-	options        = NewOptions()
+	defaultConfigFilePath = config.DefaultSupernodeConfigFilePath
+	defaultBaseProperties = config.NewBaseProperties()
```
inoc603: I think we can move these default value into `setupFlags` since it's not used anywhere else.

cmd/supernode/app/root.go
```diff
@@ -65,121 +122,180 @@ func setupFlags(cmd *cobra.Command, opt *Options) {
...
+	exitOnError(bindRootFlags(viper.GetViper()), "bind root command flags")
 }
 
-// runSuperNode prepares configs, setups essential details and runs supernode daemon.
-func runSuperNode() error {
-	// create home dir
-	if err := fileutils.CreateDirectory(options.HomeDir); err != nil {
-		return fmt.Errorf("failed to create home dir %s: %v", options.HomeDir, err)
+// bindRootFlags binds flags on rootCmd to the given viper instance.
+func bindRootFlags(v *viper.Viper) error {
```
inoc603: It would be better to eliminate all the redundant `if err := v.BindFlag...` using a for loop over something like this:

```go
[]struct{
    Key string
    Flag string
}{
    {"base.listenPort", "port"},
    // ...
}
```

cmd/supernode/app/root.go
```diff
@@ -65,121 +122,180 @@ func setupFlags(cmd *cobra.Command, opt *Options) {
...
+			}
+		}
+		return data, nil
+	}
 }
 
 // initLog initializes log Level and log format.
 func initLog(logger *logrus.Logger, logPath string) error {
-	logFilePath := path.Join(options.HomeDir, "logs", logPath)
+	logFilePath := path.Join(viper.GetViper().GetString("base.homeDir"), "logs", logPath)
```
inoc603: If you're always using the global viper instance, you can drop `GetViper()`.

cmd/supernode/app/root.go
SataQiu: Good idea!

SataQiu:
ping @inoc603 @yeya24 @Starnop 
PTAL!

Starnop:
also cc/@fengzixu

Starnop:
I will review this PR ASAP. Thanks for your job.

cmd/supernode/app/root.go
```diff
@@ -65,121 +121,211 @@ func setupFlags(cmd *cobra.Command, opt *Options) {
...
-		}
+// getConfigFromViper returns supernode config from the given viper instance
+func getConfigFromViper(cmd *cobra.Command, v *viper.Viper) (*config.Config, error) {
+	cfg := config.NewConfig()
+
+	if err := v.Unmarshal(cfg, func(dc *mapstructure.DecoderConfig) {
+		dc.TagName = "yaml"
+		dc.DecodeHook = decodeWithYAML(
+			reflect.TypeOf(time.Second),
+			reflect.TypeOf(rate.B),
```
Starnop: Since #878 has been merged, maybe you should make a handle for type `fileutils.Fsize`.

cmd/supernode/app/root.go
```diff
@@ -65,121 +121,211 @@ func setupFlags(cmd *cobra.Command, opt *Options) {
...
+	return nil
+}
 
-	// set supernode advertise ip
-	if stringutils.IsEmptyStr(cfg.AdvertiseIP) {
-		if err := setAdvertiseIP(); err != nil {
-			return err
-		}
+// getConfigFromViper returns supernode config from the given viper instance
+func getConfigFromViper(cmd *cobra.Command, v *viper.Viper) (*config.Config, error) {
```
Starnop: unnecessary cmd param?

cmd/supernode/app/root.go
```diff
@@ -65,121 +121,211 @@ func setupFlags(cmd *cobra.Command, opt *Options) {
...
-	flagSet.IntVar(&opt.DownloadPort, "download-port", opt.DownloadPort,
+	flagSet.Int("download-port", defaultBaseProperties.DownloadPort,
 		"downloadPort is the port for download files from supernode")
 
-	flagSet.StringVar(&opt.HomeDir, "home-dir", opt.HomeDir,
+	flagSet.String("home-dir", defaultBaseProperties.HomeDir,
 		"homeDir is the working directory of supernode")
 
-	flagSet.StringVar(&opt.DownloadPath, "download-path", opt.DownloadPath,
+	flagSet.String("download-path", defaultBaseProperties.DownloadPath,
```
Starnop: In fact, if you change the `HomeDir`, the `DownloadPath` should follow the update in default. And you can fix that [here](https://github.com/dragonflyoss/Dragonfly/blob/master/supernode/config/config.go#L96).

SataQiu: Done!

cmd/supernode/app/root.go
```diff
@@ -65,121 +122,208 @@ func setupFlags(cmd *cobra.Command, opt *Options) {
...
+		return data, nil
 	}
-
-	return d.Run()
 }
 
 // initLog initializes log Level and log format.
 func initLog(logger *logrus.Logger, logPath string) error {
-	logFilePath := path.Join(options.HomeDir, "logs", logPath)
+	logFilePath := path.Join(supernodeViper.GetString("base.homeDir"), "logs", logPath)
```
Starnop: In fact, supernode will always use the default homeDir here because it's executed before reading config file. And #990 has tried to fix this. So I will merge this PR firstly.

Starnop:
@SataQiu So sorry for the late reply. And LGTM for now. Thanks for your good job. 

And then there are something that you may need to pay attention to:
+ Could you update the first item status of issue #959?
+ If you can, please pay attention to the PR #990.
+ Would you mind to continue to fix the issue #959 and support hot reload config files?

Starnop:
LGTM>

PR #1003, opened 2019-10-15
version: add error handling

hwdef:

add error handling

Ref: #952

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -123,9 +123,10 @@ func Handler(w http.ResponseWriter, r *http.Request) {
 	data, err := json.Marshal(DFVersion)
 	if err != nil {
 		http.Error(w, fmt.Sprintf("error encoding JSON: %s", err), http.StatusInternalServerError)
+	} else if _, err := w.Write(data); err != nil {
+		http.Error(w, fmt.Sprintf("error writing the data to the connection: %s", err), http.StatusInternalServerError)
 	} else {
 		w.WriteHeader(http.StatusOK)
-		w.Write(data)
 	}
 }
 


version/version.go
```diff
@@ -125,7 +125,9 @@ func Handler(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, fmt.Sprintf("error encoding JSON: %s", err), http.StatusInternalServerError)
 	} else {
 		w.WriteHeader(http.StatusOK)
-		w.Write(data)
+		if _, err := w.Write(data); err != nil {
+			http.Error(w, fmt.Sprintf("error writing the data to the connection: %s", err), http.StatusInternalServerError)
+		}
```
allencloud: How about:
```
if _, err := w.Write(data); err != nil {
    http.Error(w, fmt.Sprintf("error writing the data to the connection: %s", err), http.StatusInternalServerError)
} else {
    w.WriteHeader(http.StatusOK)
}
```

Starnop: Agree.

hwdef: You are right,
I fixed it.

allencloud: If that, you can combine the code like:
```
if err != nil {
    http.Error(w, fmt.Sprintf("error encoding JSON: %s", err), http.StatusInternalServerError)
} else if _, err := w.Write(data); err != nil {
    http.Error(w, fmt.Sprintf("error writing the data to the connection: %s", err), http.StatusInternalServerError)
} else {
    w.WriteHeader(http.StatusOK)
}
```
And in this way, you could reduce the ident? right? WDYT

hwdef: Thank you！
You are very good at the code style of golang.
I will change it.

allencloud:
LGTM，code format is the way you talk, communicate. We are always bringing convenience when delivering code. Thanks for your update. Wish to write strict and friendly code together. @hwdef

PR #1017, opened 2019-10-24
bugfix: delete the default value for supernodes from dfdaemon

Starnop:

Somebody will define the dfget configuration by using dfget config file which locates at `/etc/dragonfly/dfget.yml`. And then they will feel confused about why dfget don't use the supernodes that I defined in the dfget config file. In fact, dfdaemon will specify a default value for supernodes and dfdaemon call the dfget by using cmd flags. So the default value defined by dfdaemon will override the dfget config. So I think we should delete the default value for supernode specified by dfdaemon.

NONE.

No need.

1. config the nodes in the dfget config file and neither define the nodes through the configuration file nor through cli flags for dfdaemon
2. start to pull a file by dfdaemon


fengzixu:
LGTM

zhujian7:
If dfdaemon and dfget neither configured the node, what will happen?

Starnop:
> If dfdaemon and dfget neither configured the node, what will happen?

Dfget will add default values `127.0.0.1` for it. 😄

PR #1019, opened 2019-10-24
feature: support set home dir in dfget config file.

zcc35357949:

Dfget server uses current user's home directory as root workdir, and it is not configurable now.
I wish that I can change the dfget work dir by dfget's config file.

None

### DIFF
--- a/cmd/dfget/app/root.go
+++ b/cmd/dfget/app/root.go
@@ -19,6 +19,7 @@ package app
 import (
 	"fmt"
 	"os"
+	"os/user"
 	"path"
 	"strings"
 	"time"
@@ -34,6 +35,12 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
+
+type propertiesResult struct {
+	prop     *config.Properties
+	fileName string
+	err      error
+}
 
 var filter string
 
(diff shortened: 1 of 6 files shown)


cmd/dfget/app/root.go
```diff
@@ -64,16 +64,16 @@ func init() {
 
 // runDfget does some init operations and starts to download.
 func runDfget() error {
+	// get config from property files
+	initProperties()
+
```
yunfeiyanggzq: you make init config before init log,I have see when init config  it  put out log by `Logrus.debugf`,so there may be something wrong.There are no place the log  put out. You'd  better make sure it

cmd/dfget/app/root.go
```diff
@@ -135,6 +155,24 @@ func initProperties() {
 	if cfg.ClientQueueSize == 0 {
 		cfg.ClientQueueSize = properties.ClientQueueSize
 	}
+
+	currentUser, err := user.Current()
+	if err != nil {
+		printer.Println(fmt.Sprintf("get user error: %s", err))
+		os.Exit(config.CodeGetUserError)
+	}
+	cfg.User = currentUser.Username
```
Starnop: In fact, `cfg.User` is not a necassary field. I will open a new PR to fix that. Make a record here.

cmd/dfget/app/root.go
```diff
@@ -64,15 +71,24 @@ func init() {
...
+	propResults := initProperties()
+
 	// initialize logger
 	if err := initClientLog(); err != nil {
 		return err
 	}
 
-	cfg.Filter = transFilter(filter)
+	// print property load result
+	for _, propRst := range propResults {
```
Starnop: It's weird. However, I have no idea to make it elegant at present. 😢  @zcc35357949 Could you please add a TODO record here in the source code? THX.

Starnop:
@zcc35357949 Could you please update your code and make it pass the CI? THX.

Starnop:
LGTM.

PR #1021, opened 2019-10-25
feature: add weight for node config

Starnop:

Now we choose the supernodes randomly and fairly. In the actual scenario, the service capabilities of multiple supernodes may be different. And we hope that the machine with good serviceability will be selected more times. So I add a weight configuration for supernode which defined by dfget. 

For now, users can specify the supernode with weight like format: `host:port/weight`. For example:
Using cli flag`dfget  --node 192.168.33.21/2 --node 127.0.0.1/1`
or using config file
```
# cat /etc/dragonfly/dfget.yml
nodes:
    - 192.168.33.21/2
    - 127.0.0.1/1 
```

And then the supernode with `192.168.33.21` and the supernode with `127.0.0.1` will be selected in a `2:1` ratio.

NONE.

The host is necessary,  port and weight is optional
Added

Take a look at dfclient.log file and search `do register to one of`

And then you will find something like this
```
do register to one of [192.168.33.21:8002 192.168.33.21:8002 127.0.0.1:8002 192.168.33.21:8002 192.168.33.21:8002 127.0.0.1:8002]
```

And the order in which you perform download or from different clients is random each time.

### DIFF
--- a/cmd/dfget/app/root.go
+++ b/cmd/dfget/app/root.go
@@ -72,7 +72,10 @@ func init() {
 // runDfget does some init operations and starts to download.
 func runDfget() error {
 	// get config from property files
-	propResults := initProperties()
+	propResults, err := initProperties()
+	if err != nil {
+		return err
+	}
 
 	// initialize logger
 	if err := initClientLog(); err != nil {
@@ -91,10 +94,6 @@ func runDfget() error {
 
 	cfg.Filter = transFilter(filter)
 
-	if err := handleNodes(); err != nil {
-		return err
-	}
-
 	if err := checkParameters(); err != nil {
 		return err
 	}
@@ -106,10 +105,10 @@ func runDfget() error {
 	logrus.Infof("get init config:%v", cfg)
 
 	// enter the core process
-	err := core.Start(cfg)
-	printer.Println(resultMsg(cfg, time.Now(), err))
-	if err != nil {
-		os.Exit(err.Code)
+	dfError := core.Start(cfg)
+	printer.Println(resultMsg(cfg, time.Now(), dfError))
+	if dfError != nil {
+		os.Exit(dfError.Code)
 	}
 	return nil
 }
@@ -122,12 +121,12 @@ func checkParameters() error {
 }
 
 // initProperties loads config from property files.
-func initProperties() []*propertiesResult {
+func initProperties() ([]*propertiesResult, error) {
 	var results []*propertiesResult
 	properties := config.NewProperties()
 	for _, v := range cfg.ConfigFiles {
-		var err error
-		if err = properties.Load(v); err == nil {
+		err := properties.Load(v)
+		if err == nil {
 			break
 		}
 		results = append(results, &propertiesResult{
@@ -137,8 +136,12 @@ func initProperties() []*propertiesResult {
 		})
 	}
 
-	if cfg.Nodes == nil {
-		cfg.Nodes = properties.Nodes
+	supernodes := cfg.Supernodes
+	if supernodes == nil {
+		supernodes = properties.Supernodes
+	}
+	if supernodes != nil {
+		cfg.Nodes = config.NodeWightSlice2StringSlice(supernodes)
 	}
 
 	if cfg.LocalLimit == 0 {
@@ -173,7 +176,7 @@ func initProperties() []*propertiesResult {
 	cfg.RV.SystemDataDir = path.Join(cfg.WorkHome, "data")
 	cfg.RV.FileLength = -1
 
-	return results
+	return results, nil
 }
 
 // initClientLog initializes dfget client's logger.
@@ -234,8 +237,8 @@ func initFlags() {
 			"\nin this way, different but actually the same URLs can reuse the same downloading task")
 	flagSet.StringSliceVar(&cfg.Header, "header", nil,
 		"http header, eg: --header='Accept: *' --header='Host: abc'")
-	flagSet.StringSliceVarP(&cfg.Nodes, "node", "n", nil,
-		"specify the addresses(host:port) of supernodes")
+	flagSet.VarP(config.NewSupernodesValue(&cfg.Supernodes, nil), "node", "n",
+		"specify the addresses(host:port=weight) of supernodes where the host is necessary, the port(default: 8002) and the weight(default:1) are optional. And the type of weight must be integer")
 	flagSet.BoolVar(&cfg.Notbs, "notbs", false,
 		"disable back source downloading for requested file when p2p fails to download it")
 	flagSet.BoolVar(&cfg.DFDaemon, "dfdaemon", false,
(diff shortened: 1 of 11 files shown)


cmd/dfget/app/root.go
```diff
@@ -234,19 +242,61 @@ func transFilter(filter string) []string {
...
+			return nil, errors.Wrapf(errortypes.ErrInvalidValue, "node: %s %v", v, err)
+		}
+
+		node, err = handleDefaultPort(node)
+		if err != nil {
+			return nil, errors.Wrapf(errortypes.ErrInvalidValue, "node: %s %v", v, err)
+		}
+
+		for i := 0; i < weight; i++ {
+			nodes = append(nodes, node)
```
yunfeiyanggzq: if  i  input --node 10.10.10.1:8002/20  10.10.10.2:8002/20 ,so they will add 40 times. and at last the dfget will get a random supernode list with length of 80,so does it reasonable

yunfeiyanggzq: so can we do with it like 20/20=1/1 or just ignore it?

docs/config/dfget_properties.md
```diff
@@ -8,7 +8,7 @@ The following startup parameters are supported for `dfget`
 
 | Parameter | Description |
 | ------------- | ------------- |
-| nodes	| Nodes specify supernodes |
+| nodes	| Nodes specify supernodes with format host:port/weight. Where the host is necessary, the port(default: 8002) and the weight(default:1) are optional. |
```
yunfeiyanggzq: `Nodes specify supernodes with format host:port/weight. Where the host is necessary, the port(default: 8002) and the weight(default:1) are optional.`  ,emmmm I think `Where` will make the gramma wrong ,haha my English is bad,But I feell `Nodes specify supernodes with format host:port/weight. The host is necessary, the port(default: 8002) and the weight(default:1) are optional.` or `Nodes specify supernodes with format host:port/weight where the host is necessary, the port(default: 8002) and the weight(default:1) are optional.` is better.

cmd/dfget/app/root.go
Starnop: I will make a consideration. Thanks for your advice.

cmd/dfget/app/root.go
```diff
@@ -275,19 +283,90 @@ func transFilter(filter string) []string {
...
+	}
+
+	return "", errortypes.ErrInvalidValue
+}
+
+func splitNodeAndWight(value string) (string, int, error) {
+	result := strings.Split(value, "/")
+	splitLength := len(result)
+
+	if splitLength < 1 || splitLength > 2 {
```
inoc603: I think a `switch` here would be more clear, since we're handling only 2 valid options.

```go
switch len(result) {
case 1:
    //
case 2:
    //
default:
    //
}
```

cmd/dfget/app/root.go
```diff
@@ -234,8 +242,8 @@ func initFlags() {
 			"\nin this way, different but actually the same URLs can reuse the same downloading task")
 	flagSet.StringSliceVar(&cfg.Header, "header", nil,
 		"http header, eg: --header='Accept: *' --header='Host: abc'")
-	flagSet.StringSliceVarP(&cfg.Nodes, "node", "n", nil,
-		"specify the addresses(host:port) of supernodes")
+	flagSet.StringSliceVarP(&cfg.Supernodes, "node", "n", nil,
+		"specify the addresses(host:port/weight) of supernodes where the host is necessary, the port(default: 8002) and the weight(default:1) are optional.")
```
inoc603: I think it's necessary to tell the user weight must be an integer.  Maybe we can also give some advice on how to set the weight, e.g. use multiples of 10

cmd/dfget/app/root.go
```diff
@@ -275,19 +283,90 @@ func transFilter(filter string) []string {
...
+	}
+
+	if splitLength == 1 {
+		return fmt.Sprintf("%s:%d", node, config.DefaultSupernodePort), nil
+	}
+
+	return "", errortypes.ErrInvalidValue
+}
+
+func splitNodeAndWight(value string) (string, int, error) {
```
allencloud: s/splitNodeAndWight/splitNodeAndWeight ?

cmd/dfget/app/root.go
```diff
@@ -275,19 +283,90 @@ func transFilter(filter string) []string {
...
+		return "", 0, errortypes.ErrInvalidValue
+	}
+	if splitLength == 2 {
+		v, err := strconv.Atoi(result[1])
+		if err != nil {
+			return "", 0, err
+		}
+		return result[0], v, nil
+	}
+	return result[0], 1, nil
```
allencloud: 1 is the default value? If that, please add more comment in the function annotation.

dfget/config/config.go
```diff
@@ -50,14 +50,14 @@ import (
...
-// 		    - 10.10.10.1
+// 		    - 127.0.0.1/1
+// 		    - 10.10.10.1:8002/2
 // 		localLimit: 20M
 // 		totalLimit: 20M
 // 		clientQueueSize: 6
 type Properties struct {
-	// Nodes specify supernodes.
-	Nodes []string `yaml:"nodes,omitempty" json:"nodes,omitempty"`
+	// Supernodes specify supernodes with weight.
```
allencloud: please add an example in code. ` ["aaaaaa:1233", "sdfghj"]`

cmd/dfget/app/root.go
```diff
@@ -234,8 +242,8 @@ func initFlags() {
 			"\nin this way, different but actually the same URLs can reuse the same downloading task")
 	flagSet.StringSliceVar(&cfg.Header, "header", nil,
 		"http header, eg: --header='Accept: *' --header='Host: abc'")
-	flagSet.StringSliceVarP(&cfg.Nodes, "node", "n", nil,
-		"specify the addresses(host:port) of supernodes")
+	flagSet.StringSliceVarP(&cfg.Supernodes, "node", "n", nil,
```
zhouhaibing089: Nice! For this specific flag type, isn't it better to introduce a new [flag.Value][1] for it?

[1]: https://golang.org/pkg/flag/#Value

Starnop:
> Nice! For this specific flag type, isn't it better to introduce a new flag.Value for it?

IMOP, it's not user-friendly if there are two flags that can configure nodes. WDYT?

> I now feel / might not be a good separator: Imagine that I configure supernode as a backend service for an ingress, it might have an address like https://foo.example.com/supernode.

Yeah, I agree with you. Thanks very much for pointing this out. And how about `*` as the 
separator? In fact, I have no good idea for this at present. 🤕

zhouhaibing089:
> IMOP, it's not user-friendly if there are two flags that can configure nodes. WDYT?

I didn't mean that. I mean, since this flag is special, I recommend create a specific flag value for it instead of keeping using stringslice. Something like below:

```go
type superNodeVal struct {
    ...
}

// implement flag.Value
func (s *superNodeVal) String() string {
    ...
}

func (s *superNodeVal) Set(string) error {

}

// then use is as:
flag.Var(&superNodeVal{}, "superNodes", ...)
```

Starnop:
request a review again.

And something that you need to know:

If you use `dfget --node 127.0.0.1=1 --node 127.0.0.2=1`, and then only the `127.0.0.2` will be applied. The last one will override all previous node values. And what's the reason that I don't combine them? In my implementation, I need to calculate the gcd(greatest common divisor) for all nodes. And it's difficult to handle by multi cli params.

PR #1036, opened 2019-10-29
pkg: add error handling

hwdef:

add error handling

ref: #952

### DIFF
--- a/pkg/digest/digest.go
+++ b/pkg/digest/digest.go
@@ -20,7 +20,9 @@ import (
 	"crypto/sha1"
 	"crypto/sha256"
 	"encoding/hex"
+	"fmt"
 	"io"
+	"os"
 )
 
 // Sha256 returns the SHA-256 checksum of the data.
@@ -34,7 +36,9 @@ func Sha256(value string) string {
 func Sha1(contents []string) string {
 	h := sha1.New()
 	for _, content := range contents {
-		io.WriteString(h, content)
+		if _, err := io.WriteString(h, content); err != nil {
+			fmt.Fprintf(os.Stderr, "failed to write string to io.Write: %v", err)
+		}
 	}
 	return hex.EncodeToString(h.Sum(nil))
 }
--- a/pkg/fileutils/fileutils.go
+++ b/pkg/fileutils/fileutils.go
@@ -60,7 +60,9 @@ func DeleteFile(filePath string) error {
 func DeleteFiles(filePaths ...string) {
 	if len(filePaths) > 0 {
 		for _, f := range filePaths {
-			DeleteFile(f)
+			if err := DeleteFile(f); err != nil {
+				fmt.Fprintf(os.Stderr, "fail to delete file: %v", err)
+			}
 		}
 	}
 }


pkg/digest/digest.go
```diff
@@ -34,7 +35,9 @@ func Sha256(value string) string {
 func Sha1(contents []string) string {
 	h := sha1.New()
 	for _, content := range contents {
-		io.WriteString(h, content)
+		if _, err := io.WriteString(h, content); err != nil {
+			fmt.Println(err)
```
allencloud: `fmt.Println(err)` would send error message to the stdout, I am wondering if we should send it to stderr.

In addition, I wish to see the principle that where indeed we should handle the returned error?

pkg/fileutils/fileutils.go
```diff
@@ -60,7 +60,9 @@ func DeleteFile(filePath string) error {
 func DeleteFiles(filePaths ...string) {
 	if len(filePaths) > 0 {
 		for _, f := range filePaths {
-			DeleteFile(f)
+			if err := DeleteFile(f); err != nil {
+				fmt.Println(err)
```
allencloud: same issue.

pkg/digest/digest.go
hwdef: I think the error in this function that should not return.
becouse this func be called in append() func.
can only output.

allencloud: I do not mean `return err`, but stream error message into stderr rather than stdout.

hwdef: I see,I fixed it.

pkg/digest/digest.go
```diff
@@ -34,7 +36,9 @@ func Sha256(value string) string {
 func Sha1(contents []string) string {
 	h := sha1.New()
 	for _, content := range contents {
-		io.WriteString(h, content)
+		if _, err := io.WriteString(h, content); err != nil {
+			fmt.Fprintf(os.Stderr, "fail to write string to io.Write %v", err)
```
allencloud: failed to write string to io.Write: %v

please use `failed` and `:`

hwdef: ok, I add them.

pkg/fileutils/fileutils.go
```diff
@@ -60,7 +60,9 @@ func DeleteFile(filePath string) error {
 func DeleteFiles(filePaths ...string) {
 	if len(filePaths) > 0 {
 		for _, f := range filePaths {
-			DeleteFile(f)
+			if err := DeleteFile(f); err != nil {
+				fmt.Fprintf(os.Stderr, "fail to delete file: %v", err)
```
yeya24: Please keep compatibility here. s/fail/failed/g .  Also, we need to log the failed file path in the error message.

pkg/digest/digest.go
```diff
@@ -34,7 +36,9 @@ func Sha256(value string) string {
 func Sha1(contents []string) string {
 	h := sha1.New()
 	for _, content := range contents {
-		io.WriteString(h, content)
+		if _, err := io.WriteString(h, content); err != nil {
+			fmt.Fprintf(os.Stderr, "failed to write string to io.Write: %v", err)
```
yeya24: @hwdef  I don' t think we need an error handling here.
If you really want to add it here, it is better to change the error description.
Maybe failed to calculate sha-256 hash?

hwdef: I think you are right.
I will close this PR.

Issue #997, opened 2019-10-12
Dfdaemon calls dfget and zombie mileage occurs


ftlynx:
## Question
docker pull 后，dfget出现了僵尸进程。采用dragonflyoss/dfclient:0.4.3 镜像部署
[root@host-192-168-55-118 logs]# ps -ef | grep dfget
root      8717 18498  0 13:58 ?        00:00:00 [dfget] <defunct>
root      8725 18498  0 13:58 ?        00:00:00 [dfget] <defunct>
root      8727 18498  0 13:58 ?        00:00:00 [dfget] <defunct>
root      8738 18498  0 13:58 ?        00:00:00 [dfget] <defunct>
root     18198 22070  0 14:20 pts/1    00:00:00 grep --color=auto dfget
root     18910 18498  0 13:02 ?        00:00:00 [dfget] <defunct>
root     18917 18498  0 13:02 ?        00:00:00 [dfget] <defunct>
root     18926 18498  0 13:02 ?        00:00:00 [dfget] <defunct>
root     18933 18498  0 13:02 ?        00:00:00 [dfget] <defunct>
root     21832 18498  0 13:10 ?        00:00:00 [dfget] <defunct>

zhouhaibing089:
https://github.com/dragonflyoss/Dragonfly/blob/e51ade6ca36f097018bf7a035aa2f7a233d65e50/dfdaemon/downloader/dfget/dfget.go#L110

This command does not have any timeout configured. We probably should call `exec.CommandContext` instead.

zcc35357949:
Dfget server process is created by dfget process by `StartPeerServerProcess`.At the beginning and ending of this function, `checkPeerServerExist` will be invoken to check whether dfget http server is ready.
But in concurrent situation dfget server may be created many times and generate many dfget server processes. Only one of them can listen on the peer port,  other processes will exit for `address already in use` before deget process finish. These processes will become defunc process.

When dfget server process does not exist, pulling a multi-layers image can reproduce the problem.
```
[...]

PR #1045, opened 2019-11-02
bugfix: avoid creating dup-process of dfget server.

zcc35357949:

Add a file lock (lock meta dir) to make sure dfget server process would not be recreated many times by dfget process before the dfserver start up.

#997

### DIFF
--- a/dfget/core/uploader/peer_server_executor.go
+++ b/dfget/core/uploader/peer_server_executor.go
@@ -21,12 +21,14 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"sync/atomic"
 	"time"
 
 	"github.com/dragonflyoss/Dragonfly/dfget/config"
+	"github.com/dragonflyoss/Dragonfly/pkg/fileutils"
 	"github.com/dragonflyoss/Dragonfly/version"
 
 	"github.com/sirupsen/logrus"
@@ -73,6 +75,12 @@ func (pe *peerServerExecutor) StartPeerServerProcess(cfg *config.Config) (port i
 	if port = pe.checkPeerServerExist(cfg, 0); port > 0 {
 		return port, nil
 	}
+
+	fileLock := fileutils.NewFileLock(filepath.Dir(cfg.RV.MetaPath))
+	if err = fileLock.Lock(); err != nil {
+		return 0, err
+	}
+	defer fileLock.Unlock()
 
 	cmd := exec.Command(os.Args[0], "server",
 		"--ip", cfg.RV.LocalIP,
--- a/pkg/fileutils/filelock.go
+++ b/pkg/fileutils/filelock.go
@@ -1,0 +1,78 @@
+/*
+ * Copyright The Dragonfly Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package fileutils
+
+import (
+	"fmt"
+	"os"
+	"syscall"
+
+	"github.com/pkg/errors"
+)
+
+// FileLock defines a file lock implemented by syscall.Flock
+type FileLock struct {
+	fileName string
+	fd       *os.File
+}
+
+// NewFileLock create a FileLock instance
+func NewFileLock(name string) *FileLock {
+	return &FileLock{
+		fileName: name,
+	}
+}
+
+// Lock locks file.
+// If the file is already locked, the calling goroutine blocks until the file is unlocked.
+// If lock has been invoked without unlock, lock again will return an error.
+func (l *FileLock) Lock() error {
+	var (
+		fd  *os.File
+		err error
+	)
+
+	if l.fd != nil {
+		return fmt.Errorf("file %s has already been locked", l.fileName)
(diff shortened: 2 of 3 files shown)


pkg/fileutils/filelock.go
```diff
@@ -0,0 +1,82 @@
...
+	l.mutex.Lock()
+	if l.fd == nil {
+		l.mutex.Unlock()
+		return fmt.Errorf("file %s descriptor is nil", l.fileName)
+	}
+	fd := l.fd
+	l.fd = nil
+	l.mutex.Unlock()
+
+	defer fd.Close()
```
jyizheng: What if fd.Close() has an error? Is it possible to return the error from it?

allencloud: no, I do not think so.

pkg/fileutils/filelock.go
```diff
@@ -0,0 +1,82 @@
...
+	if l.fd != nil {
+		l.mutex.Unlock()
+		return fmt.Errorf("invoke Unlock() before lock %s again", l.fileName)
+	}
+	l.fd = fd
+	l.mutex.Unlock()
+
+	err = syscall.Flock(int(l.fd.Fd()), syscall.LOCK_EX)
+	if err != nil {
+		return fmt.Errorf("file %s lock failed error: %s", l.fileName, err)
```
Starnop: maybe it's better to use errors.Wrapf(err,"file %s lock failed",l.fileName) 😄

dfget/core/uploader/peer_server_executor.go
```diff
@@ -70,6 +72,13 @@ type peerServerExecutor struct {
 var _ PeerServerExecutor = &peerServerExecutor{}
 
 func (pe *peerServerExecutor) StartPeerServerProcess(cfg *config.Config) (port int, err error) {
+	fileLock := fileutils.NewFileLock(filepath.Dir(cfg.RV.MetaPath))
+	err = fileLock.Lock()
```
Starnop: How about using the format as follows?
```
	if err := fileLock.Lock(); err != nil {
		return 0, err
	}
```

So do with the other codes.

pkg/fileutils/filelock.go
zcc35357949: I have wrapped the errors in lock&unlock.

pkg/fileutils/filelock.go
zhouhaibing089: Interesting. This **is** possible if it is using named return value.

```go
func (l *FileLock) Unlock() (err error) {
    ...
    defer func() {
        err = fd.Close()
    }()
    ...
}
```

But it is confusing anyway. 

1. What if `syscall.Flock` fails but `fd.Close()` succeeds?
1. What if `syscall.Flock` succeeds but `fd.Close()` fails?
1. What if both of them are failing?

dfget/core/uploader/peer_server_executor.go
```diff
@@ -70,6 +72,12 @@ type peerServerExecutor struct {
 var _ PeerServerExecutor = &peerServerExecutor{}
 
 func (pe *peerServerExecutor) StartPeerServerProcess(cfg *config.Config) (port int, err error) {
+	fileLock := fileutils.NewFileLock(filepath.Dir(cfg.RV.MetaPath))
```
zhouhaibing089: Is a simple mutex lock enough? Do you mind share more insights on why a file lock is required?

zcc35357949: Because dfget server processes are created by different dfget processes.

pkg/fileutils/filelock.go
zcc35357949: Yes, named return value can be changed in defer func indeed. Maybe we can merge the two errors in this func?  But it will look not so elegant.

Starnop: IMOP, I think we do not have to deal with all the returned errors, especially such as io.close(), file.close(), etc. And most of the similar cases in the golang source are handled in this way too.

Starnop:
LGTM.

dfget/core/uploader/peer_server_executor.go
zhouhaibing089: Oh, I see, thanks!

Starnop:
PTAL. @zhouhaibing089 @lowzj

pkg/fileutils/filelock.go
```diff
@@ -0,0 +1,85 @@
...
+
+	l.mutex.Lock()
+	if l.fd != nil {
+		l.mutex.Unlock()
+		return fmt.Errorf("invoke Unlock() before lock %s again", l.fileName)
+	}
+	l.fd = fd
+	l.mutex.Unlock()
+
+	if err := syscall.Flock(int(l.fd.Fd()), syscall.LOCK_EX); err != nil {
```
lowzj: The `l.fd` may be `nil`:

```
1. Lock:                   | Unlock:                  
2.   l.mutex.Unlock()      |   ...
3.                         |   l.mutex.Lock()
4.                         |   l.fd = nil
5.   syscall.Flock(l.fd)   |   ...
6.                         |   ...
```

I think that you can just ignore the thread-safe feature of `filelock` because that the `StartPeerServerProcess` will not be invoked in multi goroutines.

dfget/core/uploader/peer_server_executor.go
```diff
@@ -70,6 +72,12 @@ type peerServerExecutor struct {
...
+	if err = fileLock.Lock(); err != nil {
```
lowzj: It's better to run ` fileLock.Lock()` after the `pe. checkPeerServerExist()` below.

pkg/fileutils/filelock.go
```diff
@@ -0,0 +1,78 @@
...
+func (l *FileLock) Lock() error {
+	var (
+		fd  *os.File
+		err error
+	)
+	if fd, err = os.Open(l.fileName); err != nil {
+		return err
+	}
+
+	if l.fd != nil {
```
Starnop: Why not put it ahead and you should not have to open the file and close it immediately.

Starnop:
LGTM.

PR #1049, opened 2019-11-06
feature: prohibit the IPv6 to communicate with supernode

fengzixu:
Signed-off-by: fengzixu <hnustphoenix@gmail.com>

For prohibiting the IPv6, we have two methods

1. validate the IP addr which was passed by parameters in dfclient side
2. validate the IP addr in supernode side

Both methods will throw an error when users use the IPv6 address as the IP of dfget.
Before we support the IPv6 in Dragonfly, we can prohibit using IPv6 in dfclient.

### DIFF
--- a/pkg/httputils/http_util.go
+++ b/pkg/httputils/http_util.go
@@ -312,7 +312,9 @@ func CheckConnect(ip string, port int, timeout int) (localIP string, e error) {
 
 	var conn net.Conn
 	addr := fmt.Sprintf("%s:%d", ip, port)
-	if conn, e = net.DialTimeout("tcp", addr, t); e == nil {
+	// Just temporarily limit users can only use IP addr for IPv4 format.
+	// In the near future, if we want to support IPv6, we can revise logic as below.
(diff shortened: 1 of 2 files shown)


fengzixu:
Any suggestion?

fengzixu:
ping @zhouhaibing089 @Starnop

Starnop:
LGTM. also cc/ @zhouhaibing089 @lowzj

pkg/httputils/http_util.go
```diff
@@ -312,7 +312,7 @@ func CheckConnect(ip string, port int, timeout int) (localIP string, e error) {
 
 	var conn net.Conn
 	addr := fmt.Sprintf("%s:%d", ip, port)
-	if conn, e = net.DialTimeout("tcp", addr, t); e == nil {
+	if conn, e = net.DialTimeout("tcp4", addr, t); e == nil {
```
zhouhaibing089: Could you add a comment here so we are clear why it is like this?

fengzixu:
@Starnop All Done

PR #1052, opened 2019-11-06
Using path/filepath instead of path for file operations

### LINKED ISSUE #1051: distinct path&path/filepath package usage in the code
## Question
There are two packages path&path/filepath which have the same function named `Join()` in golang sdk.  I find that the usages are casual in our code. 
For example, https://github.com/dragonflyoss/Dragonfly/blob/master/cmd/dfget/app/root.go#L175
https://github.com/dragonflyoss/Dragonfly/blob/master/cmd/dfget/app/root.go#L188
...

In fact, they are different. 
- `path` package use `/` as separator. When we use paths to communicate with other programs, we should use this package to keep the consistency between different platforms, such as http path.
- `path/filepath` package depends on `os` package to get separator by operater system type. Any file operations should use this package instead of  `path`.

SataQiu:
Signed-off-by: SataQiu <1527062125@qq.com>

Using `path/filepath` instead of `path` for file operations.

fixes #1051

### DIFF
--- a/cmd/dfget/app/root.go
+++ b/cmd/dfget/app/root.go
@@ -20,7 +20,7 @@ import (
 	"fmt"
 	"os"
 	"os/user"
-	"path"
+	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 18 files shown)


Starnop:
LGTM. @SataQiu Well done. 😄

PR #1056, opened 2019-11-07
bugfix: support truncate the file when put

### LINKED ISSUE #1055: supernode fails to read .meta file with json parsing error
### Ⅰ. Issue Description
supernode log:
```
/home/admin/supernode/logs/app.log:6341:2019-11-07 20:02:16.478 INFO sign:11 : failed to update status and result fileMetaData(&{TaskID: URL: PieceSize:0 HTTPFileLen:0 Identifier: AccessTime:0 Interval:0 FileLength:6319 Md5: RealMd5:79f455139ea1e0df5837df78b1e99dfa LastModified:0 ETag: Finish:true Success:true}) for taskID(b4770705033636a548f0d9c4f404208ee0783d4184587c8503f71777f6396a36): invalid character '}' after top-level value
/home/admin/supernode/logs/app.log:6342:2019-11-07 20:02:16.478 ERRO sign:11 : failed to report cache for taskId: b4770705033636a548f0d9c4f404208ee0783d4184587c8503f71777f6396a36 : invalid character '}' after top-level value
```
I got the above errors when I pulled images.   I find that the meta file's json format is incorrect.
```
{"taskID":"b4770705033636a548f0d9c4f404208ee0783d4184587c8503f71777f6396a36","url":"https://hub.bilibili.co/v2/kube-system/metricbeat/blobs/sha256:4f3b2f97987199aa306fbe92de3e99e08a865c2c397993fae794825491523e6e","pieceSize":4194304,"httpFileLen":6314,"bizId":"","accessTime":1573128136477,"interval":485968,"fileLength":6319,"md5":"","realMd5":"79f455139ea1e0df5837df78b1e99dfa","lastModified":0,"eTag":"\"sha256:4f3b2f97987199aa306fbe92de3e99e08a865c2c397993fae794825491523e6e\"","finish":true,"success":true}}
```
There is a redundant `}` in the end. Maybe the local storage driver of supernode do not overwrite entire meta file, and the `}` is left from the last writing. 

### Ⅵ. Environment:

- dragonfly version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Starnop:

When we put data to the storage driver, it should be possible to choose whether it's erased firstly if it already exists. So I add a `trunc bool` field in this PR. And for the issue #1055, we should make  `trunc` as true.

fixes #1055

### DIFF
--- a/supernode/daemon/mgr/cdn/file_meta_data.go
+++ b/supernode/daemon/mgr/cdn/file_meta_data.go
@@ -28,6 +28,7 @@ import (
 	"github.com/dragonflyoss/Dragonfly/supernode/store"
 	"github.com/dragonflyoss/Dragonfly/supernode/util"
 
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
@@ -86,7 +87,7 @@ func (mm *fileMetaDataManager) writeFileMetaDataByTask(ctx context.Context, task
 func (mm *fileMetaDataManager) writeFileMetaData(ctx context.Context, metaData *fileMetaData) error {
 	data, err := json.Marshal(metaData)
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "failed to marshal metadata")
 	}
 
 	return mm.fileStore.PutBytes(ctx, getMetaDataRawFunc(metaData.TaskID), data)
@@ -96,14 +97,14 @@ func (mm *fileMetaDataManager) writeFileMetaData(ctx context.Context, metaData *
 func (mm *fileMetaDataManager) readFileMetaData(ctx context.Context, taskID string) (*fileMetaData, error) {
 	bytes, err := mm.fileStore.GetBytes(ctx, getMetaDataRawFunc(taskID))
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrapf(err, "failed to get metadata bytes")
 	}
 
 	metaData := &fileMetaData{}
 	if err := json.Unmarshal(bytes, metaData); err != nil {
-		return nil, err
+		return nil, errors.Wrapf(err, "failed to unmarshal metadata bytes")
 	}
-	logrus.Infof("success to read metadata: %+v for taskID: %s", metaData, taskID)
+	logrus.Debugf("success to read metadata: %+v for taskID: %s", metaData, taskID)
 
 	if metaData.PieceSize == 0 {
 		metaData.PieceSize = config.DefaultPieceSize
@@ -117,7 +118,7 @@ func (mm *fileMetaDataManager) updateAccessTime(ctx context.Context, taskID stri
 
 	originMetaData, err := mm.readFileMetaData(ctx, taskID)
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "failed to get origin metaData")
 	}
 
 	interval := accessTime - originMetaData.AccessTime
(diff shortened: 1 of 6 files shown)


supernode/store/local_storage.go
```diff
@@ -174,7 +174,12 @@ func (ls *localStorage) Put(ctx context.Context, raw *Raw, data io.Reader) error
 	lock(path, raw.Offset, false)
 	defer unLock(path, raw.Offset, false)
 
-	f, err := fileutils.OpenFile(path, os.O_WRONLY|os.O_CREATE, 0644)
+	var f *os.File
+	if raw.Trunc {
+		f, err = fileutils.OpenFile(path, os.O_WRONLY|os.O_TRUNC|os.O_CREATE, 0644)
+	} else {
+		f, err = fileutils.OpenFile(path, os.O_WRONLY|os.O_CREATE, 0644)
```
yeya24: Under what conditions the raw.Trunc bool == false? If not, we can make truncate as the default behavior instead of using a bool

supernode/daemon/mgr/cdn/file_meta_data.go
```diff
@@ -96,14 +97,14 @@ func (mm *fileMetaDataManager) writeFileMetaData(ctx context.Context, metaData *
 func (mm *fileMetaDataManager) readFileMetaData(ctx context.Context, taskID string) (*fileMetaData, error) {
 	bytes, err := mm.fileStore.GetBytes(ctx, getMetaDataRawFunc(taskID))
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrapf(err, "failed to get metadata bytes")
```
yeya24: remove "bytes"?

supernode/daemon/mgr/cdn/file_meta_data.go
```diff
@@ -96,14 +97,14 @@ func (mm *fileMetaDataManager) writeFileMetaData(ctx context.Context, metaData *
...
 	}
 
 	metaData := &fileMetaData{}
 	if err := json.Unmarshal(bytes, metaData); err != nil {
-		return nil, err
+		return nil, errors.Wrapf(err, "failed to unmarshal metadata bytes")
```
yeya24: ditto

supernode/store/local_storage.go
Starnop: In fact, it's a function that unrelated to the business code. And the value of `Truc` should be false when we want to append the file instead of truncate it firstly.

supernode/daemon/mgr/cdn/file_meta_data.go
Starnop: I don't think so. 😃

supernode/store/local_storage.go
yeya24: Do you mean that raw.Trunc is a function that returns a bool value?

zcc35357949: It is necessary to make raw.Trunc=false during putting piece data into file. Can we set this value by `raw.Offset` and `raw.Length` ?

Starnop: > Can we set this value by raw.Offset and raw.Length ?

What's that mean? Could you please provide an example?

zcc35357949: > > Can we set this value by raw.Offset and raw.Length ?
> 
> What's that mean? Could you please provide an example?

I mean that use trunc when `raw.Length`==-1 and `raw.offset`==0 in an implicit form.

jyizheng: > In fact, it's a function that unrelated to the business code. And the value of `Truc` should be false when we want to append the file instead of truncate it firstly.

Is it a good idea to also update the unit test code (local_storage_test.go) to reflect this change?

PR #1062, opened 2019-11-11
fix a comment for localStorage

jyizheng:

PR #1063, opened 2019-11-12
fix comment for localStorage

jyizheng:


Starnop:
I think this PR is repeated with #1062.

jyizheng:
> I think this PR is repeated with #1062. Cloud you please close it?

Which one should I delete, #1062 or #1063?  #1063 is to squash some old commits.

PR #1068, opened 2019-11-13
feature: try to download from back source even if download timeout with dragonfly

Starnop:

There is a probability that it will fail when dfget use `p2p` pattern to download the file from supernode. So the `dfget` should try to download from the source station again if the value of the `notbs` flag is false even if failed to download timeout with dragonfly.

NONE.

### DIFF
--- a/dfget/core/core.go
+++ b/dfget/core/core.go
@@ -168,20 +168,13 @@ func registerToSuperNode(cfg *config.Config, register regist.SupernodeRegister) 
 
 func downloadFile(cfg *config.Config, supernodeAPI api.SupernodeAPI,
 	register regist.SupernodeRegister, result *regist.RegisterResult) error {
-	var getter downloader.Downloader
-	if cfg.BackSourceReason > 0 {
-		getter = backDown.NewBackDownloader(cfg, result)
-	} else {
-		printer.Printf("start download by dragonfly...")
-		getter = p2pDown.NewP2PDownloader(cfg, supernodeAPI, register, result)
-	}
-
 	timeout := netutils.CalculateTimeout(cfg.RV.FileLength, cfg.MinRate, config.DefaultMinRate, 10*time.Second)
 	if timeout == 0 && cfg.Timeout > 0 {
 		timeout = cfg.Timeout
 	}
+
 	success := true
-	err := downloader.DoDownloadTimeout(getter, timeout)
+	err := doDownload(cfg, supernodeAPI, register, result, timeout)
 	if err != nil {
 		success = false
 	} else if cfg.RV.FileLength < 0 && fileutils.IsRegularFile(cfg.RV.RealTarget) {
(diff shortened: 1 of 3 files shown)


dfget/core/core.go
```diff
@@ -209,6 +200,40 @@ func downloadFile(cfg *config.Config, supernodeAPI api.SupernodeAPI,
...
+		isBackDownload = true
+	} else {
+		printer.Printf("start download by dragonfly...")
+		getter = p2pDown.NewP2PDownloader(cfg, supernodeAPI, register, result)
+	}
+
+	err := downloader.DoDownloadTimeout(getter, timeout)
+	if err == nil {
+		// report finished task to supernode when success to download by dragonfly
+		reportFinishedTask(cfg, getter)
```
yeya24: I think reportFinishedTask is reporting to the uploader instead of the supernode? Is this comment correct?

dfget/core/core.go
```diff
@@ -209,6 +200,40 @@ func downloadFile(cfg *config.Config, supernodeAPI api.SupernodeAPI,
...
+		return nil
+	}
+
+	if isBackDownload {
+		return fmt.Errorf("failed to download file from source station: %v", err)
```
yeya24: It is better to just say **from source**. Remove **station**.

dfget/core/core.go
```diff
@@ -209,6 +200,40 @@ func downloadFile(cfg *config.Config, supernodeAPI api.SupernodeAPI,
...
+	}
+
+	logrus.Errorf("failed to download by dragonfly: %v, and start try to download from source station", err)
+	printer.Printf("failed to download by dragonfly: %v, and start try to download from source station", err)
```
yeya24: ditto

dfget/core/core.go
```diff
@@ -209,6 +200,40 @@ func downloadFile(cfg *config.Config, supernodeAPI api.SupernodeAPI,
...
+
+	// try to download the file from the source directly
+	getter = backDown.NewBackDownloader(cfg, result)
+	if err := downloader.DoDownloadTimeout(getter, timeout); err != nil {
+		return err
```
yeya24: return fmt.Errorf("failed to download file from source: %v", err) .   Be compatible with the previous error.

dfget/core/core.go
```diff
@@ -209,6 +200,40 @@ func downloadFile(cfg *config.Config, supernodeAPI api.SupernodeAPI,
...
+	if cfg.BackSourceReason > 0 {
+		getter = backDown.NewBackDownloader(cfg, result)
+		isBackDownload = true
+	} else {
+		printer.Printf("start download by dragonfly...")
+		getter = p2pDown.NewP2PDownloader(cfg, supernodeAPI, register, result)
+	}
+
+	err := downloader.DoDownloadTimeout(getter, timeout)
+	if err == nil {
```
yeya24: if err := ...; err == nil {}

dfget/core/core.go
Starnop: Yeah, you are right.

dfget/core/core.go
Starnop: I don't think so. And the err will be used later.

PR #1073, opened 2019-11-13
bump dockerfile to golang 1.13.4 and alpine 3.10

ducksecops:
PR updates golang to latest version as well as updating to alpine 3.10

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.12.10-alpine as builder
+FROM golang:1.13.4-alpine as builder
 
 WORKDIR /go/src/github.com/dragonflyoss/Dragonfly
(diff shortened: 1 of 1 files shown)


yeya24:
@ducksecops  
Thanks for this PR. But If you want to update the golang version, it is not enough to only update it in the Dockerfile. You need to update the golang version in [circleci](https://github.com/dragonflyoss/Dragonfly/blob/master/.circleci/config.yml) and [travis](https://github.com/dragonflyoss/Dragonfly/blob/master/.travis.yml) as well. 
Btw, Dragonfly uses [pouchlinter](https://github.com/pouchcontainer/pouchlinter) as its base ci image for linting, you also need to update it.

Starnop:
@ducksecops Thanks for your contribution. And could you please provide some reasons for why you submit this PR?

PR #1085, opened 2019-11-19
refactor: handle errors gracefully

Starnop:


dfget/core/core.go
```diff
@@ -213,9 +213,9 @@ func doDownload(cfg *config.Config, supernodeAPI api.SupernodeAPI,
 	}
 
 	err := downloader.DoDownloadTimeout(getter, timeout)
+	// report finished task to uploader when success to download by dragonfly
```
lowzj: regardless of the result of downloading from dragonfly

PR #1087, opened 2019-11-20
bugfix: lock file before check server port

### LINKED ISSUE #1086: the version 1.0.0 of dfget still have zombie processes after  pull image
## phenomenon 

[root@dragonfly01 dragonfly]# docker ps 
CONTAINER ID        IMAGE                                 COMMAND                  CREATED             STATUS              PORTS                      NAMES
99b11e0b94f0        192.168.56.102/common/dfclient:1.0.0   "/opt/dragonfly/df-c…"   22 minutes ago      Up 6 minutes        0.0.0.0:65001->65001/tcp   dfclient
[root@dragonfly01 dragonfly]# ps -ef | grep dfget
root      5039  4919  0 10:58 ?        00:00:00 [dfget] <defunct>
root      5052  4919  0 10:58 ?        00:00:00 [dfget] <defunct>
root      5057  4919  0 10:58 ?        00:00:00 [dfget] <defunct>
root      5063  4919  0 10:58 ?        00:00:00 [dfget] <defunct>
root      5604  2456  0 11:05 pts/0    00:00:00 grep --color=auto dfget

zcc35357949:

Meta dir should be locked before `checkPeerServerExist`. Otherwise, any processes that wait for the lock will execute dfget server command again.

fixes #1086

### DIFF
--- a/dfget/core/uploader/peer_server_executor.go
+++ b/dfget/core/uploader/peer_server_executor.go
@@ -82,6 +82,10 @@ func (pe *peerServerExecutor) StartPeerServerProcess(cfg *config.Config) (port i
 	}
 	defer fileLock.Unlock()
 
+	if port = pe.checkPeerServerExist(cfg, 0); port > 0 {
+		return port, nil
+	}
+
 	cmd := exec.Command(os.Args[0], "server",
 		"--ip", cfg.RV.LocalIP,
 		"--port", strconv.Itoa(cfg.RV.PeerPort),


zcc35357949:
@lowzj  What do you think?

liy516:
I have compiled this code and tested it.There are no zombies when dfget pulls a new mirror.

Starnop:
LGTM.

lowzj:
> @lowzj What do you think?

I'm sorry for that, it's my mistake. You're right, it's correct to check the server port after locking. And it's better to check twice:
1. the first checking before locking: avoid executing multi-dfgets one by one at the same time if the server has already been exist
2. the second checking after locking: avoid launching multi-servers if the server doesn't exist

zcc35357949:
> > @lowzj What do you think?
> 
> I'm sorry for that, it's my mistake. You're right, it's correct to check the server port after locking. And it's better to check twice:
> 
> 1. the first checking before locking: avoid executing multi-dfgets one by one at the same time if the server has already been exist
> 2. the second checking after locking: avoid launching multi-servers if the server doesn't exist

https://github.com/dragonflyoss/Dragonfly/blob/master/dfget/core/uploader/peer_server_executor.go#L104
In fact, we have checked twice after locking now. Is it nessceary to check here?

lowzj:
> > > @lowzj What do you think?
> > 
> > 
> > I'm sorry for that, it's my mistake. You're right, it's correct to check the server port after locking. And it's better to check twice:
> > 
> > 1. the first checking before locking: avoid executing multi-dfgets one by one at the same time if the server has already been exist
> > 2. the second checking after locking: avoid launching multi-servers if the server doesn't exist
> 
> https://github.com/dragonflyoss/Dragonfly/blob/master/dfget/core/uploader/peer_server_executor.go#L104
> In fact, we have checked twice after locking now. Is it nessceary to check here?

https://github.com/dragonflyoss/Dragonfly/blob/master/dfget/core/uploader/peer_server_executor.go#L104

```
1. check to avoid serially executing multi-dfgets
2. lock and defer unlock
3. check to avoid launching multi servers
4. start server
5. check to ensure the sever is available
```

zcc35357949:
> > > > @lowzj What do you think?
> > > 
> > > 
> > > I'm sorry for that, it's my mistake. You're right, it's correct to check the server port after locking. And it's better to check twice:
> > > 
> > > 1. the first checking before locking: avoid executing multi-dfgets one by one at the same time if the server has already been exist
> > > 2. the second checking after locking: avoid launching multi-servers if the server doesn't exist
> > 
> > 
> > https://github.com/dragonflyoss/Dragonfly/blob/master/dfget/core/uploader/peer_server_executor.go#L104
> > In fact, we have checked twice after locking now. Is it nessceary to check here?
> 
> https://github.com/dragonflyoss/Dragonfly/blob/master/dfget/core/uploader/peer_server_executor.go#L104
> 
> ```
> 1. check to avoid serially executing multi-dfgets
> 2. lock and defer unlock
> 3. check to avoid launching multi servers
> 4. start server
> 5. check to ensure the sever is available
> ```

I'm also confused that whether the first checking is necessary. 
If the server has already existed, the other processes will refuse to execute multi-dfgets because they can not pass the second check.

lowzj:
Suppose that there is already a running `dfget server`, and 10 `dfget` processes are going to start using pattern `p2p` at the same time, and the function `checkPeerServerExist` needs time `T ms` to finish.

Then the expected result is that all the 10 `dfget` processes can run `checkPeerServerExist` in parallel and finish it after `T ms`. The total time cost is `T ms`. The mean waiting time is 0. That's the value of the first checking.

Otherwise the 10 `dfget` processes will run these code one by one. They have to wait for the previous `dfget` which owns the file-lock to release it. Their waiting time would be `0T ms` to `9T ms`, the total time cost is `10T ms`.  The mean waiting time is `4.5T ms`.

So the first checking is to avoid multi dfgets executing this code **one by one** when there is already a running `dfget server`.

zcc35357949:
> Suppose that there is already a running `dfget server`, and 10 `dfget` processes are going to start using pattern `p2p` at the same time, and the function `checkPeerServerExist` needs time `T ms` to finish.
> 
> Then the expected result is that all the 10 `dfget` processes can run `checkPeerServerExist` in parallel and finish it after `T ms`. The total time cost is `T ms`. The mean waiting time is 0. That's the value of the first checking.
> 
> Otherwise the 10 `dfget` processes will run these code one by one. They have to wait for the previous `dfget` which owns the file-lock to release it. Their waiting time would be `0T ms` to `9T ms`, the total time cost is `10T ms`. The mean waiting time is `4.5T ms`.
> 
> So the first checking is to avoid multi dfgets executing this code **one by one** when there is already a running `dfget server`.

Thank you for your answering.
If the server is not running and multi dfget tasks start up in parallel, the total time will cost `T` more ms. 
But I agree with you that time cost can save more if dfserver is already running. It is more common situation indeed.

PR #1088, opened 2019-11-20
optimize: reduce the waiting time for starting dfget server

lowzj:

### DIFF
--- a/dfget/core/uploader/uploader.go
+++ b/dfget/core/uploader/uploader.go
@@ -120,27 +120,51 @@ func launch(cfg *config.Config, p2pPtr *unsafe.Pointer) error {
 	return fmt.Errorf("start peer server error and retried at most %d times", retryCount)
 }
 
-func waitForStartup(result chan error, p2pPtr *unsafe.Pointer) error {
-	select {
-	case err := <-result:
-		tmp := loadSrvPtr(p2pPtr)
-		if err == nil {
-			logrus.Infof("reuse exist server on port:%d", tmp.port)
-			tmp.setFinished()
+// waitForStartup It's a goal to start 'dfget server' process and make it working
+// within 300ms, such as in the case of downloading very small files, especially
+// in parallel.
+// The ticker which has a 5ms period can test the server whether is working
+// successfully as soon as possible.
+// Actually, it costs about 70ms for 'dfget client' to start a `dfget server`
+// process if everything goes right without any failure. So the remaining time
+// for retrying to launch server internal is about 230ms. And '233' is just
+// right the smallest number which is greater than 230, a prime, and not a
+// multiple of '5'.
+// And there is only one situation which should be retried again: the address
+// already in use. The remaining time is enough for it to retry 10 times to find
+// another available address in majority of cases.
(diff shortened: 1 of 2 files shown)


dfget/core/downloader/downloader.go
```diff
@@ -42,7 +42,8 @@ type Downloader interface {
 // the given timeout duration.
 func DoDownloadTimeout(downloader Downloader, timeout time.Duration) error {
 	if timeout <= 0 {
-		logrus.Warnf("invalid download timeout(%.3fs)", timeout.Seconds())
+		logrus.Debugf("invalid download timeout(%.3fs), use default:(%.3fs)",
```
allencloud: I think it is better to use `Warnf`, since it is a useful message for operators.

dfget/core/uploader/uploader.go
```diff
@@ -120,27 +120,38 @@ func launch(cfg *config.Config, p2pPtr *unsafe.Pointer) error {
...
-		tmp := loadSrvPtr(p2pPtr)
-		if tmp == nil {
-			return fmt.Errorf("initialize peer server error")
-		}
-		if !uploaderAPI.PingServer(tmp.host, tmp.port) {
-			return fmt.Errorf("can't ping port:%d", tmp.port)
+func waitForStartup(result chan error, p2pPtr *unsafe.Pointer) (err error) {
+	ticker := time.NewTicker(5 * time.Millisecond)
+	defer ticker.Stop()
+	timeout := time.After(233 * time.Millisecond)
```
allencloud: why to use this value `233 * time.Millisecond`. How about declaring it a const value?

dfget/core/downloader/downloader.go
lowzj: Yes, it's better to use `WARN` level.
But it makes user confused because that this invalid value is set by dfget not specified by user. I will open another pr to fix it.

dfget/core/uploader/uploader.go
lowzj: It's not a multiple of `5`, is a prime number. And it's enough for retry to start a http server. 😄

Starnop: In fact, it's weird for 233ms. Maybe other contributors will feel the same way. Could you please add a comment for it? In addition, how about declaring it a const value as @allencloud said?

lowzj: > why is 233 ms, not others?

It's weird for us because of unfamiliarity, but it's really the one. There are many things in the world that are unexpected. I have already commented on this function to describe why this number jumps out by itself. It jumps into my brain, and I have to write it down.

> why not declare it as a const value

I think that maybe it's unnecessary to declare it as a const value. The code here can explain its meaning by a very simple and clear way by itself: it's the timeout of waiting for starting server. Just like the variable ticker with `5ms` above. It will lose its meaning when it appears in any other places. And the important thing is that it's very hard to come up with a name of the constant value which can explain itself without context.

zhouhaibing089: Oh, Man!

Starnop:
LGTM.

PR #1110, opened 2019-12-01
cleanup unused scripts and Makefile

yeya24:

Clean up the unused scripts and Makefile.
1. make check is not used and we already have golangci-lint check
2. vendor is not used as well

NONE


Starnop:
@yeya24  Does golangci-lint also checks the boilerplate?

Makefile
```diff
@@ -114,13 +114,6 @@ integration-test:  ## Run integration test
 	@go test ./test
 .PHONY: integration-test
 
-check: boilerplate-check  ## Check code formats
```
allencloud: Actually this is used in https://github.com/dragonflyoss/Dragonfly/blob/9017881ccebde40b1ea47b875b71c4122b13333e/.circleci/config.yml#L59.

I do not think we could delete these.

hack/check-docker.sh
```diff
@@ -1,14 +0,0 @@
-#!/bin/bash
```
allencloud: I agree with this. We do not need to check whether docker is installed. How about only deleting :
```
@echo "Begin to check dockerd whether is startup"	
./hack/check-docker.sh
```
in the file above.

allencloud:
> Does golangci-lint also checks the boilerplate?

I do not think so. @Starnop

Makefile
yeya24: I just removed the `check`, so we cannot use `make check` anymore.
But we can still use `make boilerplate-check` in ci. This does not conflict with my change.

hack/check-docker.sh
yeya24: check.sh is unnecessary as well. golangci-lint is the same of it.

allencloud:
Oh, I agree with you. @yeya24

Makefile
allencloud: I agree. My misunderstanding.

allencloud:
LGTM

PR #1121, opened 2019-12-04
fix: unset the default value of totallimit

lowzj:

The total limit of `dfget` is used to limit downloading speed of all the tasks on the same host. It's designed to be set by user explicitly, it should not be set by default. Otherwise it doesn't work if users only set flag `locallimit` but not set `totallimit`.

### DIFF
--- a/dfget/config/config.go
+++ b/dfget/config/config.go
@@ -96,7 +96,6 @@ func NewProperties() *Properties {
 		LocalLimit:      DefaultLocalLimit,
 		MinRate:         DefaultMinRate,
 		ClientQueueSize: DefaultClientQueueSize,
-		TotalLimit:      DefaultTotalLimit,
 	}
 }
 
--- a/cmd/dfget/app/root_test.go
+++ b/cmd/dfget/app/root_test.go
@@ -39,7 +39,7 @@ func (suit *dfgetSuit) Test_initFlagsNoArguments() {
 	initProperties()
 	suit.Equal(cfg.Nodes, []string{"127.0.0.1:8002"})
 	suit.Equal(cfg.LocalLimit, 20*rate.MB)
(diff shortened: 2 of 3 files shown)


Starnop:
Could you please add more descriptions about why to submit this PR?

dfget/config/config.go
```diff
@@ -96,7 +96,6 @@ func NewProperties() *Properties {
 		LocalLimit:      DefaultLocalLimit,
 		MinRate:         DefaultMinRate,
 		ClientQueueSize: DefaultClientQueueSize,
-		TotalLimit:      DefaultTotalLimit,
```
Starnop: Then could you please delete the const value `DefaultTotalLimit`?

Starnop:
> The total limit is used to limit downloading speed of all the tasks on the same host. It's designed to be set by user explicitly, it should not be set by default. Otherwise it doesn't work if users only set flag locallimit but not set totallimit.

In fact, dfdaemon will set a default totalLimit value for dfget. Should it be deleted?

lowzj:
> > The total limit is used to limit downloading speed of all the tasks on the same host. It's designed to be set by user explicitly, it should not be set by default. Otherwise it doesn't work if users only set flag locallimit but not set totallimit.
> 
> In fact, dfdaemon will set a default totalLimit value for dfget. Should it be deleted?

It's unnecessary to delete it from dfdaemon. The `dfdaemon` is a specific usage scenario of `dfget`, it will start serval `dfget` processes at the same time when pulling an image, so the `totallimit` should be set. And the flag `ratelimit` of `dfdaemon` will set the `totallimit` and `locallimit` of `dfget` at the same time.

Starnop:
LGTM.

PR #1122, opened 2019-12-04
feature: dfdaemon supports hot reload

SataQiu:
Signed-off-by: SataQiu <1527062125@qq.com>

feature: dfdaemon supports hot reload

### DIFF
--- a/cmd/dfdaemon/app/root.go
+++ b/cmd/dfdaemon/app/root.go
@@ -17,7 +17,9 @@
 package app
 
 import (
+	"context"
 	"encoding/json"
+	"net/http"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -32,6 +34,7 @@ import (
 	"github.com/dragonflyoss/Dragonfly/pkg/netutils"
 	"github.com/dragonflyoss/Dragonfly/pkg/rate"
 
+	"github.com/fsnotify/fsnotify"
 	"github.com/mitchellh/mapstructure"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


Starnop:
Could you please add more descriptions about why to submit this PR?

cmd/dfdaemon/main.go
```diff
@@ -17,9 +17,27 @@
...
 	"github.com/dragonflyoss/Dragonfly/cmd/dfdaemon/app"
+	dferr "github.com/dragonflyoss/Dragonfly/pkg/errortypes"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
 func main() {
-	app.Execute()
+	rand.Seed(time.Now().UnixNano())
```
Starnop: What is the meaning of this line of code?

cmd/dfdaemon/main.go
```diff
@@ -17,9 +17,27 @@
...
+
+	command := app.NewDFDaemonCommand()
+
```
Starnop: delete this blank line?

pkg/cmd/version.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/dragonflyoss/Dragonfly/version"
+
+	"github.com/spf13/cobra"
+)
+
+// versionLongTempl is used to describe version command in detail and auto generate command doc.
+var versionLongTempl = "Display the version and build information of Dragonfly %s, " +
+	"including GoVersion, OS, Arch, Version, BuildDate and GitCommit."
+
+func NewVersionCommand(name string) *cobra.Command {
```
allencloud: Please do not forget to add the `Example` field. And it is be used to generate docs.

cmd/dfdaemon/app/version.go
```diff
@@ -1,56 +0,0 @@
...
-	Use:           "version",
-	Short:         "Show the current version of dfdaemon",
-	Long:          versionDescription,
-	SilenceErrors: true,
-	SilenceUsage:  true,
-	RunE: func(cmd *cobra.Command, args []string) error {
-		printer.Printf(version.Print("dfdaemon"))
-		return nil
-	},
-	Example: versionExample(),
```
allencloud: see here.

inoc603:
I think the `options` package may not be very necessary. It's only used by the `dfdaemon` command.  It has a single file with only 2 functions, and the filename `generic.go` suggests that the purpose of that file is kind of unclear. Setting flags and parsing command line arguments are tightly coupled with the command itself, so it might be better to keep them where they're used.

cmd/dfdaemon/main.go
zhouhaibing089: See https://golang.org/pkg/math/rand/#Rand.Seed.

SataQiu:
Hi @Starnop What do you think about this PR?

SataQiu:
> very nice work! one suggestion: `dfdaemon` should check the changed config file before stopping, it cannot stop if the new config is invalid.

Done!

SataQiu:
I'm going to close this PR now.
Maybe we can do what kube-proxy has done: exit the program after detecting a configuration change and let k8s pull up the program again automatically to apply the new configuration.
Ref: https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-proxy/app/server.go#L250-L275

/close

PR #1145, opened 2019-12-27
Configure dfclient/dfdaemon/dfget log file path via 'logConfig.path' property

YanzheL:
Signed-off-by: YanzheL lee.yanzhe@yanzhe.org

Currently, there is no option for user to customize the log file path, which causes inconvenience while using `dfclient` in container.

For a typical use case, it is expected that the dfclient container logs to stdout so we can inspect the logs via `docker logs -f <dfclient>` instead of checking the log file in default location `~/.small-dragonfly/logs/dfdaemon.log`

This PR enables user to configure the log file path via configuration property `logConfig.path`.

For example
```yaml
logConfig:
  path: /dev/stdout
```

NONE

Already tested in container.

1. Build the dockerfile

2. Add this property in dfclient configuration file

   ```yaml
   logConfig:
     path: /dev/stdout
   ```

3. You will see the log in `stdout` instead of the default location `~/.small-dragonfly/logs/dfdaemon.log`

### DIFF
--- a/pkg/dflog/log.go
+++ b/pkg/dflog/log.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -35,6 +36,10 @@ type LogConfig struct {
 	// MaxBackups is the maximum number of old log files to retain.
 	// The default value is 1.
 	MaxBackups int `yaml:"maxBackups" json:"maxBackups"`
+
+	// Path is the location of log file
+	// The default value is logs/dfdaemon.log
+	Path string `yaml:"path" json:"path"`
 }
 
 // DefaultLogTimeFormat defines the timestamp format.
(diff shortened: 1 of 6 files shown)


pkg/dflog/log.go
```diff
@@ -72,7 +77,7 @@ func getLumberjack(l *logrus.Logger) *lumberjack.Logger {
 func WithLogFile(f string, maxSize, maxBackups int) Option {
 	return func(l *logrus.Logger) error {
 		if f == "" {
-			return nil
+			f = filepath.Join("logs", "dfdaemon.log")
```
Starnop: In fact, not only dfdaemon will use this function, so do with `dfget` and `supenrode`.

lowzj:
@YanzheL thanks for your contributing very much! Could you please rebase the code and fix the conflict and reviewed comment, then I will merge this pr.

pkg/dflog/log.go
YanzheL: Ok, I rollbacked the change of this line, and set the default log path for dfget/dfdaemon in caller (See my new commits).

As for supernode, I didn't touch its code. Because it [writes to two different log files](https://github.com/dragonflyoss/Dragonfly/blob/def2b217d4faab07dd9fa03e527d3dc3cc106b8a/cmd/supernode/app/root.go#L84-L92), a single `logConfig.path` property is not enough to configure the location of these files.

For now, this PR configures the log path of dfget/dfdaemon, not for supernode.

YanzheL:
> @YanzheL thanks for your contributing very much! Could you please rebase the code and fix the conflict and reviewed comment, then I will merge this pr.

I updated and rebased the PR, thanks.

PR #1152, opened 2019-12-31
bugfix: update cdn fail when the content is being read and the source server is down

### LINKED ISSUE #1150: In case of data source exception, the task in progress on supernode will not be terminated automatically
1. Dfget downloads files through supernode

// 1、dfget通过supernode下载文件 

Expected results:

// 期望结果：

***!!!!WE STRONGLY ENCOURAGE YOU TO DESCRIBE YOUR ISSUE IN ENGLISH!!!!***

Starnop:

If supernode is writing to the CDN cache and the source server goes down, the cdn status of task will not be updated failed for now. This PR will fix it.

In addition, if dfget failed to download from `supernode xxx`, it will migrate to other supenodes in the node list of dfget. It's unreasonable to register to the supernode  that just failed, and it will bring a few questions:
+ It doesn't make much sense to try again for the same supernode.
+ The cid of dfgetTask will not change. So two dfgetTask with same cid will be registered on the same supernode, which will cause some data confusion.

fixes #1150

**Attention**
In the meantime, another problem has also been found.

```
dragonfly_supernode_dfgettasks{callsystem="",status="RUNNING"} 1
dragonfly_supernode_dfgettasks{callsystem="",status="WAITING"} 0
```
 When dfget migrate to the other supernodes, the current supernode doesn't know that. So the state of dfgettask will keep running. I will start a new PR to do that.

### DIFF
--- a/dfget/core/regist/register.go
+++ b/dfget/core/regist/register.go
@@ -40,8 +40,9 @@ type SupernodeRegister interface {
 }
 
 type supernodeRegister struct {
-	api api.SupernodeAPI
-	cfg *config.Config
+	api                api.SupernodeAPI
+	cfg                *config.Config
+	lastRegisteredNode string
 }
 
 var _ SupernodeRegister = &supernodeRegister{}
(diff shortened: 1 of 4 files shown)


Starnop:
@lowzj PTAL. THX.

dfget/core/regist/register.go
```diff
@@ -68,6 +69,10 @@ func (s *supernodeRegister) Register(peerPort int) (*RegisterResult, *errortypes
 	nodes, nLen := s.cfg.Nodes, len(s.cfg.Nodes)
 	req := s.constructRegisterRequest(peerPort)
 	for i = 0; i < nLen; i++ {
+		if s.lastRegisteredNode == nodes[i] {
+			logrus.Warnf("the last registered node is the same(%s)", nodes[i])
+			continue
+		}
```
lowzj: What's the reason for this? How and when to retry  a supernode?

Starnop: > In addition, if dfget failed to download from supernode xxx, it will migrate to other supenodes in the node list of dfget. It's unreasonable to register to the supernode that just failed, and it will bring a few questions:

> It doesn't make much sense to try again for the same supernode.
The cid of dfgetTask will not change. So two dfgetTask with same cid will be registered on the same supernode, which will cause some data confusion.

PR #1166, opened 2020-01-09
fix: use StringArrayVar to parse http header

xujihui1985:
StringSliceVar will separated by comma, which is not suitable for field
like Date:Mon, 30 Dec 2019, use StringArrayVar instead


cmd/dfget/app/root_test.go
```diff
@@ -24,11 +24,12 @@ import (
 	"testing"
 	"time"
 
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/suite"
+
```
lowzj: imports sequence:
```go
import (
    stdlib imports

    dragonfly imports

    third-party import
)
```

xujihui1985: modified as requested

PR #1185, opened 2020-01-15
bugfix: retry multi times if failed to report pieces

Starnop:

Sometimes dfget will fail to report pieces to supernode for some reason, such as supernode cannot handle it. And then supernode will mark that this task has not been completed forever. That's not expected. So we should add a retry mechanism to reduce the chance of errors.

### DIFF
--- a/dfget/core/downloader/p2p_downloader/client_writer.go
+++ b/dfget/core/downloader/p2p_downloader/client_writer.go
@@ -20,6 +20,7 @@ import (
 	"bufio"
 	"context"
 	"io"
+	"math/rand"
 	"os"
 	"time"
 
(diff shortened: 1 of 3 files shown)


dfget/core/downloader/p2p_downloader/client_writer.go
```diff
@@ -202,12 +203,34 @@ func startSyncWriter(q queue.Queue) queue.Queue {
...
+	reportPieceRequest := &types.ReportPieceRequest{
 		TaskID:     piece.TaskID,
 		Cid:        cw.cfg.RV.Cid,
 		DstCid:     piece.DstCid,
 		PieceRange: piece.Range,
-	})
+	}
+
+	var retry = 0
+	var maxRetryTime = 3
```
lowzj: The following codes will send request to supernode at most 4 times, is it expected?

dfget/core/downloader/p2p_downloader/client_writer.go
```diff
@@ -202,12 +203,34 @@ func startSyncWriter(q queue.Queue) queue.Queue {
...
+				logrus.Warnf("success to report piece with request(%+v) after retrying (%d) times", reportPieceRequest, retry)
+			}
+			break
+		}
+		if retry >= maxRetryTime {
+			logrus.Errorf("failed to report piece to supernode with request(%+v) even after retrying max retry time", reportPieceRequest)
+			break
+		}
+
+		sleepTime := time.Duration(rand.Intn(1000)+500) * time.Millisecond
```
lowzj: The time to sleep is a bit long. It's better to decrease the interval time of first retry. How about [50, 500]ms?

dfget/core/downloader/p2p_downloader/client_writer.go
Starnop: I don't think so. I use `>=` here `if retry >= maxRetryTime`.

dfget/core/downloader/p2p_downloader/client_writer.go
Starnop: DONE.

dfget/core/downloader/p2p_downloader/client_writer.go
Starnop: Updated.

PR #1213, opened 2020-02-11
feature: support source cdn pattern

Starnop:

As present, supernode will trigger CDN module to download the file from the source if not exist. However, that will cost some time. It's now allowed when someone wants to use the ability of  P2P distribution without caching. 

## Restrictions
1. The source URL support for getting `ContentLength`.
2. The source URL support for range requests.
3. MD5 validation is not supported. 

NONE.

NONE.


Starnop:
> The function `dfget/core/uploader/peer_server#amendRange` should not compute the piece meta size to support `cdn-source` pattern.

DONE. PTAL again.

PR #1221, opened 2020-02-14
add unit test for NodeWightSlice2StringSlice

q384566678:
Signed-off-by: Zhou Hao <zhouhao@cn.fujitsu.com>

### DIFF
--- a/dfget/config/supernode_value_test.go
+++ b/dfget/config/supernode_value_test.go
@@ -98,3 +98,14 @@ func (suit *SupernodeValueSuite) TestHandleNodes() {
 		}
 	}
 }
+
+func (suit *SupernodeValueSuite) TestNodeWeightSlice2StringSlice() {
+	var nodeWithWeight = []*NodeWeight{
+		{"127.0.0.1:8002", 0},
+		{"127.0.0.2:8003", 1},
+		{"127.0.0.2:8004", 2},
+	}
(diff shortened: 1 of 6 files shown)


dfget/config/supernode_value_test.go
```diff
@@ -98,3 +98,14 @@ func (suit *SupernodeValueSuite) TestHandleNodes() {
 		}
 	}
 }
+
+func (suit *SupernodeValueSuite) TestNodeWightSlice2StringSlice() {
```
allencloud: I think the `Wight` should be the word `Weight`, do you think so ? @Starnop @lowzj 
If that, would change all the `Wight` into `Weight`in file https://github.com/dragonflyoss/Dragonfly/blob/b32d1ca03b4111978b2d8a0f398704946e071d99/dfget/config/supernode_value.go and some other files?

q384566678: I think this should be a typo, it should be Weight, not Wight.

allencloud: Could you add a pull request to change them all?

q384566678: Done, PTAL.

allencloud: Thank you so much.

allencloud:
LGTM

PR #1223, opened 2020-02-14
pkg/fileutils: add verifications and unit test for SymboliLlink

littleroad:

### DIFF
--- a/pkg/fileutils/fileutils.go
+++ b/pkg/fileutils/fileutils.go
@@ -94,7 +94,18 @@ func Link(src string, linkName string) error {
 
 // SymbolicLink creates target as a symbolic link to src.
 func SymbolicLink(src string, target string) error {
-	// TODO Add verifications.
+	if !PathExist(src) {
+		return fmt.Errorf("failed to symlink %s to %s: src no such file or directory", target, src)
+	}
+	if PathExist(target) {
+		if IsDir(target) {
+			return fmt.Errorf("failed to symlink %s to %s: link name already exists and is a directory", target, src)
+		}
+		if err := DeleteFile(target); err != nil {
+			return fmt.Errorf("failed to symlink %s to %s when deleting target file: %v", target, src, err)
+		}
+
+	}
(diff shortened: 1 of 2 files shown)


pkg/fileutils/fileutils.go
```diff
@@ -94,7 +94,18 @@ func Link(src string, linkName string) error {
 
 // SymbolicLink creates target as a symbolic link to src.
 func SymbolicLink(src string, target string) error {
-	// TODO Add verifications.
+	if !PathExist(src) {
+		return fmt.Errorf("Symlink %s to %s: error, src no such file or directory", target, src)
```
allencloud: when constructing an error, we usually use `failed to do sth: <detailed failing reason>`, in this case:
```
fmt.Errorf("failed to symlink %s to %s: src no such file or directory", target, src)
```

And please follow this rule in other places, thx.

pkg/fileutils/fileutils.go
```diff
@@ -94,7 +94,18 @@ func Link(src string, linkName string) error {
...
+	}
+	if PathExist(target) {
+		if IsDir(target) {
+			return fmt.Errorf("Symlink %s to %s: error, link name already exists and is a directory", target, src)
+		}
+		if err := DeleteFile(target); err != nil {
+			return err
```
allencloud: why not to add more details in the error message rather than directly return the error? 
for example:
```
return fmt.Errorf("failed to symlink %s to %s when deleting target file: %v", target, src, err)
```

pkg/fileutils/fileutils.go
littleroad: Understood it. Thanks.

pkg/fileutils/fileutils.go
littleroad: Done.

pkg/fileutils/fileutils.go
littleroad: Done with #1225

allencloud:
LGTM

PR #1225, opened 2020-02-17
Construct errors following conventions

littleroad:
According to [the suggestions given by Allen](https://github.com/dragonflyoss/Dragonfly/pull/1223#discussion_r379912052), reconstruct all errors in pkg/fileutils following conventions.

### DIFF
--- a/pkg/fileutils/fileutils.go
+++ b/pkg/fileutils/fileutils.go
@@ -36,11 +36,14 @@ const BufferSize = 8 * 1024 * 1024
 // CreateDirectory creates directory recursively.
 func CreateDirectory(dirPath string) error {
 	f, e := os.Stat(dirPath)
-	if e != nil && os.IsNotExist(e) {
-		return os.MkdirAll(dirPath, 0755)
+	if e != nil {
+		if os.IsNotExist(e) {
+			return os.MkdirAll(dirPath, 0755)
+		}
+		return fmt.Errorf("failed to create dir %s: %v", dirPath, e)
 	}
-	if e == nil && !f.IsDir() {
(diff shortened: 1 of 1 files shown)


pkg/fileutils/fileutils.go
```diff
@@ -105,33 +108,33 @@ func CopyFile(src string, dst string) (err error) {
...
 	defer s.Close()
 
 	if PathExist(dst) {
-		return fmt.Errorf("copy file:%s error, dst file already exists", dst)
+		return fmt.Errorf("failed to copy %s to %s: dst file already exists", src, dst)
 	}
 
 	if d, err = OpenFile(dst, os.O_RDWR|os.O_TRUNC|os.O_CREATE, 0755); err != nil {
-		return err
+		return fmt.Errorf("failed to copy %s to %s when opening dst file: %v", src, dst, err)
```
allencloud: Please make this distinguished from the Line 114 of this file, since when log shows the error message, you will be confused whether it is from L114 or L123.
I am wondering if I am right. Thanks.

littleroad: L114: "failed to copy %s to %s when opening *src* file: %v"
L114: "failed to copy %s to %s when opening *dst* file: %v"
These two strings are the same length so that they cannot be easily distinguished, I will modify src to source and dst to destination.
Thanks for your comment.

allencloud:
Please squash all your commits into a single one. @littleroad thx

pkg/fileutils/fileutils.go
littleroad: Done.

allencloud:
LGTM

Issue #1231, opened 2020-02-26
dfdaemon HTTPS protocol is not valid


readerx:
### Describetion:

dfclient config：

```
registry_mirror:
  remote: https://192.168.1.254:5000
  insecure: false
  certs:
  - /certs/ca-registry.crt

port: 8443
keypem: /certs/dragonfly.key
certpem: /certs/dragonfly.cer
```

docker command:

```
docker run -d --name dfclient -p 443:8443 \
    -v `pwd`/certs:/certs \
    -v `pwd`/dfclient.yaml:/dfclient.yaml \
    dragonflyoss/dfclient:1.0.0 --verbose --config=/dfclient.yaml
```

docker pull error:

```
Error response from daemon: Get https://192.168.1.111/v2/: http: server gave HTTP response to HTTPS client
```

error code in dfdaemon/server.go:

```
// Start runs dfdaemon's http server.
func (s *Server) Start() error {
	_ = proxy.WithDirectHandler(handler.New())(s.proxy)
	s.server.Handler = s.proxy
	if s.server.TLSConfig != nil {
		logrus.Infof("start dfdaemon https server on %s", s.server.Addr)
	} else {
		logrus.Infof("start dfdaemon http server on %s", s.server.Addr)
	}

	return s.server.ListenAndServe()
}
```

###  Environment:

- dragonfly version: 1.0.0

readerx:
@lowzj

readerx:
It was solved by #1232

PR #1232, opened 2020-02-29
Fix an issue where HTTPS is not supported

readerx:
Fix bug

Issue: #1231 

Need integration testing

Don't config docker mirrors，pull the images from a non-local address（eg：docker pull 192.168.1.100:8443/alpine:3.9）

### DIFF
--- a/dfdaemon/server.go
+++ b/dfdaemon/server.go
@@ -117,14 +117,17 @@ func NewFromConfig(cfg config.Properties) (*Server, error) {
 
 // Start runs dfdaemon's http server.
 func (s *Server) Start() error {
+	var err error
 	_ = proxy.WithDirectHandler(handler.New())(s.proxy)
 	s.server.Handler = s.proxy
 	if s.server.TLSConfig != nil {
 		logrus.Infof("start dfdaemon https server on %s", s.server.Addr)
+		err = s.server.ListenAndServeTLS("", "")
 	} else {
 		logrus.Infof("start dfdaemon http server on %s", s.server.Addr)
+		err = s.server.ListenAndServe()
 	}
-	return s.server.ListenAndServe()
+	return err
 }
 
 // Stop gracefully stops the dfdaemon http server.


readerx:
travis-ci time out

allencloud:
> travis-ci time out

restarted. @readerx

allencloud:
Thanks for your contribution. It is indeed what we missed.

inoc603:
Thanks for the PR. It's indeed a bug. We do need to use different methods for http and https.

Though according to the code in [net/https](https://golang.org/src/net/http/server.go?s=88453:88494#L2962), it seems if we set `server.TLSConfig`, we don't need to pass filenames for the key pair. A good reason for setting `server.TLSConfig` instead of using the filenames is that it will make unit tests easier since we don't have to generate a key pair to disk.

readerx:
> 
> 
> Thanks for the PR. It's indeed a bug. We do need to use different methods for http and https.
> 
> Though according to the code in [net/http](https://golang.org/src/net/http/server.go?s=88453:88494#L2962), it seems if we set `server.TLSConfig`, we don't need to pass filenames for the key pair. A good reason for setting `server.TLSConfig` instead of using the filenames is that it will make unit tests easier since we don't have to generate a key pair to disk.

`server.TLSConfig` only works for http2
Maybe I am wrong and need to confirm  : )

inoc603:
@readerx  I wrote a simple test, and it seems to work. One thing needed to be noticed is that without `NextProtos:   []string{"http/1.1"},`, the response protocol version will be http2.

```go
package test

import (
	"crypto/tls"
	"fmt"
	"net/http"
	"testing"
	"time"
)

func handler() http.Handler {
	mux := http.NewServeMux()
	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		w.WriteHeader(200)
		w.Write([]byte("response"))
	})
	return mux
}

func TestWithoutFile(t *testing.T) {
	s := http.Server{
		Addr:    "127.0.0.1:6789",
		Handler: handler(),
	}

	cert, err := tls.LoadX509KeyPair("server.crt", "server.key")
	if err != nil {
		t.Error(err)
	}

	s.TLSConfig = &tls.Config{
		NextProtos:   []string{"http/1.1"},
		Certificates: []tls.Certificate{cert},
	}

	stopped := make(chan error)
	go func() {
		stopped <- s.ListenAndServeTLS("", "")
	}()

	time.Sleep(time.Second)

	http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
	resp, err := http.Get("https://127.0.0.1:6789")
	fmt.Println(resp, err)

	s.Close()
	<-stopped
}

func TestWithFile(t *testing.T) {
	s := http.Server{
		Addr:    "127.0.0.1:6789",
		Handler: handler(),
	}

	stopped := make(chan error)
	go func() {
		stopped <- s.ListenAndServeTLS("server.crt", "server.key")
	}()

	time.Sleep(time.Second)

	http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
	resp, err := http.Get("https://127.0.0.1:6789")
	fmt.Println(resp, err)

	s.Close()
	<-stopped
}

```

readerx:
> 
> 
> @readerx I wrote a simple test, and it seems to work. One thing needed to be noticed is that without `NextProtos: []string{"http/1.1"},`, the response protocol version will be http2.
> 
> ```go
> package test
> 
> import (
> 	"crypto/tls"
> 	"fmt"
> 	"net/http"
> 	"testing"
> 	"time"
> )
> 
> func handler() http.Handler {
> 	mux := http.NewServeMux()
> 	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
> 		w.WriteHeader(200)
> 		w.Write([]byte("response"))
> 	})
> 	return mux
> }
> 
> func TestWithoutFile(t *testing.T) {
> 	s := http.Server{
> 		Addr:    "127.0.0.1:6789",
> 		Handler: handler(),
> 	}
> 
> 	cert, err := tls.LoadX509KeyPair("server.crt", "server.key")
> 	if err != nil {
> 		t.Error(err)
> 	}
> 
> 	s.TLSConfig = &tls.Config{
> 		NextProtos:   []string{"http/1.1"},
> 		Certificates: []tls.Certificate{cert},
> 	}
> 
> 	stopped := make(chan error)
> 	go func() {
> 		stopped <- s.ListenAndServeTLS("", "")
> 	}()
> 
> 	time.Sleep(time.Second)
> 
> 	http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
> 	resp, err := http.Get("https://127.0.0.1:6789")
> 	fmt.Println(resp, err)
> 
> 	s.Close()
> 	<-stopped
> }
> 
> func TestWithFile(t *testing.T) {
> 	s := http.Server{
> 		Addr:    "127.0.0.1:6789",
> 		Handler: handler(),
> 	}
> 
> 	stopped := make(chan error)
> 	go func() {
> 		stopped <- s.ListenAndServeTLS("server.crt", "server.key")
> 	}()
> 
> 	time.Sleep(time.Second)
> 
> 	http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
> 	resp, err := http.Get("https://127.0.0.1:6789")
> 	fmt.Println(resp, err)
> 
> 	s.Close()
> 	<-stopped
> }
> ```

good，i change code

dfdaemon/server.go
```diff
@@ -117,14 +117,17 @@ func NewFromConfig(cfg config.Properties) (*Server, error) {
 
 // Start runs dfdaemon's http server.
 func (s *Server) Start() error {
+	var err error
 	_ = proxy.WithDirectHandler(handler.New())(s.proxy)
 	s.server.Handler = s.proxy
 	if s.server.TLSConfig != nil {
 		logrus.Infof("start dfdaemon https server on %s", s.server.Addr)
+		err = s.server.ListenAndServeTLS("", "")
```
inoc603: We can directly return the error here, then the `err` variable will not be needed.

readerx:
> 
> 
> I think we need to set `TLSConfig.NextProtos` in `WithTLSConfigFromFile`.

http/server.go

```go
func (srv *Server) ServeTLS(l net.Listener, certFile, keyFile string) error {
	// Setup HTTP/2 before srv.Serve, to initialize srv.TLSConfig
	// before we clone it and create the TLS Listener.
	if err := srv.setupHTTP2_ServeTLS(); err != nil {
		return err
	}

	config := cloneTLSConfig(srv.TLSConfig)
	if !strSliceContains(config.NextProtos, "http/1.1") {
		config.NextProtos = append(config.NextProtos, "http/1.1")
	}

	configHasCert := len(config.Certificates) > 0 || config.GetCertificate != nil
	if !configHasCert || certFile != "" || keyFile != "" {
		var err error
		config.Certificates = make([]tls.Certificate, 1)
		config.Certificates[0], err = tls.LoadX509KeyPair(certFile, keyFile)
		if err != nil {
			return err
		}
	}

	tlsListener := tls.NewListener(l, config)
	return srv.Serve(tlsListener)
}
```

It's going to set the protocol http/1.1

dfdaemon/server.go
readerx: The code check will fail if written this way

inoc603: > The code check will fail if written this way

What's the error message exactly? I'm guessing it complains about the `else` block, since if your return in the `if` block, `else` is not needed.

inoc603:
> > I think we need to set `TLSConfig.NextProtos` in `WithTLSConfigFromFile`.
> 
> http/server.go
> 
> ```go
> func (srv *Server) ServeTLS(l net.Listener, certFile, keyFile string) error {
> 	// Setup HTTP/2 before srv.Serve, to initialize srv.TLSConfig
> 	// before we clone it and create the TLS Listener.
> 	if err := srv.setupHTTP2_ServeTLS(); err != nil {
> 		return err
> 	}
> 
> 	config := cloneTLSConfig(srv.TLSConfig)
> 	if !strSliceContains(config.NextProtos, "http/1.1") {
> 		config.NextProtos = append(config.NextProtos, "http/1.1")
> 	}
> 
> 	configHasCert := len(config.Certificates) > 0 || config.GetCertificate != nil
> 	if !configHasCert || certFile != "" || keyFile != "" {
> 		var err error
> 		config.Certificates = make([]tls.Certificate, 1)
> 		config.Certificates[0], err = tls.LoadX509KeyPair(certFile, keyFile)
> 		if err != nil {
> 			return err
> 		}
> 	}
> 
> 	tlsListener := tls.NewListener(l, config)
> 	return srv.Serve(tlsListener)
> }
> ```
> 
> It's going to set the protocol http/1.1

Sure. It seems to support http2 and http1.1 at the same time. No problem.

dfdaemon/server.go
readerx: It's not a structured program

inoc603: I'm not quite sure what you mean. But I guess this would be what you want:

```go
if s.server.TLSConfig != nil {
        logrus.Infof("start dfdaemon https server on %s", s.server.Addr)
        return s.server.ListenAndServeTLS("", "")
}
logrus.Infof("start dfdaemon http server on %s", s.server.Addr)
return s.server.ListenAndServe()
```

readerx: This code is not easy to understand，I can revise it if you insist

inoc603: I can certainly understand your argument, since the symmetry of the `if {} else {}` statement kind of matches the actual logic of the code. I'll probably insist on using early returns to reduce nesting and avoiding unnecessary intermediate variables in some other cases. But since this function is very short, I think it's fine to keep your current implementation.

inoc603:
LGTM.

@lowzj @Starnop I wonder if there's any tests we need for this? I think bugs like this ought be found much earlier.

inoc603:
@readerx would you please rebase your commits into one? Then we can merge this PR.

readerx:
> 
> 
> @readerx would you please rebase your commits into one? Then we can merge this PR.

ok, I've rebase it

allencloud:
Could you please squash the three commits into a single one? I think we do not have to add these codes via three commits. @readerx

readerx:
> 
> 
> Could you please squash the three commits into a single one? I think we do not have to add these codes via three commits. @readerx

ok，I have dealt with the problem

PR #1234, opened 2020-03-04
bugfix: fix untested case due to prefix missing

fenggw-fnst:
Signed-off-by: Guangwen Feng <fenggw-fnst@cn.fujitsu.com>

Fix untested unit test case due to prefix missing. The check.Suite requires any method to start with the "Test" prefix. So fix this.


allencloud:
LGTM, thx.

PR #1235, opened 2020-03-06
Feature: support custom protocol

jim3ma:
Signed-off-by: 楚贤 <chuxian.mjj@antfin.com>

This PR enables customization protocols for dfget and supernode.

> currently, golang plugins is not production ready, we must compile customization protocol code will dragonfly.

Register custom protocol example:

1. cmd/supernode/main.go

```
package main

import (
	helloworld

	"github.com/dragonflyoss/Dragonfly/cmd/supernode/app"
	"github.com/dragonflyoss/Dragonfly/pkg/httputils"
)

func init() {
	// here we register "helloworld" protocol
	helloworldTransport := helloworld.NewTransport()
	httputils.RegisterProtocol("helloworld", helloworldTransport)
}

func main() {
	app.Execute()
}
```

2. cmd/dfget/main.go

```
package main

import (
	helloworld

	"github.com/dragonflyoss/Dragonfly/cmd/dfget/app"
	"github.com/dragonflyoss/Dragonfly/pkg/httputils"
)

func init() {
	// here we register "helloworld" protocol
	helloworldTransport := helloworld.NewTransport()
	httputils.RegisterProtocol("helloworld", helloworldTransport)
}

func main() {
	app.Execute()
}
```

3. run with dfget

```
dfget --url helloworld://example/path/to/helloworld --output /path/to/helloworld
```

No

N/A


pkg/httputils/http_util.go
```diff
@@ -454,3 +469,22 @@ func handlePairRange(rangeStr string, length int64) (*RangeStruct, error) {
...
+//   protocols := "helloworld"
+//   newTransport := funcNewTransport
+//   httputils.RegisterProtocol(protocols, newTransport)
+func RegisterProtocol(scheme string, rt http.RoundTripper) {
+	protocols.Store(scheme, rt)
+}
+
+// RegisterProtocolOnTransport registers all new protocols in "protocols" for a special Transport
+// this function will be used in supernode and dfwget
+func RegisterProtocolOnTransport(tr *http.Transport) {
```
allencloud: Actually I am not on the way to implement register protocol. I have to say we can make it by this indeed. While I am afraid that in your code, maybe I have two kinds of suggestions:
1.  `func (tr  *http.Transport) RegisterProtocol( key string, value http.RoundTripper)`  would be much more direct. We could try to make the pattern more object oriented. If we cannot add a function for the built-in struct, we can design one based on it.
2. If we could do things above, the function would be a fundamental one, and we can use the range loop in caller rather the implementation part.

WDYT?

pkg/netutils/netutils_test.go
```diff
@@ -145,15 +148,19 @@ func (suite *NetUtilSuite) TestIsValidURL(c *check.C) {
...
 			"q=is%3Aissue+is%3Aclosed": true,
+		// FIXME because x://y/z is valid, below urls is valid
+		//"http:/www.a.b.com":     false,
+		//"https://github.com/dragonflyoss/Dragonfly/issues?" +
+		//	"q=is%3Aissue+is%3Aclosed": false,
 	}
 
 	for k, v := range cases {
-		for _, scheme := range []string{"http", "https", "HTTP", "HTTPS"} {
+		for _, scheme := range []string{"http", "https", "HTTP", "HTTPS", "test"} {
```
lowzj: add upper case of protocol `TEST` for testing

PR #1239, opened 2020-03-07
Bumped golang to 1.14.0 and alpine to 3.11

ducksecops:


allencloud:
Thanks for your work. @ducksecops  I have several thoughts here:
1. we would like your sign on each commits;
2. We have two Dockerfiles here, one for dfget that is `Dockerfile`, the other is for supernode which is `Dockerfile.supernode`. Maybe you could help to search globally in this repo inorder to double check if there is something you missed, when you wish to change some details. 
🐯 ❤️

PR #1254, opened 2020-03-19
bugfix: dfget overwrite /dev/null

xuyixin1996:
1. bugfix: dfget will overwrite the `/dev/null` and caused some issue
2. bugfix: supernode will panic when receive metrics report from dfget which the fileLength is -1 (default value)

N.A

N.A

```
dfget -u http://google.com -o /dev/null
ls /dev/null
```
before the patch, /dev/null will be a regular file
after the patch, /dev/null will not be changed

### DIFF
--- a/apis/types/task_metrics_request.go
+++ b/apis/types/task_metrics_request.go
@@ -74,6 +74,10 @@ func (m *TaskMetricsRequest) Validate(formats strfmt.Registry) error {
 		res = append(res, err)
 	}
 
+	if err := m.validateFileLength(formats); err != nil {
+		res = append(res, err)
+	}
+
 	if len(res) > 0 {
 		return errors.CompositeValidationError(res...)
 	}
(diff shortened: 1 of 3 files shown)


apis/types/task_metrics_request.go
```diff
@@ -74,6 +74,10 @@ func (m *TaskMetricsRequest) Validate(formats strfmt.Registry) error {
 		res = append(res, err)
 	}
 
+	if err := m.validateFileLength(formats); err != nil {
```
allencloud: Please do not edit this file directory, you should update swagger.yml first, and then auto-generate this file:
https://github.com/dragonflyoss/Dragonfly/blob/master/apis/README.md

ML99bug:
😄😄😄😄

allencloud:
@xuyixin1996 conflict happens, could you help to rebase to solve it? Thanks

PR #1259, opened 2020-03-24
fix: cdn-source pattern supports range task

lowzj:

#1213 has support `cdn source` pattern. The download behavior is different when user wants to download a part of the source file, just like:

```bash
dfget -u $url -o /tmp/a.test --header "Range:bytes=1-9000000"
```
1. **download piece from source**: the range of piece get from `supernode` is not the real range of the source file, so `dfget` should correct it based on the _Range_ header passed by user.
2. **download from peer**: the range of piece get from `supernode` is the range of file stored on the other peers, so `dfget` can directly use it.

### DIFF
--- a/dfget/core/api/download_api.go
+++ b/dfget/core/api/download_api.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/dragonflyoss/Dragonfly/dfget/config"
 	"github.com/dragonflyoss/Dragonfly/pkg/httputils"
+	"github.com/dragonflyoss/Dragonfly/pkg/rangeutils"
 	"github.com/dragonflyoss/Dragonfly/version"
 )
 
@@ -56,8 +57,10 @@ func NewDownloadAPI() DownloadAPI {
 }
 
 func (d *downloadAPI) Download(ip string, port int, req *DownloadRequest, timeout time.Duration) (*http.Response, error) {
+	if req == nil {
+		return nil, fmt.Errorf("nil dwonload request")
+	}
 	headers := make(map[string]string)
-	headers[config.StrRange] = config.StrBytes + "=" + req.PieceRange
 	headers[config.StrPieceNum] = strconv.Itoa(req.PieceNum)
 	headers[config.StrPieceSize] = fmt.Sprint(req.PieceSize)
 	headers[config.StrUserAgent] = "dfget/" + version.DFGetVersion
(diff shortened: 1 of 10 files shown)


dfget/core/api/download_api.go
```diff
@@ -67,12 +66,71 @@ func (d *downloadAPI) Download(ip string, port int, req *DownloadRequest, timeou
...
+		rangeStr = httputils.ConstructRangeStr(req.PieceRange)
 		url = fmt.Sprintf("http://%s:%d%s", ip, port, req.Path)
 	}
+	headers[config.StrRange] = rangeStr
 
 	return httputils.HTTPGetTimeout(url, headers, timeout)
 }
+
+func isFromSource(req *DownloadRequest) bool {
+	return req != nil && strings.Contains(req.Path, "://")
```
Starnop: It is not enough to judge that req is not nil here.

dfget/core/api/download_api.go
```diff
@@ -67,12 +66,71 @@ func (d *downloadAPI) Download(ip string, port int, req *DownloadRequest, timeou
...
+}
+
+// getRealRange
+// pieceRange: "start-end"
+// rangeHeaderValue: "bytes=sourceStart-sourceEnd"
+// return: "bytes=realStart-realEnd"
+func getRealRange(pieceRange string, rangeHeaderValue string) string {
```
Starnop: How about adding a unit test for this function. 😄

dfget/core/api/download_api.go
```diff
@@ -67,12 +66,71 @@ func (d *downloadAPI) Download(ip string, port int, req *DownloadRequest, timeou
...
+
+	realStart := start + lower
+	realEnd := end + lower
+	if realEnd > upper {
+		realEnd = upper
+	}
+	return httputils.ConstructRangeStr(fmt.Sprintf("%d-%d", realStart, realEnd))
+}
+
+func parseRange(rangeStr string) (int64, int64, error) {
```
Starnop: https://github.com/dragonflyoss/Dragonfly/blob/master/supernode/util/range_util.go#L54

It's a repeated function that we extract it as common. WDYT?

Starnop:
Please add a description for this PR, such as:
1. why did you submit the PR
2. how to use it with range task

dfget/core/api/download_api.go
lowzj: done

dfget/core/api/download_api.go
lowzj: done

lowzj:
@Starnop PTAL

dfget/core/api/download_api_test.go
```diff
@@ -0,0 +1,51 @@
...
+func (s *DownloadAPITestSuite) TestGetRealRange(c *check.C) {
+	cases := []struct {
+		pieceRange string
+		rangeValue string
+		expected   string
+	}{
+		{"0-1", "", "0-1"},
+		{"0-1", "1-100", "1-2"},
+		{"0-100", "1-100", "1-100"},
+		{"100-100", "1-100", "101-100"},
```
Starnop: Is end less than start as expected?

lowzj: Yes, this function just do compute the real range the user want to download but not do verify. The verification is done by the remote server.

Starnop: OK

Starnop:
LGTM.

Starnop:
@lowzj In additional, should we support range task for non-cdn-source pattern?

lowzj:
> @lowzj In additional, should we support range task for non-cdn-source pattern?

It has already supported.

Starnop:
> > @lowzj In additional, should we support range task for non-cdn-source pattern?
> 
> It has already supported.

👍 Merging......

PR #1272, opened 2020-04-09
Add unit test cases for download_api_test.go&algorithm_test.go&atomiccount_test.go&httpuils_test.go 

Hellcatlk:
Signed-off-by: ZouYu <zouy.fnst@cn.fujitsu.com>

 Add unit test case

NONE

NONE

NONE

NONE

### DIFF
--- a/pkg/atomiccount/atomiccount_test.go
+++ b/pkg/atomiccount/atomiccount_test.go
@@ -39,6 +39,9 @@ func (suite *AtomicCountUtilSuite) TestAdd(c *check.C) {
 
 	result := acount.Get()
 	c.Check(result, check.Equals, (int32)(12))
+
+	var nilAcount *AtomicInt
+	c.Check(nilAcount.Add(5), check.Equals, (int32)(0))
 }
 
 func (suite *AtomicCountUtilSuite) TestSet(c *check.C) {
@@ -49,4 +52,7 @@ func (suite *AtomicCountUtilSuite) TestSet(c *check.C) {
 	_ = acount.Set(1)
 	result := acount.Get()
 	c.Check(result, check.Equals, (int32)(1))
+
+	var nilAcount *AtomicInt
+	c.Check(nilAcount.Get(), check.Equals, (int32)(0))
 }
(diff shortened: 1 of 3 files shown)


dfget/util/algorithm_test.go
```diff
@@ -52,4 +52,10 @@ func (s *AlgorithmTestSuite) TestShuffle(c *check.C) {
 		Shuffle(n, func(i, j int) { isRun++ })
 		c.Assert(isRun, check.Equals, n-1)
 	}
+
+	Shuffle(1<<31, func(i, j int) {
+		if 31-1 <= i {
```
lowzj, line 57: Could you please add some description here? I'm not very clear about its meaning of this.

Hellcatlk: Because if i let the loop complete, it will time out.
So i need to make the loop stop early.

Hellcatlk: sorry, this is my mistake. I should be 1<<31-1.

pkg/atomiccount/atomiccount_test.go
```diff
@@ -39,6 +39,9 @@ func (suite *AtomicCountUtilSuite) TestAdd(c *check.C) {
 
 	result := acount.Get()
 	c.Check(result, check.Equals, (int32)(12))
+
+	var nilAcount *AtomicInt
+	c.Check(nilAcount.Add(5), check.Equals, (int32)(0))
```
lowzj, line 44: It makes me confused why there is no panic when invoking the function of nil struct pointer. After a while, I understand: `nilAcount.Add(5)` equals `(*AtomicInt).Add(nilAcount, 5)` in GoLang, and the implementation of function `AtomicInt#Add` checks `nil` pointer before dereference it.

But is it a reasonable implementation?
* It's rare to check itself in a struct's own function. Why not let it panic directly?
* The behavior of the `AtomicInt` functions are inconsistent: `Set` doesn't check.
* Any value plus zero value should be equal to itself.

@Starnop WDYT

lowzj:
Could you rebase your code and combine your commits into one commit?

Hellcatlk:
@lowzj , Of course. Done.

PR #1275, opened 2020-04-10
add lru queue and mockFileServer

wangforthinker:

Add the lru queue and mock file server. It may be used in some test.

NONE.

Already add it.

### DIFF
--- a/dfget/core/helper/helper_test.go
+++ b/dfget/core/helper/helper_test.go
@@ -17,12 +17,17 @@
 package helper
 
 import (
+	"context"
+	"fmt"
+	"io/ioutil"
 	"reflect"
 	"runtime"
 	"strings"
 	"testing"
 
 	"github.com/dragonflyoss/Dragonfly/dfget/config"
+	"github.com/dragonflyoss/Dragonfly/pkg/httputils"
+
 	"github.com/go-check/check"
 )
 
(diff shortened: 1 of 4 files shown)


hhhhsdxxxx:
lgtm!
Reviewed-by henry.hj@antfin.com

pkg/queue/lru_queue.go
```diff
@@ -0,0 +1,159 @@
...
+
+	"github.com/dragonflyoss/Dragonfly/pkg/errortypes"
+)
+
+type cQElementData struct {
+	key  string
+	data interface{}
+}
+
+type LRUQueue struct {
```
lowzj, line 31: add comments for export statement

dfget/core/helper/helper_test.go
```diff
@@ -74,3 +79,114 @@ func (s *HelperTestSuite) TestDownloadPattern(c *check.C) {
...
+	resp, err := httputils.HTTPGet(url, header)
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode >= 400 {
+		return nil, fmt.Errorf("resp code %d", resp.StatusCode)
+	}
+
+	defer resp.Body.Close()
```
allencloud, line 100: I am afraid that you should move this `defer` in front of the `400` judgement

pkg/queue/lru_queue.go
```diff
@@ -0,0 +1,159 @@
...
+package queue
+
+import (
+	"container/list"
+	"sync"
+
+	"github.com/dragonflyoss/Dragonfly/pkg/errortypes"
+)
+
+type cQElementData struct {
```
allencloud, line 26: Yes, I am on @lowzj .
Maybe you could add more comments on the struct, so that we can make it more readable.

pkg/queue/lru_queue.go
```diff
@@ -0,0 +1,159 @@
...
+		return nil, errortypes.ErrDataNotFound
+	}
+
+	q.internalPutFront(data)
+
+	return data.Value.(*cQElementData).data, nil
+}
+
+// GetFront will get several items from front and not poll out them.
+func (q *LRUQueue) GetFront(count int) []interface{} {
```
allencloud, line 89: I think we should validate the `count` here. or make the type to be unsigned.

pkg/queue/lru_queue.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	retData := data.Value.(*cQElementData).data
+	delete(q.itemMap, key)
+	q.internalRemove(data)
+
+	return retData
+}
+
+func (q *LRUQueue) internalPutFront(i *list.Element) {
```
allencloud, line 141: I do not think that we should add the prefix of `internal`, maybe we can just use `putAtFront()`, `removeFromTail` and so on.

pkg/queue/lru_queue.go
wangforthinker: done.

dfget/core/helper/helper_test.go
wangforthinker: done.

pkg/queue/lru_queue.go
wangforthinker: done.

pkg/queue/lru_queue.go
wangforthinker: done.

pkg/queue/lru_queue.go
wangforthinker: done.

pkg/queue/lru_queue.go
```diff
@@ -0,0 +1,167 @@
...
+	item := q.l.Front()
+	index := 0
+	for {
+		if item == nil {
+			break
+		}
+
+		result[index] = item.Value.(*cQElementData).data
+		index++
+		if index >= count {
```
lowzj, line 110: The `index` should be less the length of the list.

wangforthinker: if list is traversed to the end, the loop will be break by the case "item == nil"

PR #1277, opened 2020-04-10
update api definition of /peer/heartbeat and /peer/network

wangforthinker:

Update the api of /peer/heartbeat and /peer/network. 

NONE.

no.

### DIFF
--- a/apis/swagger.yml
+++ b/apis/swagger.yml
@@ -649,7 +649,7 @@ paths:
           $ref: "#/responses/500ErrorResponse"
 
   /peer/heartbeat:
-    get:
+    post:
       summary: "report the heart beat to super node."
       description: |
         This endpoint is mainly for reporting the heart beat to supernode.
@@ -657,7 +657,7 @@ paths:
       parameters:
         - name: "body"
           in: "body"
-          description: "request body which contains peer info."
+          description: "request body which contains base info of peer"
           schema:
             $ref: "#/definitions/HeartBeatRequest"
       responses:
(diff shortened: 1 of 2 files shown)


apis/swagger.yml
```diff
@@ -655,11 +655,23 @@ paths:
         This endpoint is mainly for reporting the heart beat to supernode.
         And supernode could know if peer is alive in strem mode.
       parameters:
-        - name: "body"
-          in: "body"
-          description: "request body which contains peer info."
-          schema:
-            $ref: "#/definitions/HeartBeatRequest"
+        - name: ip
```
allencloud, line 658: Is this a Post request, rather than GET?

apis/swagger.yml
```diff
@@ -655,11 +655,23 @@ paths:
...
+          in: query
+          required: true
+          description: "IP of peer."
+          type: "string"
```
allencloud, line 662: format should be `ipv4`

wangforthinker: In api defintion, paramter.types should in [string number boolean integer array], which could not set ipv4.

lowzj: `format: 'ipv4'` if we only support `ipv4`

wangforthinker: done.

apis/swagger.yml
```diff
@@ -1486,31 +1499,6 @@ definitions:
         items:
           $ref: "#/definitions/Node"
 
-  HeartBeatRequest:
```
allencloud, line 1489: If the HTTP request method could be changed into `POST`, please remain the `HeartBeatRequest ` here. That could be better.

wangforthinker: done

allencloud:
LGTM

PR #1279, opened 2020-04-11
test: add unit test case for func IsSuperCID

hs0210:
Signed-off-by: Hu Shuai <hus.fnst@cn.fujitsu.com>


Starnop:
Thanks for your contribution. And well done. However, I have to mention one thing, we need to distinguish what code needs to write unit tests, what code does not need. Maybe here is a discussion you can refer to https://stackoverflow.com/questions/1316848/unit-testing-what-not-to-test. 😄

PR #1285, opened 2020-04-15
pkg: add bitmap in pkg

wangforthinker:

Add bitmap struct.

NONE.

Add unit test.


hhhhsdxxxx:
lgtm!

pkg/bitmap/bitmap.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+// RestoreBitMap generate the BitMap by input bytes.
+func RestoreBitMap(data []byte) (*BitMap, error) {
+	bm := DecodeToUintArray(data)
+
+	return &BitMap{
+		bm:          bm,
+		maxBitIndex: int32(len(bm)*64 - 1),
+	}, nil
```
xujihui1985, line 56: 这个地方似乎不会有错？需要返回error吗

xujihui1985:
lgtm
Reviewed-by jihui.xjh@antfin.com

pkg/bitmap/bitmap.go
wangforthinker: done

pkg/bitmap/bitmap.go
```diff
@@ -0,0 +1,211 @@
...
+	"sync"
+
+	lbm "github.com/openacid/low/bitmap"
+)
+
+// BitMap is a struct which provides the Get or Set of bits map.
+type BitMap struct {
+	lock        sync.RWMutex
+	bm          []uint64
+	maxBitIndex int32
```
allencloud, line 31: could we just make it a type of `uint32` ?
Then we could ignore validation of it. Otherwise I think we had better validate it when initialization.

wangforthinker: add the validate of all parameters.

wangforthinker: done.

allencloud:
Thanks for the review, @hhhhsdxxxx @xujihui1985 
Could you help to pay another review here? @lowzj Thanks

pkg/bitmap/bitmap.go
```diff
@@ -0,0 +1,228 @@
...
+
+// Set sets or cleans the bits in range [start, end]. if setBit is true, set bits. else clean bits.
+func (b *BitMap) Set(start uint32, end uint32, setBit bool) error {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
+	return b.setWithoutLock(start, end, setBit)
+}
+
+func (b *BitMap) getWithoutLock(start uint32, end uint32, setBit bool) ([]*BitsRange, error) {
```
lowzj, line 91: The `get` operation is much slower than the `set` operation. Here is a faster algorithm to implement the `get` operation in most cases and its time complexity is as same as `set`: `O((end-start)>>6)`.

Firstly there is basic function(named `__builtin_ctz` in `C` or `runtime/internal/sys#Ctz64` in `GoLang`) should be introduced: counts trailing (low-order) zeroes.

The main flow is similar to `set` and the core steps to process one `uint64` number `x` are as following:
```golang
if !setBit
    x = ^x
while x:
    // start is the value of the trailing zeros of x
    start = Ctz64(x)
    // remove the trailing zeros of x and counts again
    end = Ctz64(^x>>start) + start
    x = x >> end << end
```

wangforthinker: Good suggestion! I have  updated it by the algorithm.

pkg/bitmap/bitmap.go
```diff
@@ -0,0 +1,351 @@
...
+		if x == 0 {
+			break
+		}
+
+		// start is the value of the trailing zeros of x
+		start = uint(Ctz64(x))
+		// remove the trailing zeros of x and counts again
+		end = uint(Ctz64(uint64(^(x >> start)))) + start
+
+		res = append(res, &BitsRange{
```
lowzj, line 192: I think the `append` is a very time-consuming operation compared to the other ops in this function. It's better to remember the `previous end position` of the whole `BitMap`, compare to `current start position` and only execute `append` when the `previous end position` is different from `current start position`, otherwise assign the value of `current start position` to `previous end` then execute the next loop.

wangforthinker: Good suggestion! Done by it.

Issue #1293, opened 2020-04-22
[feature request] make the dfget to get the list of supernodes more flexible


lowzj:
## Why you need it?
Now `dfget` gets the list of supernodes from configuration file or command line parameters.
It's inconvenient to update the list or 
* inconvenient to update client's configuration when scale up or scale down the supernodes cluster.
* cannot customize the selection logic that better fits the user's scenario

## How it could be?
Dragonfly defines the interface how to get supernodes and provides a default implementation that get supernodes from config/CLI. Then users can write a new implementation to select supernodes.

![image](https://user-images.githubusercontent.com/2419214/79968894-56532480-84c3-11ea-8103-6944b862e14e.png)

* `SupernodeLocator` is defined interface which dfget's `core flow` invokes
* `StaticLocator` is the default implementation Dragonfly provided
* `DynamicLocator` is a user customized implementation for example: 
    * it get supernode from `StaticLocator` firstly, and return it to caller if exists
    * otherwise, it get the list of supernodes from remote manager and store the result into local cache for next time

PR #1294, opened 2020-04-22
feature: define the interface SupernodeLocator

lowzj:

issues: #1293 

This pull request just defines the `SupernodeLocator` interface, I will open another pull request to refactor the dfclient to use this interface to get supernode.

### DIFF
--- a/dfget/locator/locator.go
+++ b/dfget/locator/locator.go
@@ -1,0 +1,77 @@
+/*
+ * Copyright The Dragonfly Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package locator
+
+// SupernodeLocator defines the way how to get available supernodes.
+// Developers can implement their own locator more flexibly , not just get the
+// supernode list from configuration or CLI.
+type SupernodeLocator interface {
+	// Get returns the current selected supernode, it should be idempotent.
+	Get() *Supernode
+
+	// Next chooses the next available supernode for retrying or other
+	// purpose. The current supernode should be set as this result.
+	Next() *Supernode
+
+	// GetGroup returns the group with the giving name.
+	GetGroup(name string) *SupernodeGroup
+
+	// All returns all the supernodes.
+	All() []*SupernodeGroup
+
+	// Report records the metrics of the current supernode in order to choose a
+	// more appropriate supernode for the next time if necessary.
+	Report(node string, metrics *SupernodeMetrics)
+
+	// Refresh refreshes all the supernodes.
+	Refresh() bool
+}
+
+// SupernodeGroup groups supernodes which have same attributes.
+// For example, we can group supernodes by region, business, version and so on.
+// The implementation of SupernodeLocator can select a supernode based on the
+// group.
+type SupernodeGroup struct {
+	Name  string
+	Nodes []*Supernode
+
+	// Infos stores other information that user can customized.
+	Infos map[string]string
+}
+
+// GetNode return the node with the giving index.
+func (sg *SupernodeGroup) GetNode(idx int) *Supernode {
+	if idx < 0 || idx >= len(sg.Nodes) {
+		return nil
+	}
+	return sg.Nodes[idx]
+}
+
+// Supernode holds the basic information of supernodes.
+type Supernode struct {
+	Schema    string
+	IP        string
+	Port      int
+	Weight    int
+	GroupName string
+	Metrics   *SupernodeMetrics
+}
+
+// SupernodeMetrics holds metrics used for the locator to choose supernode.
+type SupernodeMetrics struct {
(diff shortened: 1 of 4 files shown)


dfget/locator/locator.go
```diff
@@ -0,0 +1,66 @@
...
+// SupernodeGroup groups supernodes which has same attributes.
+type SupernodeGroup struct {
+	Name  string
+	Nodes []*Supernode
+
+	// Infos stores other information that user can customized.
+	Infos map[string]string
+}
+
+// Supernode basic information of supernodes.
```
SataQiu, line 53:
```suggestion
// Supernode holds the basic information of supernodes.
```

dfget/locator/locator.go
```diff
@@ -0,0 +1,66 @@
...
+type Supernode struct {
+	Schema    string
+	IP        string
+	Port      int
+	Weight    int
+	GroupName string
+	Metrics   *SupernodeMetrics
+}
+
+// SupernodeMetrics the metrics used for locator to choose supernode.
```
SataQiu, line 63:
```suggestion
// SupernodeMetrics holds metrics used for the locator to choose supernode.
```

dfget/locator/static_locator.go
```diff
@@ -0,0 +1,155 @@
...
+	"sync/atomic"
+	"time"
+
+	"github.com/dragonflyoss/Dragonfly/dfget/config"
+	"github.com/dragonflyoss/Dragonfly/dfget/util"
+	"github.com/dragonflyoss/Dragonfly/pkg/netutils"
+)
+
+func init() {
+	rand.Seed(time.Now().UnixNano())
```
SataQiu, line 30: Maybe we can consider moving `rand.Seed(time.Now().UnixNano())` to the beginning of the main function to ensure that it always work and also avoids calling it everywhere. 😕 

```go
func main() {
	rand.Seed(time.Now().UnixNano())
	
	app.Execute()
}
```

dfget/locator/static_locator.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+	return []*SupernodeGroup{s.Group}
+}
+
+func (s *StaticLocator) Report(node string, metrics *SupernodeMetrics) {
+	// unnecessary to implement this method
+	return
+}
+
+func (s *StaticLocator) Refresh() bool {
```
SataQiu, line 129: This looks more like Reset :sweat_smile: How about rename as `Reset`.

dfget/locator/static_locator.go
lowzj: How about move `rand.Seed` to the top package of `dfget` rather than `cmd`? I think this operation should be transparent to the caller.

dfget/locator/static_locator.go
lowzj: This implementation looks more like `Reset`. But considering the different behaviors of other implementations, it may get a new list of supernodes via invoking this method.
So IMOP, `Reresh` is  more appropriate. WDYT?

SataQiu: Make sense!

dfget/locator/static_locator.go
SataQiu: Sure. That's OK with me. Just don't like calling `rand.Seed` everywhere!

lowzj: In order to ensure the independence of this pr, I open another pr #1318  to do this, please take a look.

dfget/locator/static_locator.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// ----------------------------------------------------------------------------
+// constructors
+
+// NewStaticLocator constructs StaticLocator which uses the nodes passed from
+// configuration or CLI.
+func NewStaticLocator(nodes []*config.NodeWeight) *StaticLocator {
+	locator := &StaticLocator{}
+	if nodes == nil || len(nodes) == 0 {
```
Starnop, line 48: Using `len(nodes) == 0` is enough here.

dfget/locator/static_locator.go
```diff
@@ -0,0 +1,155 @@
...
+
+func (s *StaticLocator) Next() *Supernode {
+	if s.Group == nil || s.load() >= len(s.Group.Nodes) {
+		return nil
+	}
+	idx := s.inc()
+	if idx >= len(s.Group.Nodes) {
+		return nil
+	}
+	return s.Group.Nodes[idx]
```
Starnop, line 107: How about extract a common method to get node with specified index?

```
func (s *StaticLocator) getNode(index int32) *Supernode(){
    if s.Group == nil || index >= len(s.Group.Nodes) {
		return nil
    }
    return s.Group.Nodes[idx]
}
```

dfget/locator/locator.go
```diff
@@ -0,0 +1,66 @@
...
+type SupernodeLocator interface {
+	// Get returns the current selected supernode, it should be idempotent.
+	Get() *Supernode
+
+	// Next chooses the next available supernode for retrying or other
+	// purpose. The current supernode should be set as this result.
+	Next() *Supernode
+
+	// GetGroup returns the group.
+	GetGroup(name string) *SupernodeGroup
```
Starnop, line 31: Just out of curiosity, why consider add a supernode group concept here? And the `SupernodeGroup` can only be retrieved for viewing. Is it expected that I can choose a group to use?

Starnop:
Please have a rebase for this PR. 😄

dfget/locator/locator.go
lowzj: For example, we can group supernodes by region, business, version and so on. The implementation of SupernodeLocator can select a supernode based on the group. The caller does not need to care about the specific selection logic, but if it is really needed, the supernode list can also be obtained through this interface.

Starnop:
LGTM.

Issue #1284, opened 2020-04-15
Proposal: Dragonfly supports seed  pattern to solve massive proxy requests which may cause huge pressure on rss or supernode.


wangforthinker:
### Backgrounds 
Now the clients of Dragonfly should communicate with supernode when proxying every request.  Thus, when massive machines of a cluster request a large file, supernode may be the key node which influences the continuous improvement of better performance. 

On the other word, every proxy request should at least cost four RTT, such as "register", "peerwait", "getPeerInfo", "proxy from target peer". The first three RTT could be considered as the extra RTT. In our case, there are some demand for low latency. And every request may be small range request and the costs of extra RTT could not be ignored.

So a solution is needed to reduce the extra RTT and the pressure of supernode.

### Idea
`Seed pattern` provides the asynchronously communication with supernode, in which peer could asynchronously fetch the p2p network info from supernode and self simply schedule the target peer to get data.

A seed represents a file which is always defined by `taskUrl`. A peer could be selected as a seed, which provides the seed file to be downloaded by other peers.  And only the seed node could request the resource data from `rss`, by the way which reduces the pressure of `rss`.

#### Feature
-  long live peerserver in dfdaemon.
-  the control stream between peer and  supernode is asynchronization.
-  supernode selects the seed nodes.
-  a peer could be selected as various seeds which represent different url.
-  non-seed peer downloads the range bytes of file from peer nodes.
-  seed allows concurrent download requests even if seed file is not already cached.
-  seed info of peer is persistent even if peer restarts.

#### Architecture Diagram
This  diagram describes the control flow and part data flow.
![image](https://user-images.githubusercontent.com/4704263/79307480-90538200-7f29-11ea-9476-a5149477ff1f.png)

- The communication between supernode and peer is asynchronization, which  includes  `HeartBeat`,`Report`,`Fetch`. 
- `HeartBeat` reports the peer base info to supernode every serval seconds. Thus supernode could update the aliveness of peers.
- `Report` allows peer to report local seeds  info to supernode.
- `Fetch` allows peer to fetch the seeds info of p2p network, so peer knows the location of seed file to get data from seed node.

#### Seed Data Flow
This  diagram describes how the seed node provides the data services.
![image](https://user-images.githubusercontent.com/4704263/79308567-63a06a00-7f2b-11ea-90cc-0ade91d28c98.png)

- `Downloader` requests the data of seed file from `rss`.
- `Uploader` provides the data service to be downloaded by other peers.

PR #1297, opened 2020-04-23
add seed downloader

wangforthinker:
Add downloader for seed file. Please see refer https://github.com/dragonflyoss/Dragonfly/issues/1284.

NONE.

Add ut.

### DIFF
--- a/dfdaemon/seed/downloader.go
+++ b/dfdaemon/seed/downloader.go
@@ -1,0 +1,130 @@
+/*
+ * Copyright The Dragonfly Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


xujihui1985:
LGTM

dfdaemon/seed/downloader.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/dragonflyoss/Dragonfly/pkg/limitreader"
+	"github.com/dragonflyoss/Dragonfly/pkg/ratelimiter"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+// downloader manage the downloading of seed file.
+type downloader interface {
+	DownloadToWriterAt(ctx context.Context, rangeStruct httputils.RangeStruct, timeout time.Duration, writeOff int64, writerAt io.WriterAt, rateLimit bool) (length int64, err error)
```
lowzj, line 39: The parameter list doesn't contain `url` and it's initialized in creating instance of a concrete implementation. Why is it designed this way?

And this method has 2 things to do:
1. downloads data from remote server.
2. writes data into target writer.

I think that the step 2 may be not the responsibility of downloader interface. Will it be better to remove it from the interface?

wangforthinker: 1. a downloader binds to a seed, so the information of seed will be transferred  by creating instance of a concrete implementation.  In the other word, we may support other protocol like http/https/oss/hdfs ,so the parameters may not be easy to define.
2. in our case, we want to cover the error handler of download. And the data finally should be write ,so we do write in  the interface. And some optimization could be make in the interface, such as buffer Write-through which reduces the costs of copying buffer.

PR #1298, opened 2020-04-24
feature: modify dfdaemon port validate method, allow to expose more p…

Komey:

modify dfdaemon port validate method, allow to expose more port to listen.

### Ⅱ. Does this pull request to fix one issue?

NONE

Here is a case:
We have a private docker registry, named https://docker.pt.komey.com.
and we want to make our user to None care about the registry address,
So we'd like to config our DNS（docker.pt.komey.com） record for a specific region to 127.0.0.1 and expose dfdaemon on 443 port. Which may make our users have No feeling about the difference between the regions which didn't deploy dragonfly.

### DIFF
--- a/dfdaemon/config/config.go
+++ b/dfdaemon/config/config.go
@@ -118,7 +118,7 @@ type Properties struct {
 
 // Validate validates the config
 func (p *Properties) Validate() error {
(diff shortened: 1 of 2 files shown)


lowzj:
Thanks for your contributes!
Could you please combine your 3 commits into one?

PR #1302, opened 2020-04-26
fix: GCDSlice returns 1 when slice is empty

lowzj:

The `GCDSlice` does not consider the situation when the giving slice is nil or empty.
This pr makes it returns 1 when s is empty because that any number divided by 1 is still
itself.


Starnop:
LGTM.

PR #1324, opened 2020-05-06
refactor: redefine the behavior of SupernodeLocator

lowzj:

* redefine the behavior of method Get: It should return nil before first calling the Next method.
* add method Size to return the number of supernodes
* fix the implementation of StaticLocator


Starnop:
LGTM.

PR #1352, opened 2020-05-19
Feat/add implementation for dfget interface

wangforthinker:
Add implementation of http protocol of dfget interface.

NONE.

add ut.

### DIFF
--- a/pkg/protocol/http/http.go
+++ b/pkg/protocol/http/http.go
@@ -1,0 +1,191 @@
+/*
+ * Copyright The Dragonfly Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package http
+
+import (
+	"crypto/tls"
+	"fmt"
+	"net"
+	"net/http"
+	"time"
+
+	"github.com/dragonflyoss/Dragonfly/pkg/errortypes"
+	"github.com/dragonflyoss/Dragonfly/pkg/protocol"
+)
+
+var (
+	// DefaultTransport is default implementation of http.Transport.
+	DefaultTransport = newDefaultTransport()
+
+	// DefaultClient is default implementation of Client.
+	DefaultClient = &Client{
+		client:    &http.Client{Transport: DefaultTransport},
+		transport: DefaultTransport,
+	}
+)
+
+const (
+	// http protocol name
+	ProtocolHTTPName = "http"
+
+	// https protocol name
+	ProtocolHTTPSName = "https"
+)
+
+func init() {
+	protocol.RegisterProtocol(ProtocolHTTPName, &ClientBuilder{})
+	protocol.RegisterProtocol(ProtocolHTTPSName, &ClientBuilder{supportHTTPS: true})
+}
+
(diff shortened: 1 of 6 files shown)


pkg/protocol/http/http.go
```diff
@@ -0,0 +1,95 @@
...
+	// DefaultClient is default implementation of Client.
+	DefaultClient = &Client{
+		client:    &http.Client{Transport: DefaultClientOpt.transport},
+		transport: DefaultClientOpt.transport,
+	}
+)
+
+const (
+	// http protocol name
+	ProtocolName = "http"
```
lowzj, line 41: Should it contain `http` and `https`?

pkg/protocol/http/http.go
```diff
@@ -0,0 +1,95 @@
...
+	protocol.RegisterProtocol(ProtocolName, &ClientBuilder{})
+}
+
+// ClientOpt is the argument of NewProtocolClient.
+type ClientOpt struct {
+	transport http.RoundTripper
+}
+
+// Client is an implementation of protocol.Client for http protocol.
+type Client struct {
```
lowzj, line 54: create an anonymous instant to check whether the struct implements the interface, like:
```golang
var _ protocol.Client = &Client{}
```

pkg/protocol/http/http.go
```diff
@@ -0,0 +1,95 @@
...
+		hd:     hd,
+		client: cli,
+	}
+}
+
+// ClientBuilder is an implementation of protocol.ClientBuilder for http protocol.
+type ClientBuilder struct{}
+
+func (cb *ClientBuilder) NewProtocolClient(clientOpt interface{}) (protocol.Client, error) {
+	opt, ok := clientOpt.(*ClientOpt)
```
lowzj, line 86: Does it allow to `nil` clientOpt?

pkg/protocol/http/http.go
```diff
@@ -0,0 +1,95 @@
...
+
+	if md != nil {
+		h, ok := md.(*Headers)
+		if ok {
+			hd = h
+		}
+	}
+
+	if hd == nil {
+		hd = NewHTTPMetaData().(*Headers)
```
lowzj, line 72: Why must create it? I think, it can be `nil`.

pkg/protocol/http/resource.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	res, err := rs.doRequest(req)
+	if err != nil {
+		return 0, err
+	}
+
+	defer res.Body.Close()
+	lenStr := res.Header.Get(config.StrContentLength)
+	if len(lenStr) == 0 {
```
lowzj, line 78: use `lenStr == ""` to instead of this?

pkg/protocol/http/http.go
wangforthinker: Add htttps support.

pkg/protocol/http/http.go
wangforthinker: Add it.

pkg/protocol/http/http.go
wangforthinker: Allow nil clientOpt.

pkg/protocol/http/http.go
wangforthinker: Allow it.

pkg/protocol/http/resource.go
wangforthinker: Done.

pkg/protocol/http/http.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+func init() {
+	protocol.RegisterProtocol(ProtocolHTTPName, &ClientBuilder{})
+	protocol.RegisterProtocol(ProtocolHTTPSName, &ClientBuilder{supportHTTPS: true})
+}
+
+// ClientOpt is the argument of NewProtocolClient.
+type ClientOpt struct {
+	transport http.RoundTripper
```
wangforthinker, line 55: change to map[string]interface, tell the user the keys which could be set.

PR #1366, opened 2020-05-28
refactor: use package server/api to register supernode's API

lowzj:

use package server/api to register supernode's APIs


supernode/server/api/utils.go
```diff
@@ -52,26 +52,28 @@ func ParseJSONRequest(req io.Reader, target interface{}, validator ValidateFunc)
...
 }
 
-// EncodeResponse encodes response in json.
-// The response body is empty if the data is nil or empty value.
-func EncodeResponse(w http.ResponseWriter, code int, data interface{}) error {
+// SendResponse encodes response in json.
+//
+// TODO:
+//  Should the response body should be empty if the data is nil or empty
+//  string? Now it's incompatible with the client.
```
zhouhaibing089, line 59: I don't think we should. If the data is nil/empty, it basically means there is an error somehow, in that case, the `code` should reflect which error it is, and data should be used to describe what error it is.

supernode/server/api/utils.go
```diff
@@ -52,26 +52,28 @@ func ParseJSONRequest(req io.Reader, target interface{}, validator ValidateFunc)
...
+	return enc.Encode(data)
 }
 
 // HandleErrorResponse handles err from server side and constructs response
 // for client side.
 func HandleErrorResponse(w http.ResponseWriter, err error) {
 	switch e := err.(type) {
 	case *errortypes.HTTPError:
-		_ = EncodeResponse(w, e.Code, errResp(e.Code, e.Msg))
+		_ = SendResponse(w, e.Code, errResp(e.Code, e.Msg))
```
zhouhaibing089, line 73: This is odd, The error should not be ignored silently, right?

supernode/server/api/utils.go
lowzj: Ok, maybe we should provide a method to send header only.

supernode/server/api/utils.go
lowzj: done

supernode/server/router.go
```diff
@@ -50,91 +98,46 @@ func initRoute(s *Server) *mux.Router {
...
-			HandleErrorResponse(w, err)
-		}
-	}
+func handleMetrics(ctx context.Context, rw http.ResponseWriter, req *http.Request) error {
+	promhttp.Handler().ServeHTTP(rw, req)
+	return nil
 }
 
-// EncodeResponse encodes response in json.
+// SendResponse encodes response in json.
```
Starnop, line 140: s/ SendResponse/ EncodeResponse

supernode/server/api/api.go
```diff
@@ -47,6 +47,9 @@ func newCategory(name, prefix string) *category {
 	apiCategories[name] = &category{
 		name:   name,
 		prefix: prefix,
+		handlerSpecs: []*HandlerSpec{
+			listHandler(name),
```
Starnop, line 51: why should we add a default list handler here?

lowzj: I can help up to get all the available APIs provided by supernode.

Starnop:
LGTM.

PR #1368, opened 2020-05-29
add functional options for new protocol client

wangforthinker:

add functional options as arguments for NewProtocolClient.

NONE.

no need.

### DIFF
--- a/pkg/protocol/http/http.go
+++ b/pkg/protocol/http/http.go
@@ -49,12 +49,10 @@ const (
 func init() {
 	protocol.RegisterProtocol(ProtocolHTTPName, &ClientBuilder{})
 	protocol.RegisterProtocol(ProtocolHTTPSName, &ClientBuilder{supportHTTPS: true})
-}
 
-const (
-	HTTPTransport = "http.transport"
-	TLSConfig     = "tls.config"
-)
+	protocol.RegisterMapInterfaceOptFunc(ProtocolHTTPName, WithMapInterface)
+	protocol.RegisterMapInterfaceOptFunc(ProtocolHTTPSName, WithMapInterface)
+}
 
 func newDefaultTransport() *http.Transport {
 	// copy from http.DefaultTransport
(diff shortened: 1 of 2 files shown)


zhouhaibing089:
May I suggest this: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis?

wangforthinker:
> May I suggest this: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis?

Good idea!  By the way,  the argument for NewProtocolClient will be more sensible and extensive.
Now I have implemented it by functional options, could you have a review again. 
ping @zhouhaibing089  @lowzj

wangforthinker:
I define a register function which registers functional options of which argument is map[string]string, by the way, developer could easily  create protocol client without care about instance of protocol. 
ping @zhouhaibing089 @lowzj

pkg/protocol/http/http.go
```diff
@@ -72,54 +70,12 @@ func newDefaultTransport() *http.Transport {
...
-	return v
-}
-
 var _ protocol.Client = &Client{}
 
 // Client is an implementation of protocol.Client for http protocol.
 type Client struct {
 	client    *http.Client
-	transport http.RoundTripper
+	transport *http.Transport
```
zhouhaibing089, line 78: This seems to be a bad switch to me. Do you have a strong reason of doing this?

I can see a lot of benefits of using `http.RoundTripper`, e.g a mocking roundtripper like [go-vcr][1].

[1]: https://github.com/dnaeon/go-vcr

pkg/protocol/http/http.go
```diff
@@ -146,46 +102,104 @@ type ClientBuilder struct {
...
+		cli.transport = transport
+		cli.client = &http.Client{
+			Transport: transport,
 		}
+
+		return nil
+	}
+}
+
+func (cb *ClientBuilder) NewProtocolClient(opts ...func(client protocol.Client) error) (protocol.Client, error) {
```
zhouhaibing089, line 183: Do we still need ClientBuilder? What're the extra value added by using a builder?

zhouhaibing089:
I have two questions, but roughly looks good to me. :)

pkg/protocol/http/http.go
wangforthinker: Good advice, now switch to http.RoundTripper again.

pkg/protocol/http/http.go
wangforthinker: ClientBuilder is to generate client for protocol client.  For some protocol, the client needs to set some specific info to initialize client by default.

wangforthinker:
Now, the default functional option which protocol should implement is changed from func(map[string]string) to func(map[string]interface{}) for  more general.
@zhouhaibing089 @lowzj PTAL.

PR #1373, opened 2020-06-01
add config for protocol and pattern

wangforthinker:

Add config for protocol and pattern when dfdaemon starts with config.

NONE.

### DIFF
--- a/cmd/dfdaemon/app/root.go
+++ b/cmd/dfdaemon/app/root.go
@@ -203,6 +203,10 @@ func getConfigFromViper(cmd *cobra.Command, v *viper.Viper) (*config.Properties,
 		cfg.DFRepo = filepath.Join(cfg.WorkHome, "dfdaemon/data/")
 	}
 
+	if cfg.Pattern == "" {
+		cfg.Pattern = "p2p"
+	}
+
 	return &cfg, cfg.Validate()
 }
 
--- a/dfdaemon/config/config.go
+++ b/dfdaemon/config/config.go
@@ -114,6 +114,10 @@ type Properties struct {
 	LocalIP    string          `yaml:"localIP" json:"localIP"`
 	PeerPort   int             `yaml:"peerPort" json:"peerPort"`
 	StreamMode bool            `yaml:"streamMode" json:"streamMode"`
+
+	ProtocolConf []ProtocolConfig `yaml:"protocolConf" json:"protocolConf"`
+	Pattern      string           `yaml:"pattern" json:"pattern"`
+	PatternConf  PatternConfig    `yaml:"patternConf" json:"patternConf"`
 }
 
 // Validate validates the config
@@ -412,4 +416,15 @@ func NewProxy(regx string, useHTTPS bool, direct bool, redirect string) (*Proxy,
 // Match checks if the given url matches the rule.
 func (r *Proxy) Match(url string) bool {
 	return r.Regx != nil && r.Regx.MatchString(url)
+}
+
+// ProtocolConfig is the config of url protocol.
+type ProtocolConfig struct {
+	Name string `yaml:"name" json:"name"`
+	// Opts defines the opts for protocol, such as tls config.
+	Opts map[string]string `yaml:"opts" json:"opts"`
+}
+
+type PatternConfig struct {
+	Opts map[string]string `yaml:"opts" json:"opts"`
 }
--- a/dfdaemon/downloader/downloader.go
+++ b/dfdaemon/downloader/downloader.go
@@ -18,7 +18,10 @@ package downloader
 
 import (
 	"context"
+	"fmt"
 	"io"
+
+	"github.com/dragonflyoss/Dragonfly/dfdaemon/config"
 )
 
 // Interface specifies on how an plugin can download a file.
@@ -37,3 +40,24 @@ type Stream interface {
 // Factory is a function that returns a new downloader.
 type Factory func() Interface
 type StreamFactory func() Stream
+
+type StreamFactoryBuilder func(patternConfig config.PatternConfig, c config.Properties) Stream
+
+var (
+	registerFactory = map[string]StreamFactoryBuilder{}
+)
+
+func Register(pattern string, builder StreamFactoryBuilder) {
+	registerFactory[pattern] = builder
+}
+
+func NewStreamFactory(pattern string, patternConfig config.PatternConfig, c config.Properties) StreamFactory {
+	builder, ok := registerFactory[pattern]
+	if !ok {
+		panic(fmt.Sprintf("pattern %s not registered", pattern))
+	}
+
+	return func() Stream {
+		return builder(patternConfig, c)
+	}
+}
--- a/dfdaemon/downloader/p2p/dfclient.go
+++ b/dfdaemon/downloader/p2p/dfclient.go
@@ -22,7 +22,14 @@ import (
 	"io"
 
 	"github.com/dragonflyoss/Dragonfly/dfdaemon/config"
+	"github.com/dragonflyoss/Dragonfly/dfdaemon/downloader"
 )
+
+func init() {
+	downloader.Register("p2p", func(patternConfig config.PatternConfig, c config.Properties) downloader.Stream {
+		return NewClient(c.DFGetConfig())
+	})
+}
 
 type Client struct {
 }
--- a/dfdaemon/proxy/proxy.go
+++ b/dfdaemon/proxy/proxy.go
@@ -31,7 +31,6 @@ import (
 	"github.com/dragonflyoss/Dragonfly/dfdaemon/config"
 	"github.com/dragonflyoss/Dragonfly/dfdaemon/downloader"
 	"github.com/dragonflyoss/Dragonfly/dfdaemon/downloader/dfget"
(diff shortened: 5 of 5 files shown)


lowzj:
@inoc603 PTAL

dfdaemon/config/config.go
```diff
@@ -413,3 +417,14 @@ func NewProxy(regx string, useHTTPS bool, direct bool, redirect string) (*Proxy,
 func (r *Proxy) Match(url string) bool {
 	return r.Regx != nil && r.Regx.MatchString(url)
 }
+
+// ProtocolConfig is the config of url protocol.
+type ProtocolConfig struct {
+	Name string `yaml:"name" json:"name"`
+	// Opts defines the opts for protocol, such as tls config.
+	Opts map[string]string `yaml:"opts" json:"opts"`
```
inoc603, line 425: If this config will be available for plugins, maybe we should use `map[string]interface{}` to allow more complicated configuration.

wangforthinker:
Add PatternMatcher to match url request to  choose pattern.  @inoc603 @lowzj PTAL.

dfdaemon/config/config.go
```diff
@@ -114,6 +114,11 @@ type Properties struct {
 	LocalIP    string          `yaml:"localIP" json:"localIP"`
 	PeerPort   int             `yaml:"peerPort" json:"peerPort"`
 	StreamMode bool            `yaml:"streamMode" json:"streamMode"`
+
+	ProtocolConf []ProtocolConfig `yaml:"protocolConf" json:"protocolConf"`
+	// Mode is the default pattern.
+	Mode        string          `yaml:"mode" json:"mode"`
+	PatternConf []PatternConfig `yaml:"patternConf" json:"patternConf"`
```
inoc603, line 121: It would be better if `PatternConf` and `PatternConfig` gets changed as well. I assume you want to say something like "dfddaemon 以 P2P 模式运行"，in that case `mode` is the appropriate word to use. `pattern` is confusing.

wangforthinker: Here, I want to express how to package some progresses such as download, upload, data encode/decode, schedule.  In this case, pattern will be more accurate expression.

dfdaemon/config/config.go
```diff
@@ -114,6 +114,11 @@ type Properties struct {
 	LocalIP    string          `yaml:"localIP" json:"localIP"`
 	PeerPort   int             `yaml:"peerPort" json:"peerPort"`
 	StreamMode bool            `yaml:"streamMode" json:"streamMode"`
+
+	ProtocolConf []ProtocolConfig `yaml:"protocolConf" json:"protocolConf"`
+	// DefaultPattern is the default pattern.
+	DefaultPattern string          `yaml:"defaultPattern" json:"defaultPattern"`
```
zhouhaibing089, line 120: `Defaultxxx` is odd to me. Default values are normally defined by the system itself.

This is a configuration definition, in another word, this is user facing. I'd suggest calling it as just `Pattern` or as @inoc603 mentioned `Mode`.

dfdaemon/config/config.go
```diff
@@ -413,3 +418,15 @@ func NewProxy(regx string, useHTTPS bool, direct bool, redirect string) (*Proxy,
...
+// ProtocolConfig is the config of url protocol.
+type ProtocolConfig struct {
+	Name string `yaml:"name" json:"name"`
+	// Opts defines the opts for protocol, such as tls config.
+	Opts map[string]interface{} `yaml:"opts" json:"opts"`
+}
+
+type PatternConfig struct {
+	Pattern string                 `yaml:"pattern" json:"pattern"`
+	Opts    map[string]interface{} `yaml:"opts" json:"opts"`
```
zhouhaibing089, line 431: `interface{}` inside a configuration is trouble-some. It is not easy to see what it is really.

If possible, this(the type) needs to be spelled out explicitly.

cmd/dfdaemon/app/root.go
```diff
@@ -203,6 +203,10 @@ func getConfigFromViper(cmd *cobra.Command, v *viper.Viper) (*config.Properties,
 		cfg.DFRepo = filepath.Join(cfg.WorkHome, "dfdaemon/data/")
 	}
 
+	if cfg.DefaultPattern == "" {
```
zhouhaibing089, line 206: It comes to me naturally this should be written in a different way:

```golang
const DefaultPattern = "p2p"

..
    if cfg.Pattern == "" {
        cfg.Pattern = DefaultPattern
    }
```

I'm uncertain we should call it `DefaultPattern` in the configuration type. Could we just call it `Patter` or whatever fit?

dfdaemon/config/config.go
```diff
@@ -114,6 +114,11 @@ type Properties struct {
 	LocalIP    string          `yaml:"localIP" json:"localIP"`
 	PeerPort   int             `yaml:"peerPort" json:"peerPort"`
 	StreamMode bool            `yaml:"streamMode" json:"streamMode"`
+
+	ProtocolConf []ProtocolConfig `yaml:"protocolConf" json:"protocolConf"`
+	// DefaultPattern is the default pattern.
+	DefaultPattern string          `yaml:"defaultPattern" json:"defaultPattern"`
+	PatternConf    []PatternConfig `yaml:"patternConf" json:"patternConf"`
```
zhouhaibing089, line 121: Instead of having a default pattern key as specified above, does it also make sense to just assuming the first one in this list is the default one?

dfdaemon/config/config.go
inoc603: Agreed. But it seems the content of this this configuration is determined by corresponding plugins, and we're not sure what the type would be. If this was json, I would recommend using `json.RawMessage`, but there's no yaml equivalent. I a plugin, we may need to `yaml.Marshal` first then `yaml.Unmarshal` again to unmarshal this field into the specific type.

dfdaemon/config/config.go
inoc603: If `PatternConfig` is meant to be used as general plugin configration, then at least a `pattern` should be categorized. For example, `cfg.DefaultPattern` clearly means `DefaultDownloadPattern`. And even so, I'm pretty sure `pattern` doesn't mean what you want it to mean. I'd rather call it `PluginConfig` to show that it configures some customizable part of the program.

inoc603: Maybe we should call it `DownloadMode` and `DownloadConf` just like `Protocol` ?

dfdaemon/config/config.go
inoc603: > Instead of having a default pattern key as specified above, does it also make sense to just assuming the first one in this list is the default one?

If we always take the first one as default, then you need to change the order of `PatternConf` to change the default, which is kind of wierd. `PatternConf` should mean "what is available". And `PatternConf` might also be empty for a minimal quick start configuration.

zhouhaibing089: Just thinking loud, the default value `p2p` as mentioned in setup functions, what happens if there is no entry for `p2p` here?

And also from an operation perspective, the way I need to pick a key from another part just makes it more error prone - typo, non-existent, etc.

And really, do we really need to provide a default one? If client doesn't spell out it explicitly, it should be reasonable to report a failure, no?

The default pattern, as far as I understand, is really a concern from client perspective.

dfdaemon/config/config.go
zhouhaibing089: I think maybe it's fair to say we can start with `map[string]string`.

inoc603: Or maybe something like this?

```go
type RawYAML yaml.MapSlice

func (msg RawYAML) Unmarshal(v interface{}) error {
        body, _ := yaml.Marshal(msg)
        return yaml.Unmarshal(body, v)
}
```

Then each plugin can have their own typed config:

```go
var customPluginConfig struct {
  // ...
}

config.Opts.Unmarshal(&customPluginConfig)
```

zhouhaibing089: Sounds good to me. Thanks for sharing this.

wangforthinker: Actually,  echo plugin could unmarshal `map[string]interface{}` into their own config. And json and yaml should be both supported in reading config, so we could not use type of json or yaml as field of opt. On this condition, `map[string]interface{}` may be a good  choice.

dfdaemon/config/config.go
wangforthinker: In our design, we provide two ways to choose pattern for one request. One is static selection, which is configuration from config file as  the default pattern; Another is dynamic way,  in which user could choose pattern by specify param in request, and by patternMatcher specific request will run in specific pattern.

dfdaemon/config/config.go
wangforthinker: Hm, DownloadMode and DownloadConf may be a good advice!

wangforthinker: @inoc603 PTAL

dfdaemon/config/config.go
wangforthinker: @inoc603 @zhouhaibing089  PTAL

cmd/dfdaemon/app/root.go
wangforthinker: Done.

inoc603:
@wangforthinker I think we need some unit tests for `PatternMatcher`

Issue #1380, opened 2020-06-05
bug: supernode cannot re-download a uncompleted task which has header parameter


lowzj:
### Ⅰ. Issue Description
as described in the title

### Ⅱ. Describe what happened

supernode downloads the file always failed before deleting it manually or in GC-procedure.

### Ⅲ. Describe what you expected to happen

The client should download successfully after retrying.

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1. use `dfget` to download a large file with non-empty `--header` flag 
    ```bash
    dfget -u 'http://127.0.0.1:8001/a.test' -o /tmp/a.test --header 'User-Agent:dfget' --notbs
    ```
2. restart supernode before downloading completed
3. use `dfget` to download the same file again before the supernode delete the task in GC-procedure, it will get failed

### Ⅴ. Anything else we need to know?

dfclient.log

```bash
2020-06-05 17:01:51.393 INFO sign:87141-1591347711.391 : get cmd params:["./bin/darwin_amd64/dfget" "-u" "http://127.0.0.1:8003/a.test" "-o" "/tmp/a.test" "--console" "--notbs" "--header" "User-Agent:dfget"]
2020-06-05 17:01:51.393 INFO sign:87141-1591347711.391 : get init config:{"url":"http://127.0.0.1:8003/a.test","output":"/tmp/a.test","pattern":"p2p","header":["User-Agent:dfget"],"notbs":true,"console":true,"localLimit":"20MB","minRate":"64KB","clientQueueSize":6,"workHome":"/Users/zj/.small-dragonfly","logConfig":{"maxSize":0,"maxBackups":0,"path":"/Users/zj/.small-dragonfly/logs/dfclient.log"}}
2020-06-05 17:01:51.393 INFO sign:87141-1591347711.391 : target file path:/tmp/a.test
[...]

PR #1381, opened 2020-06-05
bugfix: cdn cannot re-download uncompleted task with header param

lowzj:

fixes: #1380

### DIFF
--- a/supernode/httpclient/origin_http_client.go
+++ b/supernode/httpclient/origin_http_client.go
@@ -132,18 +132,16 @@ func (client *OriginClient) GetContentLength(url string, headers map[string]stri
 
 // IsSupportRange checks if the source url support partial requests.
 func (client *OriginClient) IsSupportRange(url string, headers map[string]string) (bool, error) {
-	// set headers
-	if headers == nil {
-		headers = make(map[string]string)
-	}
-	headers["Range"] = "bytes=0-0"
+	// set headers: headers is a reference to map, should not change it
+	copied := CopyHeader(nil, headers)
+	copied["Range"] = "bytes=0-0"
 
 	// send request
-	resp, err := client.HTTPWithHeaders(http.MethodGet, url, headers, 4*time.Second)
+	resp, err := client.HTTPWithHeaders(http.MethodGet, url, copied, 4*time.Second)
 	if err != nil {
 		return false, err
 	}
-	resp.Body.Close()
+	_ = resp.Body.Close()
 
 	if resp.StatusCode == http.StatusPartialContent {
 		return true, nil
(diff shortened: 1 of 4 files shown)


supernode/httpclient/origin_http_client.go
```diff
@@ -132,18 +132,16 @@ func (client *OriginClient) GetContentLength(url string, headers map[string]stri
...
 // IsSupportRange checks if the source url support partial requests.
 func (client *OriginClient) IsSupportRange(url string, headers map[string]string) (bool, error) {
-	// set headers
-	if headers == nil {
-		headers = make(map[string]string)
-	}
-	headers["Range"] = "bytes=0-0"
+	// set headers: headers is a reference to map, should not change it
+	copied := CopyHeader(nil, headers)
+	copied["Range"] = "bytes=0-0"
```
zhouhaibing089, line 137: Is this same as `delete(copied, "Range")`?

zhouhaibing089:
I do have a general question. Header is usually typed with `map[string][]string`.

https://github.com/golang/go/blob/fca286bed3ed0e12336532cc711875ae5b3cb02a/src/net/http/header.go#L21

And it looks like there is a builtin [Clone][2] in standard library.

[2]: https://github.com/golang/go/blob/fca286bed3ed0e12336532cc711875ae5b3cb02a/src/net/http/header.go#L90

supernode/httpclient/origin_http_client.go
lowzj: Yeah, but I'm afraid that the `header` maybe read in other places, so I just copy it.

lowzj: > Is this same as `delete(copied, "Range")`?

No, it just read 1 byte from the remote server to test whether the server supports `range` request.

zhouhaibing089: Oh, I see. Thanks!

lowzj:
> I do have a general question. [Header](https://github.com/golang/go/blob/fca286bed3ed0e12336532cc711875ae5b3cb02a/src/net/http/header.go#L21) is usually typed with `map[string][]string`.
> 
> And it looks like there is a builtin [Clone](https://github.com/golang/go/blob/fca286bed3ed0e12336532cc711875ae5b3cb02a/src/net/http/header.go#L90) in standard library.

Yes, the arg `headers` will be parsed into `http.Header` in the `supernode/httpclient`. I think that it's better to change the content after parsing.

lowzj:
> And it looks like there is a builtin [Clone](https://github.com/golang/go/blob/fca286bed3ed0e12336532cc711875ae5b3cb02a/src/net/http/header.go#L90) in standard library.

The `Clone` method is not public until go version 1.13.

zhouhaibing089:
> The Clone method is not public until go version 1.13.

Oh, good to know. If we're using go1.13 for compilation, then I think it is nice to align with it. Otherwise, I think we could just copy it over.

Starnop:
LGTM.

PR #1382, opened 2020-06-08
test: add unit test case for func NewHTTPError

hs0210:
Signed-off-by: Hu Shuai <hus.fnst@cn.fujitsu.com>


wangforthinker:
LGTM

pkg/errortypes/errortypes_test.go
```diff
@@ -90,3 +90,9 @@ func (suite *ErrorTestSuite) TestIsRangeNotSatisfiable(c *check.C) {
 	c.Assert(IsRangeNotSatisfiable(*err1), check.Equals, true)
 	c.Assert(IsRangeNotSatisfiable(*err2), check.Equals, false)
 }
+
+func (suite *ErrorTestSuite) TestNewHTTPError(c *check.C) {
```
zhouhaibing089, line 94: Ah, what is the value of adding this test?

Unit tests are usually added for functions, maybe `Error()` function is the one should be tested here?

PR #1388, opened 2020-06-11
add seed pattern demo for dfget interface

wangforthinker:


lowzj:
This is just a demo of new feature.

Issue #428, opened 2019-02-27
[feature request]an area aware schedule policy to dragonfly.


xq2248:
name: Feature Request
about: Suggest your excellent idea for this project

In networks with large cross, the network are divided into different areas. Supernode is depolyed in the root node ,  the dfget clients are deployed in different area. All dfget Clients can download files from supernode with pattern cdn. But a dfget client can only download file pieces from other peers within the same area, also it can only serve pieces download requests to the peers within the same area.

I suggest:
Dfget command add a new optional flag "-a xxx"(for short) or "--area=xxx", to sepcify the area of the peer. When dfget send peer regist request to the supernode, supernode store the area property of the peer. Thus supernode can know the area of all the peers. When dfget send pull piece request to the supernode, supernode only returns the peers within the same area, or return peers within the same  area priority to peer within different area.

xq2248:
英语比较差，用中文描述以下。
在跨度很大的网络中，supernode可能部署在网络中心，supernode可以给所有区域中的peer提供CDN下载服务。而dfget peer所在的网络存在两种可能：
A dfget peer可能部署在网络隔离的不同区域，一个区域内的peer之间可以相互对等传输，而不同区域可能网络无法联通，导致无法跨区域对等传输。这个时候supernode应该只返回同一区域的peers来提供P2P下载服务。
B 有的时候，即便不同区域的peer网络能够相通，但是同一个区域之间的peers P2P对传是最快的。这个时候supernode应该优先返回同一区域内的peers来提供P2P下载服务，其次才是其他区域的peers。

实现：
dfget命令增加一个可选的参数 -a xxx 或者--area=xxx。
dfget peer向supernode发送peer regist请求时，携带这个peer的area位置参数。
supernode记录peer的area。
对于场景A，supernode只返回同一区域的peers信息。
对于场景B，supernode优先返回同一区域的peers信息，只有同一区域没有可以提供服务的peer时，才返回其他区域的peers信息。

allencloud:
Thanks for your feature request. For a straightforward understanding of your proposal, I think what you request is totally reasonable. It is an extension of scheduler and it takes the network reality into consideration.  
Would you first give some input on this topic? @lowzj @Starnop @garfield009 
And in addition, maybe we can review this in our community meeting.

lowzj:
[...]

PR #1397, opened 2020-06-23
filter peers of different areas

xq2248:
This PR implements the area awareness scheduler policy to the peers.
Previous discussion as the issue 428:
https://github.com/dragonflyoss/Dragonfly/issues/428

When dfget takes no area parameter,  supernode scheduler policy does not channge.
When dfget take a area with register reqeust, supernode keeps the area info with peerMgr and dfgetTask, and schedule the pieces downloading requests of src peer to the dst peers which is in the same areas. If no dst peer can be found within the same area, supernode will schedule the pieces downloading request to cdn mgr.

### DIFF
--- a/apis/types/df_get_task.go
+++ b/apis/types/df_get_task.go
@@ -68,6 +68,9 @@ type DfGetTask struct {
 
 	// task Id
 	TaskID string `json:"taskId,omitempty"`
+
+	// area is the area of this peer
+	Area string `json:"area,omitempty"`
 }
 
 // Validate validates this df get task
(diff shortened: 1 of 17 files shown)


jim3ma:
How about change `area` to `region` and `zone`, these is more general concepts.

xq2248:
> 
> 
> How about change `area` to `region` and `zone`, these is more general concepts.

Great!

xq2248:
@lowzj how to generate api/types objects in local repo?

xq2248:
@lowzj
The region parameter in swagger.yml has been added.

PR #1402, opened 2020-06-28
feature: get LocalIP from supernode HTTP _ping API first rather than …

### LINKED ISSUE #1399: dfget got Supernode’s Service Cluster IP as peer ip in k8s cluster
### Ⅰ. Issue Description
https://github.com/dragonflyoss/Dragonfly/blob/master/pkg/httputils/http_util.go#L393

kube-proxy mode=ipvs
supernode with service.spec.type=ClusterIP(eg. 10.254.49.255)

when dfget or dfclient startup, it will get ClusterIP(eg. 10.254.49.255) as cfg.LocalIP which will register to supernode as peer ip

### Ⅱ. Describe what happened
as above 
supernode 用service.spec.type=ClusterIP下 kube-proxy mode=ipvs。 https://github.com/dragonflyoss/Dragonfly/blob/master/pkg/httputils/http_util.go#L393  
这行返回的localIP 会等于 supernode service 的cluster ip(eg. 10.254.49.255)
dfget or dfclient 启动时会把这个ip作为peer 的ip注册到supernode

临时解决方案 Headless Services 

1.  kube-proxy mode=ipvs
2. supernode with service.spec.type=ClusterIP(eg. 10.254.49.255)

### Ⅵ. Environment:

- dragonfly version: 1.0.2
- OS (e.g. from /etc/os-release): Centos 7.3
- Kernel (e.g. `uname -a`): 4.9.2-3.0.0.std7b.el7.5.x86_64
- Install tools: k8s daemonset(dfclient) + statefulSet(supernode) with hostnetwork=true
- Others:

Komey:
feature: get LocalIP from supernode HTTP _ping API first rather than TCP connection due to Kube-service in ipvs-mod will get cluster IP for LocalAddr

1. supernode _ping API will return remote address
2. dfclient will query supernode/_ping to get localIP
3. if the localIP not in dfclient network interfaces will fallback to use TCP Dial connection

 fixes #1399

×

1. k8s cluster within kube-proxy run in ipvs mod
2. dfclient connect to supernode in service cluster ip
3. dfclient will get log like `Connect to supernode get self IP: %s`


dfget/core/api/supernode_api.go
```diff
@@ -74,6 +76,24 @@ type supernodeAPI struct {
...
+	var (
+		code int
+		body []byte
+	)
+	url := fmt.Sprintf("%s://%s%s",
+		api.Scheme, node, pingPath)
+	if code, body, e = api.HTTPClient.Get(url, api.Timeout); e != nil {
+		return "", e
+	}
+	if !httputils.HTTPStatusOk(code) {
```
allencloud, line 91: Is the StatusOK(200) the only case to judge whether a node could be pinged?

I  don't think we should use HTTP protocol (which is level 7) to test the connection.

PR #1411, opened 2020-07-05
bugfix: fix dfget panic when crossWrite is true

zcc35357949:
Signed-off-by: zhouchencheng <zhouchencheng@bilibili.com>

### Ⅰ. Describe what this PR did
when  dfget's `--output` and `--home`  directories are mounted under different disk device, it makes the hard link between them fails to create. At that time, `crossWrite` will be true.

And the same piece will be written to different files twice. Currently the `autoreset` field in piece is always true, so the piece buffer will reset after the first written, then the second written will trigger panic.

### Ⅱ. Does this pull request fix one issue?
fixes #1410

PR #1418, opened 2020-07-08
fix buffer overwrite issue

### LINKED ISSUE #1415: dfget download file success with incorrect md5
### Ⅰ. Issue Description
I use master code. Currently dfget may download file sucessfully, but the md5 value is incorrect.         
retry the same file many times, md5 values are different!! 
```
root@nvm-14:~ # /data/app/src/github.com/dragonflyoss/Dragonfly/cmd/dfdaemon/dfget -u http://172.16.17.24:8000/go1.12.9.linux-amd64.tar.gz -o /tmp/go1.12.9.linux-amd64.tar.gz --pattern p2p -s 800MB --totallimit 800MB --node 172.16.17.24:8002,172.16.17.24:8002 --verbose
--2020-07-06 21:50:33--  http://172.16.17.24:8000/go1.12.9.linux-amd64.tar.gz
dfget version:
workspace:/root/.small-dragonfly
sign:21233-1594043433.459
client:172.16.17.24 connected to node:172.16.17.24:8002
start download by dragonfly...
download SUCCESS cost:1.240s length:127961523 reason:0
root@nvm-14:~ # md5sum /tmp/go1.12.9.linux-amd64.tar.gz
6a5e658f7c0eeac60f14dc425abc007e  /tmp/go1.12.9.linux-amd64.tar.gz
root@nvm-14:~ # /data/app/src/github.com/dragonflyoss/Dragonfly/cmd/dfdaemon/dfget -u http://172.16.17.24:8000/go1.12.9.linux-amd64.tar.gz -o /tmp/go1.12.9.linux-amd64.tar.gz --pattern p2p -s 800MB --totallimit 800MB --node 172.16.17.24:8002,172.16.17.24:8002 --verbose
--2020-07-06 21:50:45--  http://172.16.17.24:8000/go1.12.9.linux-amd64.tar.gz
dfget version:
workspace:/root/.small-dragonfly
sign:22691-1594043445.594
client:172.16.17.24 connected to node:172.16.17.24:8002
start download by dragonfly...
download SUCCESS cost:0.260s length:127961523 reason:0
root@nvm-14:~ # md5sum /tmp/go1.12.9.linux-amd64.tar.gz
24da2f521dbf610f01de747fe3e04dab  /tmp/go1.12.9.linux-amd64.tar.gz
```
The file size is 123MB.

I change the default writer buffer size to 4MB，the problem goes away!!
https://github.com/dragonflyoss/Dragonfly/blob/master/pkg/pool/writer_pool.go#L28
4M is the default piece size for under 200M file. I just don't know the relation between them.

1.
2.
3.

### Ⅵ. Environment:

- dragonfly version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

jim3ma:

RawContent func returns a buffer io, when enable autoReset, the buffer io can be reused by AcquireBufferSize, so, if first piece is processed slow than an other, the faster piece will overwrite buffer.

```
func (p *Piece) RawContent(noWrapper bool) *bytes.Buffer {
	contents := p.Content.Bytes()
	length := len(contents)
	defer func() {
		if p.autoReset {
			p.ResetContent()
		}
	}()

	if noWrapper {
		return bytes.NewBuffer(contents[:])
	}
	if length >= 5 {
		return bytes.NewBuffer(contents[4 : length-1])
	}
	return nil
}
```

fix #1415 

N/A

N/A

### DIFF
--- a/dfget/core/downloader/p2p_downloader/client_writer.go
+++ b/dfget/core/downloader/p2p_downloader/client_writer.go
@@ -246,6 +246,9 @@ func writePieceToFile(piece *Piece, file *os.File, cdnSource apiTypes.CdnSource)
 
 	writer := pool.AcquireWriter(file)
 	_, err := io.Copy(writer, piece.RawContent(noWrapper))
+	if piece.autoReset {
+		piece.ResetContent()
+	}
 	pool.ReleaseWriter(writer)
 	writer = nil
 	return err
--- a/dfget/core/downloader/p2p_downloader/client_stream_writer.go
+++ b/dfget/core/downloader/p2p_downloader/client_stream_writer.go
@@ -165,6 +165,9 @@ func (csw *ClientStreamWriter) writePieceToPipe(p *Piece) error {
 		}
 
 		_, err := io.Copy(csw.pipeWriter, p.RawContent(csw.cdnSource == apiTypes.CdnSourceSource))
+		if p.autoReset {
+			p.ResetContent()
+		}
 		if err != nil {
 			return err
 		}
--- a/dfget/core/downloader/p2p_downloader/piece.go
+++ b/dfget/core/downloader/p2p_downloader/piece.go
@@ -66,11 +66,6 @@ type Piece struct {
 func (p *Piece) RawContent(noWrapper bool) *bytes.Buffer {
 	contents := p.Content.Bytes()
 	length := len(contents)
-	defer func() {
-		if p.autoReset {
-			p.ResetContent()
-		}
-	}()
 
 	if noWrapper {
 		return bytes.NewBuffer(contents[:])


jim3ma:
@lowzj @Starnop PTAL

dfget/core/downloader/p2p_downloader/client_writer.go
```diff
@@ -246,6 +246,9 @@ func writePieceToFile(piece *Piece, file *os.File, cdnSource apiTypes.CdnSource)
 
 	writer := pool.AcquireWriter(file)
 	_, err := io.Copy(writer, piece.RawContent(noWrapper))
+	if piece.autoReset {
+		piece.ResetContent()
+	}
```
zcc35357949, line 251: You can not ResetContent here in writePieceToFile.
Because the same piece may be used in target writer queue after write finish in client writer queue. 
https://github.com/dragonflyoss/Dragonfly/blob/master/dfget/core/downloader/p2p_downloader/client_writer.go#L223

jim3ma: when enable cross write, we will read piece content twice? but this is a buffer, it has index, when read twice, I think we need reset some index, #1411 is a good workaround.

zcc35357949: > when enable cross write, we will read piece content twice? but this is a buffer, it has index, when read twice, I think we need reset some index, #1411 is a good workaround.

#1411 still can not solve this problem. Underlying Buf data will be reused. Piece should be reset after all writer process.

zcc35357949: I have rewritten the fix code #1411 , could you help me review whether it can work?

lowzj:
IMOP, the pull request #1411 is an appropriate solution. I think that we can work on it to make it better.

dfget/core/downloader/p2p_downloader/client_writer.go
jim3ma: _great!_

Issue #1416, opened 2020-07-07
unit test failed in all probability in CI environment when pull request 


YesterdayxD:
### Ⅰ. Issue Description

unit test failed in all probability in CI environment when pull request 

### Ⅱ. Describe what happened
unit test printed the following error:
FAIL: TestSuite/TestReleaseBuffer

### Ⅲ. Describe what you expected to happen
unit test passes

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1. run unit test
2.
3.

### Ⅴ. Anything else we need to know?
Chencheng Zhou refered to the following comment about the GET method of the Type Pool in Dragonfly DingDing DevGroup.
```
// Get selects an arbitrary item from the Pool, removes it from the
// Pool, and returns it to the caller.
// Get may choose to ignore the pool and treat it as empty.
// Callers should not assume any relation between values passed to Put and
// the values returned by Get.
//
// If Get would otherwise return nil and p.New is non-nil, Get returns
// the result of calling p.New.
func (p *Pool) Get() interface{}
```

### Ⅵ. Environment:

- dragonfly version: master 
- OS (e.g. from /etc/os-release): CI environment(I am not sure)
- Kernel (e.g. `uname -a`): CI environment ((I am not sure)
- Install tools:
- Others:

jim3ma:
@lowzj PTAL

PR #1426, opened 2020-07-13
Fix the bug in unit tests about sync.Pool

SataQiu:
Callers should not assume any relation between values passed to `Put` and the values returned by `Get`.

Ref: https://github.com/golang/go/blob/master/src/sync/pool.go#L116-L151
```go
// Get selects an arbitrary item from the Pool, removes it from the
// Pool, and returns it to the caller.
// Get may choose to ignore the pool and treat it as empty.
// Callers should not assume any relation between values passed to Put and
// the values returned by Get.
//
// If Get would otherwise return nil and p.New is non-nil, Get returns
// the result of calling p.New.
func (p *Pool) Get() interface{} {
```

Ref: https://github.com/golang/go/blob/master/src/sync/pool_test.go#L28-L42
```go
	// Make sure that the goroutine doesn't migrate to another P
	// between Put and Get calls.
	Runtime_procPin()
	p.Put("a")
	p.Put("b")
	if g := p.Get(); g != "a" {
		t.Fatalf("got %#v; want a", g)
	}
	if g := p.Get(); g != "b" {
		t.Fatalf("got %#v; want b", g)
	}
	if g := p.Get(); g != nil {
		t.Fatalf("got %#v; want nil", g)
	}
	Runtime_procUnpin()
```

It seems that we should pin the P when do the unit test :)

Fixes: #1416

```sh
make unit-test
```

### DIFF
--- a/pkg/pool/buffer_pool_test.go
+++ b/pkg/pool/buffer_pool_test.go
@@ -1,3 +1,6 @@
+// BufferPool is no-op under race detector, so all these tests do not work.
+// +build !race
+
 /*
  * Copyright The Dragonfly Authors.
  *
@@ -18,6 +21,7 @@ package pool
 
 import (
 	"fmt"
+	"runtime"
 	"testing"
 
 	"github.com/stretchr/testify/suite"
@@ -51,6 +55,11 @@ func (s *BufferPoolTestSuite) TestAcquireBuffer() {
 }
 
 func (s *BufferPoolTestSuite) TestReleaseBuffer() {
+	// Limit to 1 processor to make sure that the goroutine doesn't migrate
+	// to another P between AcquireBuffer and ReleaseBuffer calls.
+	prev := runtime.GOMAXPROCS(1)
+	defer runtime.GOMAXPROCS(prev)
+
 	buf1 := AcquireBuffer()
 	ReleaseBuffer(buf1)
 	ReleaseBuffer(nil)
--- a/pkg/pool/writer_pool_test.go
+++ b/pkg/pool/writer_pool_test.go
@@ -1,3 +1,6 @@
+// WriterPool is no-op under race detector, so all these tests do not work.
+// +build !race
+
 /*
  * Copyright The Dragonfly Authors.
  *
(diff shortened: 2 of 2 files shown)


YesterdayxD:
#1418 @SataQiu

lowzj:
It's caused by the race detecting of `go test -race`. It has 1/4 possibility to drop the value in `sync.Pool.Put` function.

pkg/pool/buffer_pool_test.go
```diff
@@ -1,3 +1,6 @@
+// BufferPool is no-op under race detector, so all these tests do not work.
+// +build !race
+
```
SataQiu, line 3: Ref: https://github.com/golang/go/blob/master/src/sync/pool_test.go#L5-L6

pkg/pool/buffer_pool_test.go
```diff
@@ -51,6 +55,11 @@ func (s *BufferPoolTestSuite) TestAcquireBuffer() {
 }
 
 func (s *BufferPoolTestSuite) TestReleaseBuffer() {
+	// Limit to 1 processor to make sure that the goroutine doesn't migrate
+	// to another P between AcquireBuffer and ReleaseBuffer calls.
+	prev := runtime.GOMAXPROCS(1)
+	defer runtime.GOMAXPROCS(prev)
+
```
SataQiu, line 62: Ref: https://github.com/golang/go/blob/master/src/sync/pool_test.go#L28-L30

pkg/pool/buffer_pool_test.go
lowzj: It's a workaround solution, but it will not run unit-testing of this file in current CI flow.

SataQiu: Maybe we can do something in `hack/unit-test.sh`, such as adding `-race` dynamically for `go test` :)
https://github.com/dragonflyoss/Dragonfly/blob/12cfd62bfc56554a650c506bc5d51a44fcd115ad/hack/unit-test.sh#L25-L28

PR #1443, opened 2020-07-28
docs: modify the container id to be run in user guide

qhb1001:

In the [Start Supernode](https://github.com/qhb1001/Dragonfly/blob/master/docs/user_guide/install_server.md#start-supernode) section of server installation user guide, the environment variable `${supernodeDockerImageId}` is not added into the command below. 

If the default command is executed, the user-defined docker container will not be run. Rather, the official supernode would be started. And I correct the container to be run.

None

docs


Starnop:
LGTM.

Issue #1436, opened 2020-07-18
Proposal: Implementation Plan of P2P streaming based Supernode Scheduler and Uploader


qhb1001:
# Backgrounds
As mentioned in the #1164, the current download process would be a defect if the Dragonfly client is running on a cloud disk: the local file system is not the target place, and the downloader would have to write the file one more time to another file system. This is kind of like the `sendfile` function. 
# Idea
Remove the redundency of writing the file to the local file system. Directly upload the data to the target place by storing the data in memory as the sliding window in TCP protocol. Please look at #1164 to get more information about the great idea. 
# Implementation Plan
## Design of Supernode Scheduler
The key point is to maintain the sender window, and to verify the stream mode is on. 

*   Add `wnd` to `apis/types/task_create_request.go:18#TaskCreateRequest`: The task status would be initialized to progress manager during the task registry step. 

    By default, the window size is set to -1, which means that the file is downloaded in regular mode.

*   Add new struct `SlidingWindow` to the progress manager struct, which has the fields `wnd`(window size) and `una`(oldest unacknowledged sequence number [RFC 793](https://tools.ietf.org/html/rfc793#page-25)). 

*   Add `wnd` to `supernode/daemon/mgr/progress/progress_manager.go:52#manager` struct. During the registry process of task, the progress manager is initialized. Update the `SlidingWindow` field of progress manager. 

    By default, the SlidingWindow is nil, which means that the file is downloaded in regular mode.  

*   Modify the behavior of `GetPieceProgressByCID` method of progress manager based on the `SlidingWindow` field: only return the piece info within the window range. 

*   Update the `una` field of `SlidingWindow` in progress manager in the `updateTask` method of task manager. 

## Design of Uploader
*   Since there could be multiple tasks running at the same time, which means that one peer server could have multiple client stream writers at the same time, it is necessary to add `csw []*ClientStreamWriter` field to the current `peerServer` struct in `dfget/core/uploader/peer_server.go` file. [It seems that this implementation would make the downloader and uploader tightly coupled. Optimization should be considered.]

*   Add Locker for the cache field in `ClientStreamWriter` struct in `dfget/core/downloader/p2p_downloader/client_stream_writer.go` file.

*   Extract `PeerServer` launch process from `registerToSuperNode()` method to the `Start()` method.

    And register the client stream writer to peer server after the download process is created in `Start()` method. 

    The `Start()` method lies at the `dfget/core/core.go` file.

[...]

Issue #1164, opened 2020-01-09
Proposal: Dragonfly supports P2P based on Streaming to reduce disk IO


lowzj:
## Backgrounds

Now the client of Dragonfly will random read and write disk multiple times during the downloading process. 
For directly using `dfget` to download a file:
* `dfget` random writes a piece into disk after downloading it
* `dfget server` random reads the piece from disk to share it
* `dfget` sequential reads the file from disk after downloading to do checksum

And for using `dfdaemon` to pull images, there're extra disk IO by Dragonfly:
* `dfdaemon` sequential reads the file from disk to send it to `dockerd`

It's not a problem when the host has a local disk. But it will be a potential 
 bottleneck when the Dragonfly client runs on a virtual machine with a cloud disk, all the disk IO will become network IO which has a bad performance when read/write at the same time.

So a solution is needed to reduce the IO times generated by Dragonfly.

## Idea

**P2P Streaming** is a P2P based on Streaming, which sends the data downloaded by using p2p pattern to the user directly, in order to achieve the purpose of reading and writing to disk as few as possible.

### P2P Streaming Data Flow

This diagram describes the p2p streaming data flow.

![image](https://user-images.githubusercontent.com/2419214/72035542-2cd70a00-32d3-11ea-982f-a903218c8b7f.png)

* `Piece Data Cache` stores the pieces' data in memory that can be shared to the other peers. A piece's data should be putted into this cache after downloading, and be evicted according to the `LRU` strategy when the cache is full.
* `StreamIO` sends pieces' data to callers in ascending order of `piece's number`.
* In the scenario of using `dfdaemon` to pulling images and others , the `dfdaemon` and `dfget` should be merged into one process that can reduce the time of starting `dfget` process.
* Also `dfget` can be as an individual process to download files directly.

### P2P Streaming Sliding Window

The `P2P Streaming Sliding Window` is designed to control the number of pieces of a file that can be scheduled and downloaded to avoid unlimited memory usage. This idea comes from tcp sliding window, but its minimal transmission unit is a `piece` not a `byte`.

![image](https://user-images.githubusercontent.com/2419214/72036461-e505b200-32d5-11ea-929e-0955ce5be993.png)

* `Memory Cache` is the `Piece Data Cache` to share pieces in the p2p network. The larger the cache, the higher the p2p transmission efficiency.

SataQiu:
Great job! I am willing to help.

xujihui1985:
[...]

PR #1447, opened 2020-07-31
feature: support stream mode in supernode

qhb1001:

The main work is to implement the P2P stream mode in supernode side. To summarize, the streaming download process in supernode could be cateloged into two sections. 

* Maintain the sliding window in supernode. 

    * Initialize: The size of window will be registered in the dfget task registration phrase. And then, the window state is recorded in the ProgressManager as the SyncMap indexed by ClientID. The size of window is staic once it is recorded by supernode. 
    * Update: The dfget client will report to the supernode about the result of downloading for every piece. In the `report` API of supernode server, the window will be updated based on the received piece number and piece status. To be specific, the window keeps sliding until the unacknowledged piece. 
    * Finish: Update the `DeleteCID` method of ProgressManager to delete the sliding window state if it is in stream mode. 

* Schedule the pieces according to the window state

    The only modification that I made in this part lies at the `GetPieceProgressByCID` method of ProgressManager. The available pieces in regular mode means the success pieces which are not running; while for the stream mode, the available pieces means the cached pieces which are not running. After the modification, the ProgressManager would only return the available pieces inside the window when stream mode is on. 

    It should be noted that, the scheduler in stream mode currently uses the same scheduler as the regular mode, which may demands future optimization. 

Besides the above modifications, I also add two APIs in supernode server, which are `addPieceCache` and `deletePieceCache`. The dfget client calls these APIs to notify the supernode about the state of cache. 

This PR is related to #1436 #1164. 

The unit tests have been added.


qhb1001:
The stream mode is fullfilled with three more perpectives, which are DFGET downloader, DFGET uploader, and the IPC between user and DFGET. 

### DFGET Downloader

Since the client stream writer has been implemented under the `p2p_downloader` folder, which means that the pieces downloading process has been finished. My work here is to fullfill the task of handling the successfully downloaded pieces to uploader. Besides that, in the perparation phase at `dfget/core/core.go:171`, the registration of stream task is necessary for uploader. Here are all the modifications:

*   `dfget/core/core.go:210`: The  `doDownload` method is refactored. One new interface is defined as `DownloadTimeoutTask`. The interface has two implementations, which are regular downloader and the stream downloader. 

*   `dfget/core/downloader/downloader.go:79`: The `Start()` method of the `StreamDownloadTimeoutTask` struct will call the `RunStream` method of the downloader, and then pass the stream reader to the `startWriter` method. 

*   `dfget/core/downloader/downloader_util.go:55`: The `startWriter` method is used to fetch the pieces from the stream reader concurrently, and then upload the pieces to uploader in order. The logic of the `startWriter` is referenced from the stream downloading method in CDN manager at `supernode/daemon/mgr/cdn/super_writer.go:59`. 

    The unit test has been added. 

### DEGET Uploader

The previous uploader is solely used for the regular downloading method. For the stream uploader, I have added new APIs for it. Apart from that, the cache manager is added to manage the cache pieces. 

*   `dfget/core/api/supernode_api.go:359`: A new kind of updating piece status API has been implemented. It is used to change the peice status from SUCCESSFUL to UNCACHED. 

*   `dfget/core/api/uploader_api.go:108`: The uploader API `RegisterStreamTask` is added to register the stream task at cache manager. The initialization of the according entry at cache manager is finshed after the call. 

*   `dfget/core/api/uploader_api.go:127`: The uploader API `DeliverPieceToUploader` is used to handle the piece from downloader to uploader. 

*   `dfget/core/uploader/cache.go:60`: The FIFO cache manager is created to control the piece caches. It would store and pop the piece in the style of FIFO. It should be noted that the cache manager is using the sender window, which has the range of `[start, una)`.

    The unit test has been added. 

### IPC Between DFGET and User

This part demands further discussion. Currently, I propose that the DFGET can directly output the content of the pieces to stdout, and then the user who calls the DFGET command, can redirect the stdout of the DFGET command to the write side of the pipe by generally supported method [popen](https://pubs.opengroup.org/onlinepubs/009695399/functions/popen.html). 

After that, the user can directly read from the pipe, and get the successfully downloaded content. Since the method uses the unnamed pipe, the whole process is not related with the file system.

qhb1001:
@garfield009 @ansinlee @Starnop PTAL

test/cli_dfget_p2p_test.go
```diff
@@ -114,3 +114,65 @@ func (s *DFGetP2PTestSuite) TestDownloadFile(c *check.C) {
 		}
 	}
 }
+
```
qhb1001, line 117: The integration test here is annotated due to the reason that the detailed implementation of IPC between user and DFGET demands further discussion.

jinuxstyle:
Not clear what problem the PR solves. Performance? Better rate limiting? or else ?

PR #1455, opened 2020-08-07
realize feature #1427

### LINKED ISSUE #1427: Proposal: a raw method to realize dynamical rate limiting #1344 #1357
# A raw method to realize dynamical rate limiting
It's just the raw version of my future implementation, maybe rough, but could still be optimizing later.
### My implementation plan
1.  Add {weight、rateLimiter} in PeerMgr struct
2.  Design a schedule algorithm that will optimize the process that one PeerNode download from a high weight level PeerNode, also maintains weight in PeerMgr struct. (As the picture below)
image
3.  Write a program to listen host total bandwidth(HTB Listener)
4.  Design a algorithm dynamically adjust the host rateLimiter, and send it to SuperNodePeer.(The flow chart and the algorithm are as below)
image
image
5.  Add cli option to determine whether to open dynamically rateLimiter.
6.  Benchmark and optimization.
### Additional information
+ There is `ProducerLoad` in `progress_manager`, may be we should using `ProducerLoad` instead of `weight` and add `rateLimiter` in each `progress_manager`.
+  We could use gopacket to listen host local bandwidth.
+  The dynamical rate algorithem may be too rough, TCP Congestion Control Algorithm could be a better method to imitate.

zhaxzhax:
This PR just realize dynamic bandwidth limit, the interpret is as below
#### cli add
`--dynamic` which could allow you to open the dynamic bandwidth limit mode.
#### api add
Add two apis, which used to send bandwidth limit update information from peerserver to supernode. 
+ `/api/v1/peer/{id}/peerState/{dynamicRate}`   The 1.0 version api.
+ `/peer/dynamicrate`   The 0.3 version api.
#### field add
+ Add field `DynamicRate` in struct `PeerState`
#### implement scheme
+ `host_bandwidth_listener`  listen to the host bandwidth, add set the 90% of current available bandwidth 
as `totalLimit` ，also as `dynamicRate` be sent to supernode
+ supernode use `dynamicRate`  and  `ProducerLoad` to calculate the priority of `peer`, the schedule algorithm will schedule the 
 `peer` with higher priority first.
#### part which haven't implement now
+ host_bandwidth_listener

Fixes #1427

Already added.

As issue #1427 

Host_bandwidth_listener haven't been completed, which will be finished in next weeks.


zhaxzhax:
I have add the HTBListener implement, which have used [Gopacket](https://github.com/google/gopacket) that based on `libpcap`. 
So may be it's hard to pass the ci test.

jim3ma:
@Starnop PTAL

zhaxzhax:
@garfield009 @ansinlee @starnop PTAL

Issue #1459, opened 2020-08-17
 Md5 not verified after download


wuchaojing:
problem：Md5 not verified after download

the method：
// PostRun
func (cw *ClientWriter) PostRun(ctx context.Context) (err error) {

	src := cw.clientFilePath
	if cw.acrossWrite || !helper.IsP2P(cw.cfg.Pattern) {
		src = cw.cfg.RV.TempTarget
	} else {
		if _, err := os.Stat(cw.clientFilePath); err != nil {
			logrus.Warnf("client file path:%s not found", cw.clientFilePath)
			if e := fileutils.Link(cw.serviceFilePath, cw.clientFilePath); e != nil {
				logrus.Warnln("hard link failed, instead of use copy")
				fileutils.CopyFile(cw.serviceFilePath, cw.clientFilePath)
			}
		}
	}
	if err = downloader.MoveFile(src, cw.cfg.RV.RealTarget, cw.cfg.Md5); err != nil {
		return
	}
	logrus.Infof("download successfully from dragonfly")
	return nil
}

This code：downloader.MoveFile(src, cw.cfg.RV.RealTarget, **cw.cfg.Md5**)
When md5 is not specified in the parameter, cw.cfg.Md5 in this method should be returned by supernode when code=600

advice：
The response returned by the /peer/task interface should be passed to the PostRun method，and then maybe it can：

	// calcute realMd5
        var expectMd5 string
	if p2p.cfg.Md5 != "" {
		expectMd5 = p2p.cfg.Md5
	} else {
		finishData := response.FinishData()
		if finishData != nil {
			expectMd5 = finishData.Md5
		}
	}

	// move file to the target file path.
	if err := downloader.MoveFile(src, p2p.targetFile, expectMd5); err != nil {
		return err
	}

jim3ma:
Yes, the finish data contains md5 info,  we should check the md5 to ensure file is correct.

wuchaojing:
> Yes, the finish data contains md5 info, we should check the md5 to ensure file is correct.

OK，I will submit a PR to fix this problem.

PR #1462, opened 2020-08-20
feature:compare md5 after downloaded

wuchaojing:
feature: when the user does not define md5, the md5 calculated by supernode is compared with the local file md5 to prevent the file from being modified during transmission

#1459

N/A

N/A

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,7 +66,7 @@ jobs:
 
   unit-test-golang:
     docker:
-      - image: circleci/golang:1.12.10
+      - image: circleci/golang:1.13.15
     working_directory: /go/src/github.com/dragonflyoss/Dragonfly
     steps:
       - checkout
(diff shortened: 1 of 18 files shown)


dfget/core/downloader/p2p_downloader/client_writer.go
```diff
@@ -153,7 +153,18 @@ func (cw *ClientWriter) PostRun(ctx context.Context) (err error) {
...
+	var expectMd5 string
+	if cw.cfg.Md5 != "" {
+		expectMd5 = cw.cfg.Md5
+	} else {
+		finishData := response.FinishData()
+		if finishData != nil {
+			expectMd5 = finishData.Md5
+		}
+	}
+
```
jim3ma, line 166: I think updating p2p.cfg.Md5 in func p2p.run will be better.

```
if code == constants.CodePeerContinue {
				p2p.processPiece(response, &curItem)
			} else if code == constants.CodePeerFinish {
				p2p.finishTask(ctx, pieceWriter)
				p2p.cfg.Md5 = response.FinishData().Md5
				return nil
			}
```

wuchaojing: Great!

PR #1476, opened 2020-09-07
change http code from 409 to 208 when create preheat which is already…

ansinlee:
… existed


jim3ma:
LGTM

allencloud:
Why to change this?  Could you please show us more explanation? 

In addition, we also need to update the swagger.yml.

PR #1488, opened 2020-09-27
support peer to peer prefetch

hhhhsdxxxx:
Only first seed node will download resources from source, others will download(prefetch) from other peer nodes instead.

### DIFF
--- a/apis/swagger.yml
+++ b/apis/swagger.yml
@@ -1493,11 +1493,17 @@ definitions:
     properties:
       task:
         $ref: "#/definitions/TaskInfo"
-      pieces:
-        type: "array"
-        description: "The pieces which should belong to the peer node"
-        items:
-          $ref: "#/definitions/PieceInfo"
+      path:
+        type: "string"
+        description: |
+          path is used in one peer A for uploading functionality. When peer B hopes
+          to get piece C from peer A, B must provide a URL for piece C.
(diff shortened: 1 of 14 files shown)


hhhhsdxxxx:
@jim3ma cc

Starnop:
@hhhhsdxxxx What problem is this PR to solve? When

Starnop:
@hhhhsdxxxx

Starnop:
@hhhhsdxxxx Cloud you please add a detailed description for this PR problem is this PR to solve? For example, explain the usage scenarios, the problems to be solved, etc.

The current scheduling algorithm will give priority to downloading from other peers in the P2P network. When there is no peer node available in the P2P network, it will try to download from the supernode. If the supernode is currently serving too much(In default is 4.) pieces are being downloaded, the peer will wait for a while.

hhhhsdxxxx:
> @hhhhsdxxxx Cloud you please add a detailed description for this PR problem is this PR to solve? For example, explain the usage scenarios, the problems to be solved, etc.
> 
> The current scheduling algorithm will give priority to downloading from other peers in the P2P network. When there is no peer node available in the P2P network, it will try to download from the supernode. If the supernode is currently serving too much(In default is 4.) pieces are being downloaded, the peer will wait for a while.
> 
> also cc/ @jim3ma
update.

It's a different p2p system service for image-service.

PR #1492, opened 2020-09-29
fix(supernode/httpclient): use HEAD request(not GET) to GetContentLength

liangran10000:
Ⅰ. Describe what this PR did
use HEAD request(not GET) to GetContentLength
Ⅱ. Does this pull request fix one issue?
no
Ⅲ. Why don't you add test cases (unit test/integration test)? (你真的觉得不需要加测试吗？)
Ⅳ. Describe how to verify it
Ⅴ. Special notes for reviews


jim3ma:
Thanks for your work.

Head request is not work for some web server. We need to handle this issue.

PR #1498, opened 2020-10-22
Fix unlimited piece concurrent download

jim3ma:
Signed-off-by: Jim Ma <majinjing3@gmail.com>

Add a flat to limit piece concurrency.

Sometime supernode schedules a lot of pieces for one dfget task, currently, dfget will download all pieces concurrently, while cost a lot of memory.

N/A


Starnop:
@jim3ma Would you mind answer a question? As far as I know, the supernode will control only run 4 pieces will be assigned to a peer at the same time, so when will the situation mentioned in this PR occur?

jim3ma:
@Starnop the pieces count can be set in supernode config

PR #1501, opened 2020-11-03
bugfix: CDN == "source" does not request pieces from source

m0nstermind:
by dfget having suprnodes was configured with explicily defined port

When supernode listens on non default port ( e.g. other than 80 for http ) the CDN Pattern: source ( to request pieces from source url ) does not work. dfget try to load pieces for the first time from supernode and fail, because supernode does not have them.

The source of this is in comparation performed by dfget at power_client.go, which compares piece destination ip, received by downloader to the url of supernode PowerClient is initialized with ( which comes in form ip:port ). In case of default port value, it is omitted from configration and this check passes and always fails for non default port values.
The PR fixes this to compare ip to ip.

none

### Ⅲ. Why don't you add test cases (unit test/integration test)?
The fix is trivial

Prepare supernode.yaml with
```
base:
  cdnPattern: source
  listenPort: 5002
```

run supernode without running nginx

run dfget with 
`dfget --node 127.0.0.1:5002 http://source/url/to/get`

dfget will fail to download any pieces from source url, trying to get it from 127.0.0.1:8001 ( the default download port of supernode )
none


jim3ma:
Please sign your commit with `git commit -s --amend` than force push it.
Thanks

m0nstermind:
BTW unit test failed with
image_preheater_test.go:38: Get https://registry.cn-zhangjiakou.aliyuncs.com/v2/acs/alpine/manifests/3.6: net/http: TLS handshake timeout

which, i believe is not about this PR, but due to connectivity problems with registry.cn-zhangjiakou.aliyuncs.com

Starnop:
LGTM.
also cc/ @jim3ma

Starnop:
> BTW unit test failed with
> image_preheater_test.go:38: Get https://registry.cn-zhangjiakou.aliyuncs.com/v2/acs/alpine/manifests/3.6: net/http: TLS handshake timeout
> 
> which, i believe is not about this PR, but due to connectivity problems with registry.cn-zhangjiakou.aliyuncs.com

I have rerun the CI.

PR #1504, opened 2020-11-06
bugfix: fix connection leak when statuscode check is not passed

pangsq:

When CDN is triggered, the supernode will send http request to the sourceUrl to download the file.
If the http response status code is not 200 or 206(e.g. 403), the transport will not be closed.
This PR fixes this bug. 

yes.

Obviously, resp.Body.Close() is always needed.

Providing an file download url which needs authentication (for example, a s3 share link), use dfget to download that file for hundreds of  times.

Then `watch "lsof | grep socket | wc -l"` on the supernode server, lots of sockets will be open even if the dfget tasks are already failed.

none


lowzj (commented):
LGTM

PR #1528, opened 2021-01-15
testing: cleanup temporary directory

zhijianli88:
MkDir: https://github.com/go-check/check/blob/v1/check.go#L170

Signed-off-by: Li Zhijian <lizhijian@cn.fujitsu.com>


zhijianli88:
ping

PR #1561, opened 2021-08-16
Update netutils.go

guoguodan:
add param check，it may be null pointer


guoguodan:
add param check，it may be null pointer

